PR #2, opened 2019-01-09
Feature/backendskeleton

jaderabbit:
Sorry for the massive trashfire of a PR. 

Things this adds:
- The skeleton for the flask app
- A skeleton for user auth
- Dockerfile & docker compose (they worked for me. anyone else care to have a go?)
- Ability to run tests
- SQLAlchemy (Looks amazing. Loose ORM - just the way I like them :)) 
- Ability to run migrations (alembic also is loose. So manually written migrations with ORMs help, rather than Django's auto-migrations which are horrible) 
- .gitignore (@KaleabTessera maybe add the react/node things to it? 
- Adds a .circleci helloworld

Things I am still figuring out:
- Creating a single docker-compose for both frontend and backend (and how that would look)
- Busy playing with circleci. 

Any comments so far?


KaleabTessera:
Looks great! 

Everything worked well, I just had a minor issue when running `docker-compose build`. 
I got the following error:
```
 E: There are problems and -y was used without --force-yes
ERROR: Service 'web' failed to build: The command '/bin/sh -c apt-get install -y socat git software-properties-common python-software-properties postgresql-client-9.3 postgresql-client-common' returned a non-zero code: 100
```

I added --force-yes to the following lines and it worked. 
RUN apt-get install -y  **--force-yes** socat git software-properties-common python-software-properties postgresql-client-9.3 postgresql-client-common
RUN apt-get install -y  **--force-yes** python-pip python-psycopg2 libpq-dev python2.7-dev gunicorn libmagic1

Maybe we should consider adding this to the docker-compose file? I am a bit confused as to why it didn't work directly on mine though. Maybe compare versions:
`docker-compose version 1.23.1, build b02f1306
docker-py version: 3.5.0
`


Regarding the gitignore, I have already added a gitignore in the web app folder so that is not an issue. :)

I can add the front-end to the docker compose file.

PR #3, opened 2019-01-10
Feature/fixsecurityalert

jaderabbit:
I've upgraded everything to use latest flask so solve a security thing github actually notified us about


jaderabbit:
@avishkar58 Right?! I was well impressed

PR #5, opened 2019-01-11
Feature/front end skeleton part2

KaleabTessera:
Things this adds:
- Added Home Page
- Added Jest and Enzyme for testing
- Added React-Bootstrap
- Added Tests to see if pages render.
- Added Basic NavBar.
- **Updated docker-compose file to include web and moved it out of api folder.**

Let me know if you guys can run docker-compose up and see the website - port 8080. 

Things to do: 
- Decide on docker-compose version we are all going to use. The format of the files are different.
- Show how service calls will be made.


webapp/public/index.html
```diff
@@ -22,6 +22,10 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+    <!-- React Bootstrap -->
+    <!-- Latest compiled and minified CSS -->
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
```
avishkar58: Why not Bootstrap 4? Latest version is 4.2.1

<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css" integrity="sha384-GJzZqFGwb1QTTN6wy59ffF1BuGJpLSa9DkKMp0DgiMDm4iYMj70gZWKYbI706tWS" crossorigin="anonymous">

avishkar58: Oh nevermind, I see it's because react-bootstrap only supports 3.X

KaleabTessera: Yeah so the react-bootstrap say they are working on bootstrap 4 support, but that will probably take a while.

We do have the option of not using react-bootstrap and using plain bootstrap 4, but that we cant use the react-bootstrap components or some of its features. 

It is basically 
```
// React-Bootstrap
<NavItem eventKey="1" href="/home">
          NavItem 1 content
</NavItem>
```
vs
```
// Normal Bootsrap
<li className="nav-item">
      <a href="/home className="nav-link"> NavItem 1 content </a>
 </li>
```

We can also change at anytime, but react-bootstrap makes life easier for people who aren't comfortable with html + css.

I am comfortable either way, what do you two think? @avishkar58 @jaderabbit

avishkar58:
Hmmm, interesting! My first instinct is that I'm not sure how much benefit
it provides (react-bootstrap). Is it commonly used these days? Will play
with an example with and without it this weekend to inform myself

avishkar58:
I'm now getting an error when running docker-compose build/up in the root - looks like it's failing to download some of the dependencies: eg: 
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/k/krb5/libgssapi-krb5-2_1.12+dfsg-2ubuntu5.3_amd64.deb  404  Not Found [IP: 91.189.88.162 80] 
It was working before... Did you have this problem? What version of Docker are you running?

PR #26, opened 2019-01-19
feature/password_reset_email

nischalhp:
This branch provides the following features:

1. Functionality to send email via amazon SES
2. Logging


nischalhp:
screen shot 2019-01-19 at 11 10 04 pm

jaderabbit:
So, if you add the environment variables to docker-compose, then the migrations pass
But then, assumedly, people trying to run this locally won't actually be able to send verification emails (desired behaviour I suppose?) 

How did you add the environment variables to your system?

I branched your branch to add those dummy variables to play around and see if it would make docker-compose happy:
https://github.com/deep-learning-indaba/Baobab/tree/tmp/test

It allowed the migrations to run: https://circleci.com/gh/deep-learning-indaba/Baobab/59

But there is a new error. It looks like it's been caused by these changes:
- https://github.com/deep-learning-indaba/Baobab/blob/d90338e6f7671428fbc50bc04b425be501e7e7d4/api/app/__init__.py#L23
- https://github.com/deep-learning-indaba/Baobab/blob/d90338e6f7671428fbc50bc04b425be501e7e7d4/api/app/routes.py#L3

Once I changed that I got another error: 
https://circleci.com/gh/deep-learning-indaba/Baobab/59

 Are you running with `docker-compose` or did you install locally? Because it looks like we're running very different things :P

nischalhp:
@jaderabbit 

** Email**

I have added the environment variables manually with export commands, is there a script to set all the environment variables?

With regards to mocking, I am unsure on the functionality because you need to really test if the emails are reaching somewhere, maybe I disable the tests but have the test case written in case of supporting the test run only for individuals who have access to the credentials? 

**Import issue**
Also with regards to the import issues:

I ran them locally and hence that is causing issues I think. Without ```docker-compose``` any idea what to set the ```PYTHONPATH``` as? Or is ```docker-compose``` the only preferred way off running things locally?

jaderabbit:
@nischalhp 

*Email*
One can set them in the docker-compose so that the testing suite can build. I'd assume they're private so should not be set in the repository itself. In which case I propose we do the following:

1. Tell docker compose to use the environment variables from the box the dockers are running on (I figured this out the other day, so will do this, if you're happy with it? :) )
2. Tell 

*Import issue*
Well, it should not be necessary to change the PYTHONPATH in order for this to run. My thought is you're perhaps using a different version of python? The app runs on 2.7 currently. Are you running 3.6? Also, what command are you using to run the server? 

I'd recommend to use virtualenv if you're running locally. I'll try get the README setup for local installation tonight.

`docker-compose` is the preferred installation method. As someone who is generally very sceptical of docker and it's friends, even I recommend it :) That said, there should be no reason why you can't get it installed locally, if you really would like to avoid it

jaderabbit:
Update: Just ran it locally and it worked fine

nischalhp:
@jaderabbit :

Okay, this makes more sense now. Thank you for taking the time out to explain that. What I will do then is, fix the issues of imports, run it with docker and send the updates on this PR.

We still need to decide how we want to test email functionality now, maybe I turn off the test cases, just for now until we figure that out.

jaderabbit:
@nischalhp Drop the test case for now. :) Make sure it builds once with it. There will be a way to mock it or create a sandbox

nischalhp:
> @nischalhp Please see circleci build

Yes on it today. Have no idea why its failing but will fix it today and ensure the PR is ready to be merged.

nischalhp:
screen shot 2019-01-26 at 12 32 26 pm

@jaderabbit  - Any idea why I am getting this error when I do ```docker-compose up``` ??

nischalhp:
@jaderabbit - I think I have done something that has screwed things up, will remove this PR, create another branch from ```develop``` and migrate the code there. The ```develop``` branch seems to compile appropriately and run. I have no idea what is going wrong otherwise.

Issue #11, opened 2019-01-13
Add fields to User model


avishkar58:
Add the fields found in the [Application Process spec](https://docs.google.com/document/d/1173JsQzl0KnjtLZ4bE5PK2j_YDTJF55ssoR9I19tQGs/edit#) data model diagram to the backend AppUser model (api/app/users/models.py)

Create a migration to apply the changes to the database

Update the UserAPI in api/app/users/api.py to accept and populate the new fields 

**Definition of Done**
This task is done when:
* The new fields have been added to the model according to the spec
* A migration has been created and can be successfully applied to the database
* The API endpoints in api/app/user/api.py have been updated to account for these new fields
* The tests in api/app/user/tests.py have been updated to incorporate the new fields

PR #31, opened 2019-01-22
Update/usermodel

amritpurshotam:
This is the PR for this ticket https://github.com/deep-learning-indaba/Baobab/issues/11

According to the spec and a conversation I had with @avishkar58 on the spec document, we said that user category should have it's own table to enable easier reporting. I went a bit further and also split out country, gender, title, ethnicity, and disability in to their own tables as well for the same reasons. Let me know if this is fine? Also one drawback around this approach is that I'm not sure how to handle an option that requires "Other" and the accompanying string that specifies what it is. Do I add these fields in the app_user table? Or split them out into their own tables too? Or just not handle this case for now?

Lastly, I know the unit tests are failing. I'm a bit stuck on how to get them to pass. It seems like the database gets wiped after a unit test is run and I'm not sure how to get it to reset. Any help with that will be appreciated.


api/migrations/versions/7f3d601390cb_.py
```diff
@@ -0,0 +1,105 @@
...
+    op.create_foreign_key(None, 'app_user', 'user_disability', ['user_disability_id'], ['user_disability_id'])
+    op.create_foreign_key(None, 'app_user', 'user_title', ['user_title_id'], ['user_title_id'])
+    op.drop_index('unique_user_code', table_name='password_reset')
+    # ### end Alembic commands ###
+    op.bulk_insert(country, [{'country_id': 1, 'country_name': 'South Africa'}])
+    op.bulk_insert(user_category, [{'user_category_id': 1, 'name': 'Postdoc', 'description': 'Postdoc', 'group': 'Postdoc'}])
+    op.bulk_insert(user_disability, [{'user_disability_id': 1, 'disability_name': 'None'}])
+    op.bulk_insert(user_ethnicity, [{'user_ethnicity_id': 1, 'ethnicity_name': 'None'}])
+    op.bulk_insert(user_gender, [{'user_gender': 1, 'gender_name': 'Male'}])
+    op.bulk_insert(user_title, [{'user_title_id': 1, 'title_name': 'Mr'}])
```
amritpurshotam: Not sure if this is the way to add static data to the tables? Also this is just dummy data for now. Once the correct way of doing this is confirmed, I'll fill in the rest of the static data using the existing Google Form template

jaderabbit:
@amritpurshotam please check the circleci build. It's failing on the downgrade migrations

api/migrations/versions/7f3d601390cb_.py
jaderabbit: I'm sure it's fine, but I'd avoid setting the ids manually and rather make sure postgres is generating them. Also, I doubt you'd need to use a bulk insert, you could probably use the orm?

api/migrations/versions/7f3d601390cb_.py
```diff
@@ -0,0 +1,105 @@
...
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_index('unique_user_code', 'password_reset', ['user_id', 'code'], unique=True)
+    op.drop_constraint(None, 'app_user', type_='foreignkey')
```
jaderabbit: These constraints with no name seem to be failing on the downgrade

jaderabbit:
Also 

> I went a bit further and also split out country, gender, title, ethnicity, and disability in to their own tables as well for the same reasons. Let me know if this is fine? 

Besides Category and Country, I wouldn't have tables for each of those. It would create more work than benefits

api/app/users/models.py
```diff
@@ -42,3 +82,29 @@ class PasswordReset(db.Model):
 
     def __init__(self, user):
         self.user = user
+
+class UserTitle(db.Model):
+    user_title_id = db.Column(db.Integer(), primary_key=True)
+    title_name = db.Column(db.String(10))
+
+class Country(db.Model):
+    country_id = db.Column(db.Integer(), primary_key=True)
```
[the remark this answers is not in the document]
amritpurshotam: Done. Also renamed the columns to `name` instead of `table_name`

api/migrations/versions/7f3d601390cb_.py
amritpurshotam: Thanks. Just needed to give the foreign keys names.

amritpurshotam:
> Besides Category and Country, I wouldn't have tables for each of those. It would create more work than benefits

I disagree here. I think it will make reporting and filtering easier. We reduce the chance of any data integrity issues and it allows the options to be configurable without a redeploy. And the cost of the slightly extra work that needs to be done I don't think outweighs those benefits.

api/app/utils/testing.py
```diff
@@ -9,8 +9,10 @@ def setUp(self):
         app.config['TESTING'] = True
         app.config['DEBUG'] = True
         self.app = app.test_client()
+        db.drop_all()
         db.create_all()
 
     def tearDown(self):
         db.session.remove()
         db.drop_all()
+        db.engine.execute('DELETE FROM alembic_version;')
```
amritpurshotam: Can I get feedback on the changes I made here please since I'm not 100% sure if I was just missing something and this was just a hack?

The reason I do `db.drop_all()` in the `setUp()` method is because I want the db to start from a clean slate. I noticed that while dev'ing, if you run the tests, the state of the db is used as it already is making the tests slightly volatile.

The last line to clear the `alembic_version` table is to ensure that I can run the `upgrade` after running the tests. The `tearDown()` drops all the tables except the `alembic_version` which has the name of the last migration script that was run. If I run `upgrade`, it doesn't do anything then since it thinks all the tables are there but of course the tests dropped the tables.

api/app/users/mixins.py
```diff
@@ -1,8 +1,25 @@
...
 
-class SignupLoginMixin(object):
+class SignupMixin(object):
 
     req_parser = reqparse.RequestParser()
-    req_parser.add_argument('email', type=str, required=True)
+    req_parser.add_argument('email', type=str, required=True, )
+    req_parser.add_argument('firstname', type=str, required=True)
+    req_parser.add_argument('lastname', type=str, required=True),
+    req_parser.add_argument('user_title_id', type=int, required=True)
```
jaderabbit: Will need to update the id's here :)

api/app/users/models.py
```diff
@@ -11,16 +11,64 @@ def expiration_date():
...
-    password = db.Column(db.String(255))
-    active = db.Column(db.Boolean())
-    is_admin = db.Column(db.Boolean())
+    email = db.Column(db.String(255), unique=True, nullable=False)
+    firstname = db.Column(db.String(100), nullable=False)
+    lastname = db.Column(db.String(100), nullable=False)
+    user_title_id = db.Column(db.Integer(), db.ForeignKey('user_title.id'), nullable=False)
+    nationality_country_id = db.Column(db.Integer(), db.ForeignKey('country.id'), nullable=False)
+    residence_country_id = db.Column(db.Integer(), db.ForeignKey('country.id'), nullable=False)
+    user_ethnicity_id = db.Column(db.Integer(), db.ForeignKey('user_ethnicity.id'), nullable=False)
```
jaderabbit: I still see id's. Can we drop gender_id, ethnicity_id, disability_id, and user_title id?

amritpurshotam: Do you mean it should be strings and we shouldn't have tables for the options?

Because I disagree on that approach. I think it will make the reporting and filtering easier. We reduce the chance of any data integrity issues and it allows the options to be configurable without a redeploy.

amritpurshotam: Sure. It probably is a over-normalised. I'll update.

amritpurshotam: @jaderabbit Changes have been pushed. Can you also please check the comment I made on `testing.py` file? I'm not sure what I did there was correct.

api/app/utils/testing.py
jaderabbit: This is perfect

api/migrations/versions/7f3d601390cb_.py
jaderabbit: @amritpurshotam Agreed that the bulk entry is the correct way for lots of data :)

api/app/users/models.py
jaderabbit: Clearing up the db before each test is perfect :)

amritpurshotam: @jaderabbit Thanks for checking.

And woohoo, my first ever PR to be accepted to an open source project :)

jaderabbit: @amritpurshotam Congrats! It's always a tremendous feeling :smile:

PR #33, opened 2019-01-23
Added Event and EventRole models, as well as migrations for the table

Carvalho96v2:
Please confirm the migrations are okay


api/migrations/versions/38dde1052c5e_.py
```diff
@@ -0,0 +1,56 @@
...
+            'EndDate': '31 August 2019'
+        }
+    ])
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('Event')
+    op.drop_table('Event_Role')
```
amritpurshotam: From the error on CircleCI, it looks like the `downgrade()` is failing. The error it's getting is 
> sqlalchemy.exc.InternalError: (psycopg2.InternalError) cannot drop table "Event" because other objects depend on it

It seems that `Event_Role `depends on `Event` so swapping the two lines around should get the test to pass since you'll be dropping the dependent table first. Hope this helps.

api/app/events/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Event(db.Model):
+
+    __tablename__ = "Event"
+
+    id = db.Column(db.Integer(), primary_key=True)
+    name = db.Column(db.String(50))
+    description = db.Column(db.String(255))
+    startDate = db.Column(db.DateTime())
+    endDate = db.Column(DateTime())
```
amritpurshotam: Something else I noticed is that the non-primary key columns are set to be nullable by default. Reading through the spec again now, it seems that these columns should always be required. Can we set this in the database?

You can do so by setting a nullable parameter as follows
`name = db.Column(db.String(50), nullable=True)`

You will need to re-run the migration though to get the updates.

api/app/events/models.py
```diff
@@ -0,0 +1,55 @@
...
+       self.endDate = newEndDate
+
+class EventRole(db.Model):
+
+    __tablename__ = "Event_Role"
+
+    id = db.Column(db.Integer(), primary_key=True)
+    eventID = db.Column(db.Integer(), db.ForeignKey("Event.event_id"))
+    userID = db.Column(Integer(), db.ForeignKey("app_user.id"))
+    role = db.Column(db.String(50))
```
amritpurshotam: The same for these columns as well.

api/migrations/versions/38dde1052c5e_.py
```diff
@@ -0,0 +1,56 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    event = op.create_table('Event',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('Name', sa.String(length=255)),
```
jaderabbit: can we still with the convention of small letters for our column names?

api/migrations/versions/38dde1052c5e_.py
```diff
@@ -0,0 +1,56 @@
...
+    sa.Column('Description', sa.String(length=255)),
+    sa.Column('StartDate', sa.DateTime()),
+    sa.Column('EndDate', sa.DateTime()),
+    sa.PrimaryKeyConstraint('id')
+    )
+
+    op.create_table('Event_Role',
```
jaderabbit: and same goes for table names

api/migrations/versions/38dde1052c5e_.py
Carvalho96v2: Yes, this resolved the downgrade issue, thanks!

api/app/events/models.py
Carvalho96v2: Applied this change and reran the migrations

api/migrations/versions/38dde1052c5e_.py
Carvalho96v2: Done :)

Carvalho96v2:
Cool, I believe I've addressed all the abovementioned issues :)

api/app/events/models.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, role, user_id, event_id):
+        self.role = role
+        self.user_id = user_id
+        self.event_id = event_id        
+
+    def set_user(self, new_user_id):
+       self.userID = new_user_id
+
+    def set_event(self, new_event_id):
+       self.eventID = new_event_id
```
amritpurshotam: Seems these `eventID `and `userID `variables were missed during the refactor.

Carvalho96v2: Oof, missed those. Addressed in the latest push :)

PR #35, opened 2019-01-26
Feature/email logger support

nischalhp:
* Supporting email with Amazon SES
* Adding logging capabilities
* Issue 13


nischalhp:
@jaderabbit - We need to set the SMTP credentials or the user API tests will keep failing too. Let me know how to proceed.

Issue #20, opened 2019-01-14
Create ApplicationForm front-end


avishkar58:
Create the necessary React components to implement an application form for an event. 
This issue just involves creating the form with the relevant field types based on the JSON form specification returned by the API - submitting this form is another task. 

Note that it should appear to the user of the app that each section is on a separate "page". The title of this page will be the section name, followed by the section description. The user should click on a "Next" button to progress to the next section. 

You can work with a dummy API response for now so that this can be done independently of #19 

Example JSON that we expect the API to send:

{
"eventId": 1,
"is_open": true,
"deadline": "2019-04-30",
sections: [
    {
    "sectionId": 1, 
    "name": "Section 1", 
    "description": "Section 1 description", 
    "order": 1,
    "questions": [
        {"id": 1, "description": "Question 1", "type": "short-text", "required": true, "order": 1}, 
        {"id": 2, "description": "Question 2", "type": "single-choice", "required": false, "order": 2},
    ]
    },
    {
    "sectionId": 2, 
    "name": "Section 2", 
    "description": "This is section 2", 
    "order": 2,
    "questions": [
        {"id": 3, "description": "What is 2+2?", "type": "short-text", "required": true, "order": 1}, 
        {"id": 4, "description": "Upload your CV", "type": "file", "required": false, "order": 2},
    ]
    },
]

}

**Definition of Done:**
This task is done when:
* There are new components for ApplicationForm, Section and Question
* There is a new ApplicationForm page that composes components based on the JSON spec
* There is a mock applicationForm service with the relevant API call
* There is at least one basic unit test of the page

avishkar58:
PR #36

PR #36, opened 2019-01-26
Feature/app form frontend

avishkar58:
Added skeleton Application Form as per issue #20:
- Added components for ApplicationForm, Section, FieldEditor and Confirmation
- Tracking state to build up response
- Created dummy API calls with a hard-coded JSON response for now (until the backend is done as we can wire it up) 
- Created simple rendering test

There is a bug at the moment, where if you type something in the first input field, then go to the next section, the previous text entered is still visible in the "new" input field of the 2nd section. The input field seems to change (the id is different), so I'm not sure why this is happening... 

Will create new issues for the following
- Get the file upload working
- Improve UI

### DIFF
--- a/webapp/src/App.js
+++ b/webapp/src/App.js
@@ -4,6 +4,7 @@ import logo from './images/logo-32x32-white.png'
 import Home from './pages/home';
 import Login from './pages/login';
 import CreateAccount from './pages/createAccount'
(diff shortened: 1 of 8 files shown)


avishkar58:
Thanks, have updated with more defensive coding and moved strings to constants. Manually tested that it doesn't "blank screen" when I randomly remove some properties.

PR #37, opened 2019-01-27
Issue18

Carvalho96v2:
Added the ApplicationForm, Section, and Question models, as well as their migrations.

Addressed the migration issue introduced in the issue17 merge where there were multiple revision paths.

Addressed an issue in the tests where the setUp() method would fail because of newly introduced table dependencies.

Addressed an issue where deleting from the alembic_version table in the tests would fail because the table had been dropped before that.

### DIFF
--- a/api/app/applicationModel/models.py
+++ b/api/app/applicationModel/models.py
@@ -10,7 +10,7 @@ class ApplicationForm(db.Model):
     deadline = db.Column(db.DateTime(), nullable=False)
 
     def __init__(self, event_id, is_open, deadline):
-        self.eventID = event_id
+        self.event_id = event_id
         self.is_open = is_open
         self.deadline = deadline
         


api/app/applicationModel/models.py
```diff
@@ -0,0 +1,47 @@
...
+
+class ApplicationForm(db.Model):
+
+    id = db.Column(db.Integer(), primary_key=True)
+    event_id = db.Column(db.Integer(), db.ForeignKey('event.id'), nullable=False)
+    is_open = db.Column(db.Boolean(), nullable=False)
+    deadline = db.Column(db.DateTime(), nullable=False)
+
+    def __init__(self, event_id, is_open, deadline):
+        self.eventID = event_id
```
avishkar58: Don't you mean self.event_id ?

Carvalho96v2: Thanks! Fixed and pushed :)

PR #45, opened 2019-02-08
Added ApplicationForm controller get request to return app-form based on event_id

Carvalho96v2:
I believe I covered everything. First time following a TDD approach, so please let me know if I missed anything.

P.S. screw python for needing the __init__.py file, even if its blank. That took me way too long to figure out.

### DIFF
--- a/api/app/applicationModel/mixins.py
+++ b/api/app/applicationModel/mixins.py
@@ -1,0 +1,8 @@
+from flask_restful import reqparse
+
+
+class ApplicationFormMixin(object):
+
+    req_parser = reqparse.RequestParser()
+    req_parser.add_argument('event_id', type=int, required=True, help = 'Invalid event_id requested. Event_id\'s should be of type int.')
+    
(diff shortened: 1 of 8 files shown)


api/app/applicationModel/mixins.py
```diff
@@ -0,0 +1,8 @@
+from flask_restful import reqparse
+
+
+class ApplicationFormMixin(object):
```
avishkar58: What's the purpose of this mixin, since the functionality seems to already exist in the get method of ApplicationFromAPI ?

api/app/applicationModel/tests.py
```diff
@@ -0,0 +1,49 @@
...
+from datetime import date
+from app import db
+from app.events.models import Event
+from app.utils.testing import ApiTestCase
+from app.applicationModel.models import ApplicationForm, Section, Question
+
+
+
+class ApplicationFormApiTest(ApiTestCase):
+    test_event = Event('Test Event', 'Event Description', '2019/02/24', '2019/03/24')
```
avishkar58: Are these used? Looks like they're redefined in seed_static_data...

PR #47, opened 2019-02-10
Feature/gcp circleci

KaleabTessera:
Things added:
- Circle ci -> gcp Continuous Integration for website.

@jaderabbit , We can add IP address of the API to webapp/env.test, then when we merge branches to develop, it will use the correct API urls.


jaderabbit:
@KaleabTessera Let's add them as encrypted environment variables rather from circleci side.

KaleabTessera:
@jaderabbit What would be the benefit of encrypting the API urls? Won't it still be visible from the network tab as API calls are made?

jaderabbit:
@KaleabTessera Yes, that is true. Feels wrong being in the repo for some reason.

jaderabbit:
@KaleabTessera Yeah, let's do it. I'm being stupid

KaleabTessera:
@jaderabbit Cool, but I get what you are saying, all sensitive env variables we will store in circleci.

PR #55, opened 2019-02-13
Added application_form, section, and question migrations for indaba 2019

Carvalho96v2:
- Added headline, options, and is_required fields to question table
- Added "Additional Information" section with questions for additional comments and links to personal website / github

Note: The tests are failing on creating the "options" column because they use SQLite, which does not support the JSON type, whereas PostgreSQL does. 

@avishkar58  how do you recommend we go about this?


avishkar58:
Looks like SQLite 3.9+ supports JSON type? so maybe we just need to upgrade
SQLite or Sqlalchemy?

Carvalho96v2:
@avishkar58 cool upgrading SQLAlchemy seemed to do it. Of note though, had to upgrade it to a beta build, so we should just be mindful of potential instability.

api/migrations/versions/b3f7a73c98f3_.py
```diff
@@ -57,3 +65,311 @@ def downgrade():
...
+
+def seed_question_data(question):
+    op.bulk_insert(question,
+    [
+        {
+            'id': 1,
+            'application_form_id': 1,
+            'section_id': 1,
+            'type': 'short-text',
+            'headline': 'Email address',
```
avishkar58: We can remove this question because we already have their email address from their user profile :)

Carvalho96v2: Big oof. Got it

Issue #43, opened 2019-02-07
Create Response controller


avishkar58:
Create a controller to handle saving and submitting of Responses as documented in the [application process spec](https://docs.google.com/document/d/1173JsQzl0KnjtLZ4bE5PK2j_YDTJF55ssoR9I19tQGs/edit?ts=5c3c5fec#). 

The controller should contain
* A GET method that retrieves a response for a given event_id for the logged-in user. A serialized JSON response is returned of the form illustrated below.
* A POST method that accepts a JSON serialized response and saves the response to the database. This will make use of the data model created in #21. If a response already exists (the response_id is populated in the JSON received), it should be updated, else a new response should be inserted and the new ID returned. 
* A DELETE method that accepts a response_id and sets the is_withdrawn flag and the withdrawn_timestamp on the Response object 

**Definition of Done**
This task is done when

* A new controller exists in api/app/response/api.py implementing methods described above which communicate using JSON of the form illustrated below
* Unit tests exist for the response api
* Checks exist in the API to ensure that a user can only retrieve/update/delete their own responses

An example of the JSON serialization of the Response/Answer model (NOTE: The response_id field should only be populated if the user is updating their response)
{
"response_id": 1,
"answers": [
{"question_id": 1, "value": "answer 1"},
{"question_id": 2, "value": "stuff"),
{"question_id": 3, "value": "hello world")
],
"is_submitted": false
}

PR #56, opened 2019-02-14
Feature/response controller

avishkar58:
Implementation of #43


Carvalho96v2:
Looks amazing!!

Quick question, why do you use the mixin for the get request, but not for the post and put requests? What is the general best practice with regards to mixins? :)

api/app/utils/testing.py
```diff
@@ -16,5 +27,4 @@ def setUp(self):
     def tearDown(self):
         db.session.remove()
         db.reflect()
-        db.engine.execute('DELETE FROM alembic_version;')
```
Carvalho96v2: Also, why did you remove this line? My understanding of this is that it "resets" the current migration?

avishkar58:
Hey, I used the ApplicationFormMixin in the get request because it already defined a request parser with an event_id argument which I needed there. The push, post and delete didn't need event_id and I didn't create new Mixins because I'm not convinced about them :-P - it only makes sense if we're re-using a bunch of validation code in many places, which we aren't. Even if we end up doing so, it's better to follow YAGNI and refactor when that time comes...

api/app/utils/testing.py
avishkar58: This was done in conjunction with switching to SQLLite for our tests. Previously we ran against the "main" postgres database, which already had migrations run against it and therefore had an alembic_version table. I believe now the tables are created directly in SQLLite with db.reflect() (@jaderabbit can correct me if I'm wrong?) - ie. no migrations therefore no alembic_version table.

Carvalho96v2: Okay cool, that makes sense :)

Issue #49, opened 2019-02-12
Implement file-upload backend


avishkar58:
Implement a REST end-point (File) that accepts a multipart/form-data POST request with contents of a file. This file should be saved to Google Cloud Storage ([see this link](https://stackoverflow.com/questions/37003862/google-cloud-storage-how-to-upload-a-file-from-python-3)) with an auto-generated (hash) ID as the file-name. The generated ID should be returned to the client.

PR #57, opened 2019-02-16
Add file upload functionality

amritpurshotam:
Resolves issue #49 

I'm not sure how to handle the Google Cloud credential details. I committed dummy credentials so the upload functionality will obviously will not work for anyone else unless they fill in their own locally. How do we proceed with this?

P.S. I also used my own credentials to test. @KaleabTessera did give me access to the Baobab GCP but I still didn't have access to creating credentials.


KaleabTessera:
Hi Amrit,

Thanks for working on this! When you say you don't have access to creating credentials, do you mean access to create other user and roles? What would be the purpose of this?

We should pass through these credentials as environment variables via CircleCI. The Circleci user already has access credentials to push to cloud storage, @jaderabbit can we pass this credentials to API and then have these credentials used for pushing to gcp?

amritpurshotam:
@KaleabTessera So I setup the access to the storage using service account keys. You can create one by navigating to APIs and Services -> Credentials -> Create credentials -> Service Account Key. On the baobab project though when I go to the Credentials screen I get the following error 

> You don't have permission to view API keys, OAuth clients and service account keys

And yes at the moment I have setup the keys to be passed in via environment variables through the `docker-compose.yml` (I think this is where you would do it, I'm still new to Docker so correct me if this was the wrong place) file although these were new environment variables that I setup. So I think the credentials saved on Circle CI will need to be mapped to those new variables.

KaleabTessera:
I have updated your access rights, please check if you can create a service account.

Yeah, your right, the CircleCi env variables will need to be mapped to the compose file. 

I think a possible solution, will be to have multiple docker compose files, one for local, test and prod. That way we can use CircleCI credentials on test and prod and local creds when testing locally. 

Just waiting on @jaderabbit opinion on this.

amritpurshotam:
@KaleabTessera Thanks. Looks like I can access the credentials now.

I'm not sure multiple docker compose files would help. The challenge is still going to be protecting the credentials while making sure the developers can still access them. Since this is a public repo anyone could take those credentials and use them. 

I'm looking to see if there's possibly an emulated solution for local development or some other work around.

jaderabbit:
Hi @amritpurshotam @KaleabTessera 

Sorry, I've been AWOL

- *Docker compose is only used in testing*: We would set the credentials in circle CI which would get passed to the docker-compose. I can set this up if help is needed.
 
- *For deployment:* We can set the environment variables in the app.yaml (staging) and the prod.yaml (production) which are currently encrypted. I will send y'all the encrypt and decrypt commands via mail

amritpurshotam:
@jaderabbit @KaleabTessera 

> We would set the credentials in circle CI which would get passed to the docker-compose. I can set this up if help is needed.

That would be great thanks. I don't think we have access to Circle CI.

> We can set the environment variables in the app.yaml (staging) and the prod.yaml (production) which are currently encrypted. I will send y'all the encrypt and decrypt commands via mail

Thanks I got the email with the encryption and decryption details. I've updated the files now. Can you please double check that I didn't do anything wrong? I've also noticed the service account key has access to I think everything on GCP (I used the Circle CI one). I'm looking into how to limit the permissions but it seems trickier than I expected. If anyone knows, please shout.

Also one last thing I want to point out is that I've config'd a file size limit in the `docker-compose.yml` file (as well as the `app.yaml` and `prod.yaml` files now). It's set to 10MB at the moment. Should I change this limit? Is the config overkill or should I just hard code it in the codebase?

hbarnardt:
If this PR is merge before PR #82 the environment files should be updated on #82

jaderabbit:
@amritpurshotam Thank you!

Config is perfect. I'll pull and verify

Issue #53, opened 2019-02-13
Allow routing to specific pages


KaleabTessera:
Allow routing to specific pages - e.g. [url]/login . This works locally but fails through docker (served via nginx) and through gcp.

PR #58, opened 2019-02-17
Bugfix/route specific urls

KaleabTessera:
Implementation of https://github.com/deep-learning-indaba/Baobab/issues/53

Works through docker and deployed app - https://baobab.appspot.com/createAccount .


.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
```
avishkar58: Please remove this file from the PR

KaleabTessera: Will do.

Issue #14, opened 2019-01-13
Add fields to Create Account form


avishkar58:
Add the additional fields found in the User table of the [Application Process Spec](https://docs.google.com/document/d/1173JsQzl0KnjtLZ4bE5PK2j_YDTJF55ssoR9I19tQGs) to the front-end CreateAccount page. This depends on #11 being completed before this will work end-to-end.

**Definition of Done**
This task is done when: 
* The CreateAccount component has been updated with the additional fields found in the application process spec
* The fields are visible in the CreateAccount page
* The service calls have been updated and correctly pass the full set of fields to the back-end API
* A manual integration test has been performed as the functionality is working as expected.

avishkar58:
Hey Kale-ab, are you still working on this one? Busy with issue #10 and can do this one now quickly if you haven't started yet?

KaleabTessera:
Hey Avishkar, yes I am quite far with this one. I quickly stopped this, to get the Continuous Integration working.

I am doing a bit of rework on how this form works so i think it would be easier if I finish it. :)

avishkar58:
Ok awesome :) thanks!

KaleabTessera:
Closing for now, will create new tickets for bugs etc.

PR #59, opened 2019-02-17
Feature/redesign application form

KaleabTessera:
Implementation of https://github.com/deep-learning-indaba/Baobab/issues/14.

Thing to do:
- Currently the backend wants an id for user_category_id,residence_country_id and nationality_country_id. This shouldn't be the case. The front-end should never send these ids, backend should handle converting strings to IDs. For now an ID of 1 is used for 3 values. 
*We can possibly send ids if we create new endpoints, which retrieve countries and categories list. *
- Conformity is required in countries list. Backend has a list, front-end uses https://www.npmjs.com/package/country-list , which follows Maps ISO 3166-1-alpha-2.


webapp/src/components/form/FormSelect.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from "react";
+import FormGroup from "./FormGroup";
+import { default as ReactSelect } from "react-select";
+class FormTextBox extends React.Component {
```
avishkar58: Shouldn't this be called FormSelect ?

webapp/src/pages/createAccount/components/CreateAccountForm.js
```diff
@@ -1,84 +1,321 @@
...
+
+    if (this.state.password != this.state.confirmPassword) {
+      this.state.errors.$set.push({ passwords: "Passwords do not match" });
+    }
+    if (this.state.errors.$set.length > 0) return;
+
+    this.setState({ loading: true });
 
-      this.setState({ loading: true });
+    userService.create(this.state, this.state).then(
```
avishkar58: userService.create takes only 1 parameter.

webapp/src/components/form/FormSelect.js
KaleabTessera: Yes, apologies. I think this still worked because we exported by default and then when we imported it FormSelect was used as alias.

webapp/src/pages/createAccount/components/CreateAccountForm.js
KaleabTessera: Will fix.

PR #64, opened 2019-02-20
Response confirmation email

Carvalho96v2:
- Added emailing functionality to the response api's post method
- Added a common strings file for statically defined strings
- Mocked out the emailer service in debug environments

Please review if I mocked the emailer service out correctly

### DIFF
--- a/api/app/responses/api.py
+++ b/api/app/responses/api.py
@@ -4,9 +4,10 @@ import datetime
 from flask_restful import reqparse, fields, marshal_with
 from app.applicationModel.mixins import ApplicationFormMixin
 from app.responses.models import Response, Answer
-from app.applicationModel.models import ApplicationForm
+from app.applicationModel.models import ApplicationForm, Question
 from app.events.models import Event
 from app.utils.auth import auth_required
+from app.utils import emailer, strings
 
 from app.utils import errors
 
@@ -82,7 +83,15 @@ class ResponseAPI(ApplicationFormMixin, restful.Resource):
                 db.session.add(answer)
             db.session.commit()
 
-            return {'id': response.id, 'user_id': user_id}, 201  # 201 is 'CREATED' status code
+            try:
+                #recipient = g.current_user['email']
+                #subject = strings.build_response_email_subject(g.current_user['title'], g.current_user['firstname'], g.current_user['lastname'])
+                #body_text = strings.build_response_email_body()
+                summary = build_response_summary(response)
+            except:
+                print('Failed to send email')
+            finally:
+                return {'id': response.id, 'user_id': user_id}, 201  # 201 is 'CREATED' status code
         except:
             return errors.DB_NOT_AVAILABLE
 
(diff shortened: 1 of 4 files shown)


api/app/responses/api.py
```diff
@@ -82,7 +83,41 @@ def post(self):
                 db.session.add(answer)
             db.session.commit()
 
-            return {'id': response.id, 'user_id': user_id}, 201  # 201 is 'CREATED' status code
+            try:
```
avishkar58: Can we refactor this into another method (eg send_confirmation) and call it from the both POST and PUT, whenever response.is_submitted is TRUE (ie. we only send a confirmation mail when the user submits their response, not when they just save it for later).

api/app/responses/api.py
```diff
@@ -82,7 +83,41 @@ def post(self):
...
+                user = db.session.query(AppUser).filter(AppUser.id==g.current_user['id']).first()
+                subject = strings.build_response_email_subject(user.user_title, user.firstname, user.lastname)
+
+                answers = db.session.query(Answer).filter(Answer.response_id == response.id).all()
+                if answers is None:
+                    LOGGER.warn('Found no answers associated with response with id {response_id}'.format(response_id=response.id))
```
avishkar58: Thanks for adding logging to this controller! I forgot about that :)

api/app/utils/emailer.py
```diff
@@ -22,43 +22,49 @@ def send_mail(recipient, subject, body_text, body_html='', charset='UTF-8', mail
...
-            LOGGER.info("Building Mimemultipart msg")
-            msg = MIMEMultipart()
-            LOGGER.info("Setting subject %s"%(subject,))
-            msg['Subject'] = subject
-            LOGGER.info("Setting Sender name : %s and Sender email %s"%(SMTP_SENDER_NAME, SMTP_SENDER_EMAIL))
-            msg['From'] = email.utils.formataddr((SMTP_SENDER_NAME, SMTP_SENDER_EMAIL))
-            LOGGER.info("Setting recipient %s"%(recipient))
-            msg['To'] = recipient
-            #msg.add_header('X-SES-CONFIGURATION-SET', configuration_set)
+    if (not DEBUG):
```
avishkar58: Normally we'd do this by having an AmazonMailer and a MockMailer, both implementing the common send_mail interface and pass the appropriate one to the controllers (dependency injection), but that's a fair bit more complex, so this is fine for now!

api/app/responses/api.py
Carvalho96v2: Aha, gotcha :)

api/app/utils/emailer.py
Carvalho96v2: How would that piece of logic work? Where would you pass the mailer instance into the controller, and how would you know which one to pass?

avishkar58:
Here's an article that describes how it can be done in Flask

http://christophergs.github.io/python/2018/09/25/elegant-flask-apis-pt-1/

But rather leave what you have for now - we can always refactor later if we
have more cases warranting the extra complexity

api/app/utils/emailer.py
nischalhp: Just adding to @avishkar58 thought process, the dependency injection can happen using various things, for example as simple as environment variables, where we can base the injection based on a value being set vs unset and inject the appropriate controller. We do not need it at this point in time, but its something that can be done later.

Carvalho96v2: @avishkar58 @nischalhp Great, thanks guys! Will look definitely into it :)

Issue #62, opened 2019-02-19
Add descriptions to Create Account form


avishkar58:
Add more descriptions to the create account form, take inspiration from last year's form

Carvalho96v2:
I can take this

PR #80, opened 2019-02-25
Added descriptions to key validation fields as tool tips 

Carvalho96v2:
Added descriptions to fields in create account screen through tool tips on their respective labels.

Issue #62


KaleabTessera:
Great work!

In future, we need to relook at where we place the tooltips. It took me a while to find them on the header :)

Issue #75, opened 2019-02-24
Implement "Save for later" in application form front-end


avishkar58:
Depends on #66 

Implement a "save for later" button on the application form. This entails the following:
- Pre-populate the application form with the values from any existing response received from the back-end
- Add a "save for later" button to the application form UI. This button should trigger a POST or a PUT to the back-end to save the response (post if its a new response, put if it's an update to an existing response). The is_submitted flag should be set to False when using this button. 
- Ignore validation (is_required and validation_regex) when saving for later, but apply validation when submitting

Carvalho96v2:
I can take this

PR #85, opened 2019-02-28
Prepopulate form answers and added Save for Later button

Carvalho96v2:
- Checked for existing responses and used their answers to prepopulate the relevant fields
- Added "Save for Later" button, orange when there are detected unsaved changes, green when up to date
- Added integration with response api' put method as a service
- Amended the "handleSubmit" method to make use of the put method when submitting a pre-existing response

First time working with React :D please let me know if I broke any patterns :)

This closes issue #75


webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -276,25 +294,68 @@ class ApplicationForm extends Component {
...
+                  });
+            });
+        } else {
+            applicationFormService.updateResponse(this.state.responseId, this.state.formSpec.id, false, this.state.answers).then(resp=> {
+                let saveError = resp.response_id === null;
+                this.setState({
+                    isError: saveError,
+                    errorMessage: resp.message,
+                    isSubmitted: resp.is_submitted,
+                    unsaved_changes : false
```
KaleabTessera: No change on isLoading or submittedTimestamp? Maybe consider creating a re-usable function that changes the state after update response, might make it easier to not forgot fields.

Carvalho96v2: Hey, yeah I didn't update submittedTimestamp because this would just be updating the existing response, and not submitting it (setting is_submitted to true), so there would be nothing to update it to, if that makes sense

Also, why make a change to isLoading? :)

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -97,22 +98,30 @@ class FieldEditor extends React.Component {
...
-                {this.formControl(this.props.question)}
+                {this.formControl(this.props.question, this.props.answer)}
             </div>
         )
     }
   }
 
 function Section (props) {
     let questions = props.questions && props.questions.slice().sort((a, b) => a.order - b.order);
+    let answers = props.questions && props.answers.slice().sort((a, b) => a.order - b.order);
```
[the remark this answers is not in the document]
Carvalho96v2: Whoops, good spot

webapp/src/pages/applicationForm/components/ApplicationForm.js
avishkar58: I do think you should update submittedTimestamp here, that way the business logic you describe here is all implemented on the server side

avishkar58: Regarding isLoading - clicking the save button should set isLoading to true immediately and then set it to false after receiving the response from the server.

Issue #12, opened 2019-01-13
Implement PasswordReset functionality


avishkar58:
Implement password reset functionality in the front-end - this should call the following back-end API endpoints:
/api/v1/password-reset/request
/api/v1/password-reset/confirm'

This depends on #13 which will send an email to the user asking them to confirm their password reset.

**Definition of Done**
This task is done when:
* A new component has been created to request a password reset
* A password reset page has been created using the above component
* A new component has been created to confirm a password reset (set a new password)
* A confirm password reset page has been created using the above component
* Basic unit tests exist to test the rendering of the new pages
* A manual integration tests confirms that the functionality works as expected

PR #86, opened 2019-02-28
Feature/reset password frontend

hbarnardt:


hbarnardt:
This closes #12

webapp/src/services/events/events.service.js
```diff
@@ -12,6 +12,6 @@ export function getEvents() {
         })
         .catch(function (error) {
             // handle error
-            return { error: error };
+            return [], { error: error };
```
KaleabTessera: Return [] ?

KaleabTessera:
Thanks for all the great work!! @hbarnardt

webapp/src/services/events/events.service.js
hbarnardt: It was trying to render a with a null, i have since fixed the backend so that that won't happen. I will move the check to the controller

PR #109, opened 2019-03-05
Tweaks

avishkar58:
Bug fixes - this should hopefully be the last update needed to be able to successfully create an account, login, submit and application on our test instance.

### DIFF
--- a/api/app/users/api.py
+++ b/api/app/users/api.py
@@ -48,6 +48,9 @@ def user_info(user):
     }
 
 
+def get_baobab_host():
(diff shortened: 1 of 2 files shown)


arunavkonwar:
The reset password email presently leads to double back slash(after appspot.com) https://baobab.appspot.com//resetPassword?resetToken

instead of single back slashed   https://baobab.appspot.com/resetPassword?resetToken

avishkar58:
Hi @arunavkonwar - this PR should fix it for both the verifyEmail and resetPassword links, thanks!

PR #122, opened 2019-03-06
Bug fixes

avishkar58:
- Removed delete functionality from user profile
- Replaced flatMap with more browser compatible solution.

This PR should fix the issues @danielamassiceti  was experiencing on test

### DIFF
--- a/webapp/src/pages/profile/components/ProfileForm.js
+++ b/webapp/src/pages/profile/components/ProfileForm.js
@@ -3,6 +3,7 @@ import { userService } from "../../../services/user";
 import { withRouter } from "react-router";
 import FormTextBox from "../../../components/form/FromTextBox";
 import FormSelect from "../../../components/form/FormSelect";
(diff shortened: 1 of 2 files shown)


webapp/src/pages/profile/components/ProfileForm.js
```diff
@@ -373,12 +376,12 @@ class ProfileForm extends Component {
...
                                 class="btn btn-primary Button"
                                 disabled={loading}
-                                onClick={this.deleteAccount}
+                                onClick={()=>this.setState({confirmResetVisible: true})}
                             >
                                 Reset password
                             </button>
                         </div>
-                        <div class={commonColClassName}>
+                        {/* <div class={commonColClassName}>
```
KaleabTessera: Is the commented code here on purpose?

avishkar58: Oops! fixed

PR #139, opened 2019-03-09
File upload Frontend

### LINKED ISSUE #42: Implement file upload in application form front end
Implement file-upload functionality in the application form front-end. Do this by sending a separate multipart/form-data POST request when saving the response to a dummy back-end REST endpoint (eg <URL>/file). The back-end should respond with a file-id, this file-id should then be included as the 'value' for the 'Answer' corresponding to the question asking for a file-upload. 

(Chat to @avishkar58 if this doesn't make sense :) )

avishkar58:
Implements front-end functionality for file upload in application form. 
Modified backend to conditionally not use credentials to test on GCP
Resolves #42

### DIFF
--- a/webapp/src/components/form/FormFileUpload.js
+++ b/webapp/src/components/form/FormFileUpload.js
@@ -3,8 +3,6 @@ import FormGroup from "./FormGroup";
 import FormToolTip from "./FormToolTip";
 import "./Style.css";
 
-// uploading, uploadPercentComplete, uploadFile, uploaded
-
 class FormFileUpload extends React.Component {
   shouldDisplayError = () => {
     return this.props.showError && this.props.errorText !== "";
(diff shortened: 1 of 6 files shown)


webapp/src/components/form/FormFileUpload.js
```diff
@@ -0,0 +1,74 @@
+import React, { Component } from "react";
+import FormGroup from "./FormGroup";
+import FormToolTip from "./FormToolTip";
+import "./Style.css";
+
+// uploading, uploadPercentComplete, uploadFile, uploaded
```
KaleabTessera: Think you accidentally included a comment :)

avishkar58: Oh right - thanks :)

PR #150, opened 2019-03-11
Minor ui enhancements

KaleabTessera:
Changes:
- Made navbar collapse once an option has been clicked on mobile.
- Added generic error page if any page throws an exception.


webapp/src/App.js
```diff
@@ -67,47 +72,90 @@ class App extends Component {
...
+                      Apply
+                    </NavLink>
+                  </li>
+                )}
               </ul>
-              <UserDropdown logout={this.refreshUser} user={this.state.user} />
+              <UserDropdown
+                logout={this.refreshUser}
+                user={this.state.user}
+                onClick={this.toggleMenu}
```
avishkar58: Haven't tested this, but how will this affect the dropdown when the user is logged in?

KaleabTessera: Good point. Will test this tonight.

KaleabTessera: Yep, works as expected.

PR #152, opened 2019-03-12
Resend verification email backend

jaderabbit:
@KaleabTessera Here is the backend for `resend_verification`. I'm likely going to make a hash of the frontend. Not sure where to put the button either. 

Thoughts?


avishkar58:
Already merged this and realised that the ResendVerification API has auth_required attribute, but the user can't authenticate themselves because their email hasn't been verified :) I'm doing the front-end quickly now, so will fix as part of that

jaderabbit:
@avishkar58 Oh duh! sarie. That was stupid.

PR #155, opened 2019-03-13
Feature/google analytics

KaleabTessera:


KaleabTessera:
Looking at why test failed.

KaleabTessera:
Tests sorted.

KaleabTessera:
Re-opening, sorted out test stuff for react-ga.

webapp/src/App.js
```diff
@@ -154,7 +166,7 @@ class App extends Component {
             <div class="container">
               <p>
                 Baobab, © 2019 |{" "}
-                <a href="www.deeplearningindaba.com">Deep Learning Indaba</a>
+                <a href="https://www.deeplearningindaba.com">Deep Learning Indaba</a>
```
avishkar58: Curious why the https was added? the Indaba website doesn't have an SSL certificate, so although https "works", the user will get a scary certificate error.

jaderabbit: Sorry I merged :woman_facepalming:

KaleabTessera: Oh sorry, basically the link was appending and not redirecting (because there was no http/https in the prefix) so I added https. I will change to http 👍

PR #158, opened 2019-03-13
Changed https to http.

KaleabTessera:


KaleabTessera:
@avishkar58 Please have another look, I just changes some app.yaml settings to make sure all traffic is routed to https version of Baobab.

avishkar58:
Ah that's awesome - thanks!

PR #172, opened 2019-03-17
Logging phase 1 final

greg-newman:


api/app/responses/api.py
```diff
@@ -52,24 +56,32 @@ def get(self):
...
             if not form:
+                LOGGER.warn("Form not found for event_id: {}".format(args['event_id']))
                 return errors.FORM_NOT_FOUND
             
             # Get the latest response (note there may be older withdrawn responses)
             response = db.session.query(Response).filter(
                 Response.application_form_id == form.id, Response.user_id == g.current_user['id']
                 ).order_by(Response.started_timestamp.desc()).first()
             if not response:
+                LOGGER.warn("Response not found for event_id: {}".format(args['event_id']))
```
avishkar58: Lets make this one an info/debug, because the Application Form tries to get a response every time and we expect many/most of those times to have missing responses, so don't want to generate a bunch of spurious warnings

greg-newman: Ah sure thing, didn't realise it would pollute the logs like that.

Issue #174, opened 2019-03-21
Create Review data models


avishkar58:
Create the data models for the review process as per the [review process spec](https://docs.google.com/document/d/1e4MWBbyqHTXhaL-u1nPozxzCJj36sM1qjuW1MNOtj3U/edit?usp=sharing). 

This involves the creation of the following models:
- ReviewForm
- ReviewQuestion
- ReviewResponse
- ReviewScore
- ResponseReviewer

Create a migration to apply the changes to the database

**Definition of Done**
This task is done when

The relevant models exist in api/app/review/models.py
A new migration exists that creates the necessary tables in the database
The migration can be applied successfully
Unit tests are optional for this task

PR #187, opened 2019-03-23
Feature/review models

amritpurshotam:
Resolves issue #174 

To get the migrations to pick up, I had to make a skeleton review API controller which is why that's there.


api/app/reviews/models.py
```diff
@@ -0,0 +1,103 @@
...
+        self.deadline = deadline
+
+    def close(self):
+        self.is_open = False
+
+
+class ReviewQuestion(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    review_form_id = db.Column(db.Integer(), db.ForeignKey('review_form.id'), nullable=False)
+    question_id = db.Column(db.Integer(), db.ForeignKey('question.id'), nullable=False)
```
avishkar58: This should be nullable, because some ReviewQuestions will not be linked to an application form question

api/app/reviews/models.py
```diff
@@ -0,0 +1,103 @@
...
+    description = db.Column(db.String(), nullable=True)
+    headline = db.Column(db.String(), nullable=False)
```
avishkar58: This should also be nullable please

api/app/reviews/models.py
```diff
@@ -0,0 +1,103 @@
...
+        self.reviewer_user_id = reviewer_user_id
+        self.response_id = response_id
+        self.submitted_timestamp = datetime.now()
+
+
+class ReviewScore(db.Model):
+    id = db.Column(db.Integer(), primary_key=True)
+    review_response_id = db.Column(db.Integer(), db.ForeignKey('review_response.id'), nullable=False)
+    review_question_id = db.Column(db.Integer(), db.ForeignKey('review_question.id'), nullable=False)
+    value = db.Column(db.Float(), nullable=False)
```
avishkar58: We might need a string type here since we also want to be able to capture free-text comments from the reviewers - or we could add a comment field directly to the ReviewResponse model - thoughts?

api/app/reviews/models.py
amritpurshotam: Done.

api/app/reviews/models.py
amritpurshotam: Done.

api/app/reviews/models.py
amritpurshotam: Yes of course, I think the free-text makes sense. Just re-read the spec now. Don't think the extra comment field will be necessary if there will be one score per response (as opposed to a score and a comment).

Issue #181, opened 2019-03-23
Implement ReviewForm controller


avishkar58:
Create a ReviewForm controller with a GET method that accepts an "event_id" and "skip" parameter. The method should do the following:
- Get the review form for the event
- Determine which reviews are yet to be completed for the logged-in user (reviewer), by querying the ResponseReviewer and ReviewResponse tables
- Optionally Skip some reviews based on the skip parameter to determine the current review (the skip should be maxed at the number of reviews available) 
- Get the AppUser, Response and Answers related to the current review 

The method should return the following:
- The ReviewForm for the event
- List of ReviewQuestions for the ReviewForm
- The Answer value for each ReviewQuestion if applicable (remember to resolve the label in the case of a "multi-choice" question) 
- The following fields from the AppUser being reviewed: Country of Nationality, Country of Residence, Affiliation, Department, User Category   (don't return any other fields!) 

- Return an appropriate status code (eg 404?) if there are no more reviews left. 

**Definition of Done**
- New ReviewForm controller has been created in api/app/reviews/api.py
- Tests exist for the functionalitiy

amritpurshotam:
> The Answer value for each ReviewQuestion if applicable (remember to resolve the label in the case of a "multi-choice" question)

@avishkar58 Just going through this ticket again. I'm unsure what you mean by resolving the label. I don't think I did this. I'm just showing the `value` field.

avishkar58:
Hey @amritpurshotam 
If you link to a question like "Did you attend a previous Indaba", the value in the answer table will be something like "indaba-2017", but in the question table, the options field has something like: 
[
{label: "Yes I attended the 2017 Indaba", value: "indaba-2017"},
{label: Yes, I attended both Indabas", value: "indaba-2017-2018"},
]

So in this case we would like to map the value to the label and have "Yes I attended the 2017 Indaba" returned instead of the "indaba-2017" 

Hope this makes sense?

amritpurshotam:
@avishkar58 Sorry I missed this and we've already merged the branch but I can make a new one for this.

I think I understand. How will I know how to identify these kinds of questions? Or do you want to specifically handle this question?

And is this what you want the response (i took out the fields that don't matter for this example)?

```
{
    "response": {
        "answers": [
            {
                "id": 1,
                "question": "Did you attend a previous Indaba?",
                "question_id": 1,
                "value": "Yes, I attended both Indabas"
            }
        ]
    },
    "review_form": {}
[...]

PR #189, opened 2019-03-24
Feature/review controller

amritpurshotam:
Resolves issue #181 

Just a heads up to make sure to expand the tests.py file since it's hidden at the moment for being too big.

Let me know if there's any other test cases you can think of


api/app/reviews/api.py
```diff
@@ -1,7 +1,106 @@
...
+        event_id = args['event_id']
+        
+        review_form = db.session.query(ReviewForm)\
+                        .join(ApplicationForm, ApplicationForm.id==ReviewForm.application_form_id)\
+                        .filter_by(event_id=event_id)\
+                        .first()
+        if review_form is None:
+            return EVENT_NOT_FOUND
+
+        skip = 0 if args['skip'] is None else args['skip']
```
avishkar58: FYI (no need to change): This is equivalent to the more concise  ```skip = args['skip'] or 0```

api/app/reviews/api.py
```diff
@@ -1,7 +1,121 @@
...
+                        .filter_by(id=None)\
+                        .all()[0][0]
+
+        skip = args['skip'] or 0
+        skip = skip if skip < reviews_remaining_count else reviews_remaining_count-1
+
+        response = db.session.query(Response)\
+                        .filter_by(is_withdrawn=False, application_form_id=review_form.application_form_id, is_submitted=True)\
+                        .join(ResponseReviewer)\
+                        .filter_by(user_id=g.current_user['id'])\
```
avishkar58: Shouldn't this be reviewer_user_id ?

amritpurshotam: The spec has it as `user_id` but I can update it. `reviewer_user_id` is more descriptive and will avoid confusion.

amritpurshotam: Done.

api/app/reviews/api.py
```diff
@@ -1,7 +1,125 @@
...
+        reviews_remaining_count = db.session.query(func.count(ResponseReviewer.id))\
+                        .filter_by(reviewer_user_id=g.current_user['id'])\
+                        .join(Response)\
+                        .filter_by(is_withdrawn=False, application_form_id=review_form.application_form_id, is_submitted=True)\
+                        .outerjoin(ReviewResponse, and_(ReviewResponse.response_id==ResponseReviewer.response_id, ReviewResponse.reviewer_user_id==g.current_user['id']))\
+                        .filter_by(id=None)\
+                        .all()[0][0]
+
+        skip = args['skip'] or 0
+        skip = skip if skip >= 0 else 0
```
KaleabTessera: The two lines above are quite hard to read. Can we maybe refactor it a bit to make it more readable.
Example:
```
if args['skip'] is None:
	skip = 0
else:
	if args['skip'] >= 0:
		skip = args['skip']
	else:	
		skip = 0
```
Not sure if the logic above 100% matches what you were trying to say but yeah it makes it more readable.

amritpurshotam: @KaleabTessera Yeah it was a bit confusing. I think your way is what I was trying to achieve but I just shifted it around so that there's no nesting. Let me know if you're happy with what I have now

amritpurshotam:
@KaleabTessera @avishkar58 I think I've finished this ticket now. Note that once a reviewer has completed all their reviews I'm still returning the standard response but I've added an extra field called `reviews_remaining_count` and this should be zero. The front-end can then use this as a check to figure out what to display on the screen. 

The reason I chose this method is that having no reviews left is a valid use case so I didn't think an error code like 404 would be appropriate and will also result in less noise in the logs.

As an added bonus, the front-end can also display this value to the reviewer so they can see how many they still have to do.

Let me know if you're happy with this change?

amritpurshotam:
```
{
    "response": {
        "answers": [
            {
                "id": 1,
                "question": "Why is attending the Deep Learning Indaba 2019 important to you?",
                "question_id": 1,
                "value": "test1"
            },
            {
                "id": 2,
                "question": "How will you share what you have learnt after the Indaba?",
                "question_id": 2,
                "value": "test2"
            }
        ],
        "application_form_id": 1,
        "id": 4,
        "is_submitted": true,
        "is_withdrawn": false,
        "started_timestamp": "1984-12-12T00:00:00",
        "submitted_timestamp": "1984-12-12T00:00:00",
        "user_id": 1,
        "withdrawn_timestamp": null
    },
    "review_form": {
        "application_form_id": 1,
        "deadline": "2019-04-30T00:00:00",
        "id": 1,
        "is_open": true,
        "review_questions": [
            {
                "description": "something",
                "headline": "headline",
                "id": 1,
                "is_required": true,
                "options": null,
                "order": 1,
                "placeholder": "placeholder\n",
                "question_id": 1,
                "type": "type\n",
                "validation_regex": null,
                "validation_text": null,
                "weight": 50
            },
            {
                "description": "soemthing2",
                "headline": "headline2",
                "id": 2,
                "is_required": false,
                "options": null,
                "order": 2,
                "placeholder": "placeholder2\n",
                "question_id": 2,
                "type": "type2",
                "validation_regex": null,
                "validation_text": null,
                "weight": 50
            }
        ]
    },
    "reviews_remaining_count": 1,
    "user": {
        "affiliation": "Stellenbosch",
        "department": "Computer Science",
        "nationality_country": "Bahamas, The",
        "residence_country": "Bahrain",
        "user_category": "4th year/honours student"
    }
}
```

api/app/reviews/tests.py
```diff
@@ -0,0 +1,441 @@
...
+    def test_one_reviewer_one_candidate(self):
+        self.seed_static_data()
+        self.setup_one_reviewer_one_candidate()
+        header = self.get_auth_header_for('r1@r.com')
+        params = {'event_id': 1}
+
+        response = self.app.get('/api/v1/review', headers=header, data=params)
+        data = json.loads(response.data)
+
+        assert data['reviews_remaining_count'] == 1
```
avishkar58: It's better to use self.assertEqual() and other methods on the TestCase class, the advantage is that those methods have sensible error messages when the assertion fails.

api/app/reviews/api.py
```diff
@@ -1,7 +1,134 @@
...
+                        .filter_by(event_id=event_id)\
+                        .first()
+        if review_form is None:
+            return EVENT_NOT_FOUND
+
+        reviews_remaining_count = db.session.query(func.count(ResponseReviewer.id))\
+                        .filter_by(reviewer_user_id=g.current_user['id'])\
+                        .join(Response)\
+                        .filter_by(is_withdrawn=False, application_form_id=review_form.application_form_id, is_submitted=True)\
+                        .outerjoin(ReviewResponse, and_(ReviewResponse.response_id==ResponseReviewer.response_id, ReviewResponse.reviewer_user_id==g.current_user['id']))\
```
KaleabTessera: Would we need to do this:
```
ReviewResponse.reviewer_user_id==g.current_user['id'])
```

If we have done,
```
.filter_by(reviewer_user_id=g.current_user['id'])
``` 
above?

api/app/reviews/api.py
```diff
@@ -1,7 +1,134 @@
...
+        if skip is None:
+            skip = 0
+
+        if skip < 0:
+            skip = 0
+
+        if reviews_remaining_count == 0:
+            skip = 0
+        elif skip >= reviews_remaining_count:
+            skip = reviews_remaining_count - 1
```
KaleabTessera: Why set `skip = reviews_remaining_count - 1` and not `skip = reviews_remaining_count` (if Skip > reviews_remaining_count)?

api/app/reviews/api.py
```diff
@@ -1,7 +1,134 @@
...
+        response = db.session.query(Response)\
+                        .filter_by(is_withdrawn=False, application_form_id=review_form.application_form_id, is_submitted=True)\
+                        .join(ResponseReviewer)\
+                        .filter_by(reviewer_user_id=g.current_user['id'])\
+                        .outerjoin(ReviewResponse, and_(ReviewResponse.response_id==ResponseReviewer.response_id, ReviewResponse.reviewer_user_id==g.current_user['id']))\
+                        .filter_by(id=None)\
+                        .order_by(ResponseReviewer.response_id)\
+                        .offset(skip)\
+                        .limit(1)\
+                        .first()
```
KaleabTessera: Do we need `limit(1)` and `.first()` ?

KaleabTessera:
Amazing work @amritpurshotam ! It took me a while to review this, I can only imagine how long it took to implement. Thanks a lot!

I just left some minor questions on a couple lines.

api/app/reviews/tests.py
amritpurshotam: Done.

(When my tests were failing, I was wondering why it wasn't being helpful at all haha)

api/app/reviews/api.py
amritpurshotam: Say `reviews_remaining_count=2`, if you skip 1, you go on to the second review. If you skip 2, you get nothing back. So the subtraction is to get you to the last review.

api/app/reviews/api.py
amritpurshotam: We definitely need the `.first()` since that actually executes the preceding query but you're right the `.limit(1)` is redundant.

api/app/reviews/api.py
amritpurshotam: So both are actually required.

The `.filter_by` is to limit the ResponseReviewers to just the logged in user. The reason is that we're doing an outer join later which means all the ResponseReviewers are preserved and so every single one of those are counted (instead of just the logged in user's).

The `ReviewResponse.reviewer_user_id==g.current_user['id'])` in the outer join is to make sure that if two reviewers are assigned to a response and one of them completes a review, the query will still bring back the response for the second reviewer. (I'm struggling to explain this better so let me know if it doesn't make sense. This one was a bit confusing for me to figure out.)

amritpurshotam:
@KaleabTessera @avishkar58 Thanks for reviewing this quite comprehensively. I know there was a quite a bit to take in. I think I've addressed all the comments now. Let me know if there's anything else.

api/app/reviews/api.py
KaleabTessera: No it makes sense thanks.

Issue #175, opened 2019-03-21
Populate Review Form


avishkar58:
Create a migration that populates seed data in the ReviewForm and ReviewQuestion tables. The data required is as follows:

- A review form for the current ApplicationForm that exists in the DB (with id 1), which is open and has a deadline set to "10 May 2019"
The ReviewForm should have the following ReviewQuestions (in this order):

- Headline: "Encourage field of study?"; description: "Is the applicant from a field of study we want to encourage? ie outside of CS/Maths/Stats?"; type="checkbox"; 

- Link to "Why is attending the Deep Learning Indaba 2019 important to you?"; type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: true;

- Link to "How will you share what you have learnt after the Indaba?"; type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: true;

- Link to "Share with us a favourite machine learning resource you use: a paper, blog post, algorithm, result, or finding. Tell us why."; type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: true;

- Link to "Are you or have you been a tutor for any relevant course, or part of any machine learning or data science society or meetup? If yes, give details."; description: "Based on this answer, do you believe the applicant has been a tutor for a relevant course?"; type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: true;

- Link to "Have you worked on a project that uses machine learning? If yes, give a short description."; type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: true;

- Link to "Did you attend the 2017 or 2018 Indaba"; type="information"; 

- Link to "If you attended a previous Indaba, tell us how it has helped you grow or how you have used what you have learnt." type="multi-choice"; options=Poor: 0, Below Average: 1, Average: 2, Above Average: 3, Excellent: 4; is_required: false;

- Link to "Would you like to be considered for a travel award?"; type="information"

- Link to "If you do not receive a travel award from us, will you still be able to attend?"; type="information"

[...]

PR #192, opened 2019-03-26
update/review-seed-data

amritpurshotam:
Resolves issue #175


api/migrations/versions/1349a2c924f4_.py
```diff
@@ -0,0 +1,281 @@
...
+                    {'label': 'Yes', 'value': 2}
+                ],
+                'is_required': True,
+                'order': 14,
+                'weight': 0
+            }
+        ]
+    )
+
+    # ### end Alembic commands ###
```
avishkar58: Since you're using bulk_insert, please (re)set the ID sequence on these tables to prevent issues with future inserts:

op.execute("""SELECT setval('review_question_id_seq', (SELECT max(id) FROM review_question));""")
op.execute("""SELECT setval('review_form_id_seq', (SELECT max(id) FROM review_form));""")

(also do this in the downgrade)

amritpurshotam: Cool done.

Issue #180, opened 2019-03-23
Redirect user to login if auth fails


avishkar58:
Authorization might fail due to an expired token, in which case the API will return a 401 status code. At the moment, many of our pages will just display an error to the user if  this happens. Instead, we should redirect the user to the login page.  Create some common functionality in the front-end services (eg in base.service.js) to implement this.

KaleabTessera:
@avishkar58 
So I have implemented and tested this and it intercepts errors with a 401 and redirects to login as expected.

However, there are other errors which return a 401, for example: 
- EMAIL_VERIFY_CODE_NOT_VALID
- RESET_PASSWORD_CODE_NOT_VALID
- BAD_CREDENTIALS

I am assuming we only want this behaviour for UNAUTHORIZED errors?

In this case I will change the error codes returned for the other errors.

avishkar58:
Thanks! Yeah we definitely only want this for genuine unauthorized errors
:)

KaleabTessera:
Closed by https://github.com/deep-learning-indaba/Baobab/pull/195.

PR #195, opened 2019-03-28
Bugfix/better routing auth failure

KaleabTessera:
Implementation of https://github.com/deep-learning-indaba/Baobab/issues/180 .


api/app/users/tests.py
```diff
@@ -274,7 +276,7 @@ def test_password_reset(self):
             'code': "bad code",
             'password': 'abc123'
         })
-        assert response.status_code == 401
+        assert response.status_code == 418
```
avishkar58: LOL

webapp/src/services/base.service.js
```diff
@@ -1,10 +1,31 @@
...
+  },
+  function(error) {
+    if (
+      error &&
+      error.response &&
+      error.response.status === 401 &&
+      error.response.data &&
+      error.response.data.type === "UNAUTHORIZED"
+    ) {
+      localStorage.removeItem("user");
```
avishkar58: We also need to signal (somehow?) to the UserDropdown component that the user has been logged out...

KaleabTessera: Currently working on this, it is a little more complicated than I expected.

KaleabTessera: So I sorted this issue. 

I tried adding a listener to localstorage changes and then this updates user in the User.js, but the listener wasn't been triggered from changes to localstorage from axios. 

I found a work around, but just always checking localstorage and comparing it to the props received.

Quite a trick problem, redux was made to solve these kinds of issues, with communicating state changes to a wide variety of components, but it works so it fine for now :)

Issue #205, opened 2019-03-31
Create ReviewAssignment controller


avishkar58:
Create a new ReviewAssignment controller in api/app/reviews/api.py with a POST request that accepts 3 parameters:
- event_id
- reviewer_user_email
- num_reviews

The method should return a FORBIDDEN error if the logged-in user is not an event-admin or system admin 

The method should return a NOT_FOUND error if the reviewer_user_email does not correspond to a user in AppUser

The method should add the "reviewer" role to the EventRole table if the review_user does not already have this role for the event. 

The method should query the Response and ResponseReviewer tables for responses that don't already have 3 reviewers assigned to them. 

From the pool of available responses to review, the method should select a maximum of num_reviews responses (capped at the number available of course), ensuring that the reviewer does not get their own response to review, and then create entries in the ResponseReviewer table to assign the reviewer to the response. 

**Definition of Done**
- New controller created in api/app/reviews/api.py
- Tests added for new functionality

amritpurshotam:
@avishkar58 In the case of a the request being sent twice for the same reviewer how many reviews should they be assigned? Does the num_reviews that get assigned reset to the num_reviews argument or does  it include what they have already had assigned to them i.e. num responses to be assigned =  num_reviews - number assigned already

avishkar58:
Hi Amrit, good question! I think our usage will be something along the lines of "lets assign X *more* reviews to this reviewer", so ignore the count of what they're currently assigned and just assign num_reviews MORE to them (taking care not to assign the same candidate to the same reviewer more than once of course) - hope this makes sense?

amritpurshotam:
Cool thanks

PR #212, opened 2019-04-03
Feature/reviewassignment

amritpurshotam:
Resolves issue #205

### DIFF
--- a/api/app/reviews/api.py
+++ b/api/app/reviews/api.py
@@ -252,8 +252,9 @@ class ReviewAssignmentAPI(PostReviewAssignmentMixin, restful.Resource):
         reviewer_user_email = args['reviewer_user_email']
         num_reviews = args['num_reviews']
 
+        is_user_admin = db.session.query(AppUser).get(user_id).is_admin
         current_user_roles = self.get_roles(user_id, event_id)
-        if 'event_admin' not in current_user_roles and 'system_admin' not in current_user_roles:
+        if 'admin' not in current_user_roles and not is_user_admin:
             return FORBIDDEN
         
         reviewer_user = self.get_reviewer_user(reviewer_user_email)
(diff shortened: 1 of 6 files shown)


api/app/reviews/api.py
```diff
@@ -238,4 +240,57 @@ def validate_scores(self, scores):
...
+    @auth_required
+    def post(self):
+        args = self.post_req_parser.parse_args()
+        user_id = g.current_user['id']
+        event_id = args['event_id']
+        reviewer_user_email = args['reviewer_user_email']
+        num_reviews = args['num_reviews']
+
+        current_user_roles = self.get_roles(user_id, event_id)
+        if 'event_admin' not in current_user_roles and 'system_admin' not in current_user_roles:
```
amritpurshotam: Is `event_admin` and `system_admin` the correct the role names?

api/app/reviews/api.py
```diff
@@ -238,4 +240,57 @@ def validate_scores(self, scores):
...
+            return FORBIDDEN
+        
+        reviewer_user = self.get_reviewer_user(reviewer_user_email)
+        if reviewer_user is None:
+            return USER_NOT_FOUND
```
amritpurshotam: Note this was an existing error message but it's returning 409 CONFLICT currently? Should I update this to be a 404? Not sure of the knock on effects on the front end though.

amritpurshotam: @KaleabTessera I had a look at the `user.service.js` file and it looks like any error code is handled the same way. So changing it shouldn't affect the front end right?

api/app/reviews/api.py
avishkar58: "System admin" isn't a role, it's when the is_admin flag on the user object is True

avishkar58: And the event admin role is just "admin" (since it's already in the EventRole table, the event_ prefix is redundant)

api/app/reviews/api.py
avishkar58: I think it should be fine to change to 404

api/app/reviews/api.py
amritpurshotam: Ah I see now. Thanks. Done.

api/app/reviews/api.py
amritpurshotam: Done.

PR #226, opened 2019-04-08
Stopped prev versions in test.

KaleabTessera:


KaleabTessera:
Hmm that is weird that it does this only in Test :( 

So once merged, it doesn't deploy? Does it timeout when building the containers? Maybe we can just inc the timeout?

jaderabbit:
@KaleabTessera It does it in prod as well. I actually have that problem right now :P

jaderabbit:
@KaleabTessera Check out this: https://github.com/deep-learning-indaba/Baobab/pull/230

jaderabbit:
Nevermind. It did eventually deploy. I must be going mad

jaderabbit:
@KaleabTessera So it literally ignored the field which is as per the link in the PR above. :/ Annoying gcloud

PR #227, opened 2019-04-08
Attempting to add a button to send reminders

jaderabbit:
- Mah first React PR. 
- Has a single button to send reminders to both unsubmitted and not started. Can't see the use case in separating the requests. 
- Excuse the roughness, still learning. let me know where I can improve :)

### DIFF
--- a/webapp/src/pages/eventStats/components/EventStatsComponent.js
+++ b/webapp/src/pages/eventStats/components/EventStatsComponent.js
@@ -32,9 +32,6 @@ class EventStatsComponent extends Component {
 
   handleSubmit = event => {
     event.preventDefault();
-    this.setState({ submitted: true, showErrors: true });
-
-
     this.setState({ buttonLoading: true });
 
     eventStatsService.sendReminderToSubmit(DEFAULT_EVENT_ID).then(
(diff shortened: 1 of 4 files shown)


webapp/src/pages/eventStats/components/EventStatsComponent.js
```diff
@@ -11,23 +11,47 @@ class EventStatsComponent extends Component {
...
 
   componentDidMount() {
     eventStatsService.getStats(DEFAULT_EVENT_ID).then(result=>{
       this.setState({
         loading: false,
+        buttonLoading: false,
         stats: result.stats,
-        error: result.error
+        error: result.error,
+        emailSendStatus: ""
```
avishkar58: It's not necessary to specify the state items that don't change, but it doesn't do any harm either :)

webapp/src/pages/eventStats/components/EventStatsComponent.js
```diff
@@ -11,23 +11,47 @@ class EventStatsComponent extends Component {
...
       });
     });
   }
 
+  handleSubmit = event => {
+    event.preventDefault();
+    this.setState({ submitted: true, showErrors: true });
```
avishkar58: These don't get used anywhere?

Issue #218, opened 2019-04-07
Create ReviewHistory controller


avishkar58:
Create a ReviewHistoryAPI controller in api/app/reviews/api.py that implements a GET method. The method should take an event_id as a parameter and should return a list of reviews that the logged in user has completed. The list should contain the following items from each review:

* review_response_id
* submitted_timestamp
* user.nationality_country
* user.residence_country
* user. affiiliation
* user.department
* user.user_category
* the final verdict score (obtain this by looking for the question with headline "Final Verdict") 
* num_entries (the total number of entries for pagination)

The controller should support paging and sorting by taking as parameters a `page_number`, a `limit` and a `sort-column`. The query should sort by the `sort-column`, skip `page_number*limit` entries and limit to `limit` entries

**Definition of Done**
* New controller created
* Tests exist for new functionality

PR #234, opened 2019-04-11
Resolves ticket #218

DonneStevenson:


DonneStevenson:
Just trying to figure out the issue with the line endings so the diffs work properly

Issue #224, opened 2019-04-08
Create ReviewAllocation front-end


avishkar58:
Depends on #225 

Create a front-end page that is accessible to an event_admin or system_admin. The page should query the ReviewAllocation controller to get a list of reviewers along with the number of allocated reviews and the number of completed reviews per reviewer for the default event. Display this information in a table with a text-box next to each row allowing the admin to allocate more reviews to a particular reviewer (which sends a POST request to the ReviewAssignment controller with the reviewer's email and number of reviews to allocate). 

Also include a textbox and button outside the table to allow the admin to allocate to new reviewers.

PR #243, opened 2019-04-13
Feature/review assignment frontend

jaderabbit:
Adds #224

 Also included a "Unallocated" count.


webapp/src/pages/reviewAssignment/components/ReviewAssignmentComponent.js
```diff
@@ -0,0 +1,187 @@
+import React, { Component } from "react";
+import { reviewService } from "../../../services/reviews";
+import { withRouter } from "react-router";
+import ReactTable from 'react-table'
+
+import 'react-table/react-table.css'
+
+import FormTextBox from "../../../components/form/FormTextBox";
+
+const DEFAULT_EVENT_ID = process.env.DEFAULT_EVENT_ID || 1;
```
KaleabTessera: Please change this to REACT_APP_DEFAULT_EVENT_ID or it doesn't get passed in as env variable.

webapp/src/pages/reviewAssignment/components/ReviewAssignmentComponent.js
```diff
@@ -0,0 +1,187 @@
...
+    super(props);
+
+    this.state = {
+      loading: true,
+      reviewers: null,
+      error: "",
+      newReviewerEmail: "",
+      reviewSummary: {}
+    };
+    this.renderEditable = this.renderEditable.bind(this);
```
KaleabTessera: This will work, but removing this line and making the function an arrow function will be cleaner.
```renderEditable(cellInfo)``` to ```renderEditable = cellInfo =>```

jaderabbit:
@avishkar58 @KaleabTessera So

1. The code works
2. I was fighting the ORM so needed to hit up some raw postgres
3. The query isn't valid sqlite (kill me now) so the one test doesn't pass, despite the fact the query works beautifully. So I've skipped it. 
4. Will rewrite with ORM later, but this PR needs to land before tomorrow when we start assigning reviews :P 

Hope that is cool!

KaleabTessera:
Yeah the postgres sqlite mismatch has caused me pains in the past too.

No prob. 👍

Issue #242, opened 2019-04-12
Create UserProfile backend


avishkar58:
Create a UserProfileList controller that does the following things: 
- Accepts an event_id as parameter
- Checks that the logged in user is an event admin or system admin
- Returns a list of all users who have a record in the Response table associated with that event (containing all the information they entered when creating their account, except password of course, application date, application status (withdrawn, submitted)) 

Create a UserProfile controller that does the following things (adapt the existing UserAPI if possible)
- Accepts a single user_id
- Ensures the logged in user is an event admin or system admin)
- If the logged in user is a system admin, return all the user's information 
- If the logged in user is an event admin - ensure that the user whose information has been requested applied to attend an event that the logged in user is admin for (eg. If I am an admin for Indaba 2019 and IndabaX SA, I should not be able to access the profile for user Joe Soap who only applied to attend IndabaX Sudan)

PR #251, opened 2019-04-19
Get user profile when system or event admin

amritpurshotam:
Resolves the second part of #242 

Just a quick question around running tests, how do we run a specific test, class, or file? We have 100 tests now and it takes about 3 and half minutes on my machine to run them all. I want to run specific ones while I'm developing them to have a shorter feedback loop. I tried looking at the nosetests docs but the commands for running specific tests don't seem to work. I'm assuming the `docker-compose run web nosetests` is hooking into something else that is actually running the actual nosetest commands but I can't seem to find where that is (and this where I'm thinking the commands from the nosetest docs would actually work).


avishkar58:
I was also trying to figure out how to run specific tests yesterday and was also unsuccessful :( @jaderabbit any ideas?

jaderabbit:
Was a small nightmare but I've figured it/ Turns out the following works:

`docker-compose run web cd api; nosetests -v app.responses` where app.responses would be the package tests you want to run

amritpurshotam:
@jaderabbit Thanks this will definitely come in handy :)

PR #262, opened 2019-04-22
Added ability to flag user category

avishkar58:


api/migrations/versions/358d0e0282bb_add_user_comment.py
```diff
@@ -0,0 +1,37 @@
...
+revision = '358d0e0282bb'
+down_revision = '3dcef2e3c442'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('user_comment',
```
jaderabbit: how did we miss this?! :P

avishkar58: That would be my bad :D

PR #267, opened 2019-04-30
Added more event stats

avishkar58:


jaderabbit:
@avishkar58 Happy for us to skip the test until I get the postgresql tests working

PR #271, opened 2019-05-22
Updated homepage with clear application status

avishkar58:
Updated homepage with clear application status for "complete your application" campaign.


avishkar58:
@jaderabbit  Oops - wasn't meant to include the event stats stuff with the other PR - here is a new clean one :) I am thinking of doing an email campaign as well, but wondering whether to do that with MailChimp since I'm not convinced that our reminder email functionality is actually sending all the emails?

PR #281, opened 2019-05-29
Create an offer Table and do migrations

LindanyRetroRabbit:
Created an Offer Model and did the migrations... Next ill be doing the controllers


LindanyRetroRabbit:
Auto

PR #282, opened 2019-05-29
Create an offer Table and do migrations

LindanyRetroRabbit:
created Offer model with its migrations and working on the controllers in the registration/api.py on the following issue


api/app/registration/models.py
```diff
@@ -2,6 +2,24 @@
...
+
+    id = db.Column(db.Integer(), primary_key=True)
+    user_id = db.Column(db.Integer(),  db.ForeignKey("app_user.id"), nullable=False)
+    event_id = db.Column(db.Integer(), db.ForeignKey("event.id"), nullable=False)
+    offer_date = db.Column(db.DateTime(), nullable=False)
+    expiry_date = db.Column(db.DateTime(), nullable=False)
+    payment_required = db.Column(db.String(50), nullable=False)
+    travel_award = db.Column(db.String(50), nullable=False)
+    accommodation_award = db.Column(db.String(50), nullable=False)
+    accepted = db.Column(db.String(50), nullable=False)
```
avishkar58: accepted, rejected, rejected_reason and updated_at should all be nullable please - these will initially be null when the offer is created and will be populated based on the candidate's response.

avishkar58:
@jaderabbit please have a look for final approval - do you think the 2 migrations are fine, or should it be a single one?

jaderabbit:
Looks good! I'm happy with 2. It's not serious

PR #286, opened 2019-06-03
Card ProfileList-Display:

mthembbu:
*Included nav-links in App.js to view profiles as system / event admin.
 *ProfileServices, ProfileComponents and Tests for each as per project structure.

### DIFF
--- a/webapp/src/App.js
+++ b/webapp/src/App.js
@@ -177,7 +177,7 @@ class App extends Component {
                       className="dropdown-item"
                       onClick={this.toggleMenu}
                     >
-                      Profiles Lists
+                      Applicant Profiles
                     </NavLink>
                     </div>
                   </li>
--- a/webapp/src/App.js
+++ b/webapp/src/App.js
@@ -12,6 +12,7 @@ import Profile from "./pages/profile";
 import ReviewAssignment from "./pages/reviewAssignment";
 import ReviewHistory from "./pages/reviewHistory";
 import EventStats from "./pages/eventStats";
+import ProfileList from "./pages/profileList";
 import { PrivateRoute } from "./components";
 import UserDropdown from "./components/User";
 import InvitedGuests from "./pages/invitedGuests";
@@ -171,13 +172,13 @@ class App extends Component {
                       >
                         Review Assignment
                       </NavLink>
-                      <NavLink
-                        to="/invitedGuests"
-                        className="dropdown-item"
-                        onClick={this.toggleMenu}
-                      >
-                        Invited Guests
-                      </NavLink>
+                    <NavLink
+                      to="/profile-list"
+                      className="dropdown-item"
+                      onClick={this.toggleMenu}
+                    >
+                      Profiles Lists
+                    </NavLink>
                     </div>
                   </li>
                 )}
@@ -279,6 +280,7 @@ class App extends Component {
                 />
                 <PrivateRoute exact path="/review" component={Review} />
                 <PrivateRoute exact path="/review/:id" component={Review} />
+                <PrivateRoute exact path="/profile-list" component={ProfileList} />
               </Switch>
             </div>
           </div>
--- a/webapp/src/App.js
+++ b/webapp/src/App.js
@@ -172,6 +172,13 @@ class App extends Component {
                       >
                         Review Assignment
                       </NavLink>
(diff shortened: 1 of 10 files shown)


mthembbu:
@avishkar58 please see the PR

avishkar58:
Thank you @mthembbu ! Looks good at first glance, will have a closer look tonight :)

mthembbu:
Thank you... I'm humbled. What Jade suggested was that onClick under fullname, the admin should be lead to viewing the clicked profile.

webapp/src/App.js
```diff
@@ -171,13 +172,13 @@ class App extends Component {
                       >
                         Review Assignment
                       </NavLink>
-                      <NavLink
```
KaleabTessera: Was this removed on purpose? :)

avishkar58:
Yes, that is the intention, but to be fair, it's beyond the scope of issue
#263 - will create a new issue and you can work on that too if you like?
:)

mthembbu:
@avishkar58 will be greateful.

webapp/src/App.js
```diff
@@ -171,13 +172,13 @@ class App extends Component {
...
-                        onClick={this.toggleMenu}
-                      >
-                        Invited Guests
-                      </NavLink>
+                    <NavLink
+                      to="/profile-list"
+                      className="dropdown-item"
+                      onClick={this.toggleMenu}
+                    >
+                      Profiles Lists
```
avishkar58: Please change to "Applicant Profiles"

webapp/src/pages/profileList/components/ProfileListComponent.js
```diff
@@ -0,0 +1,91 @@
...
+        }
+}
+
+ isListEmpty(){
+
+    return this.state.List.length === 0;
+ }
+
+ componentDidMount(){ 
+    this.setState({List:profileService.getProfilesList(DEFAULT_EVENT_ID)
```
avishkar58: This outer setState isn't necessary

webapp/src/pages/profileList/components/ProfileListComponent.js
```diff
@@ -0,0 +1,91 @@
...
+                         .then(results => {
+                                this.setState(
+                                    {
+                                        List : results.List,
+                                        isEmpty:this.isListEmpty(),
```
avishkar58: Rather pass results.List to your isListEmpty function, because at the moment when isListEmpty executes, the contents of state is unknown (see https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)

webapp/src/services/profilelist/profilelist.service.js
```diff
@@ -0,0 +1,20 @@
...
+
+function getProfilesList(event_id){
+    return axios
+        .get(baseUrl+"/api/v1/userprofilelist?event_id="+event_id, { headers: authHeader()})
+        .then((response)=>{
+            return { List: response.data, error: ""};
+         })
+        .catch((error)=>{
+            //handling an error, then assign an empty / null array
+            return [], { error:error.message }
```
avishkar58: Shouldn't this be something like {List: [], error: error.message} ? 
Also, it's better to set error like this to ensure that you capture any status message from the server first and then fall back to the generic error message:
```
error: error.response && error.response.data ? error.response.data.message : error.message
```

webapp/src/pages/profileList/components/ProfileListComponent.js
mthembbu: When I console logged my List, it firsts prints undefined, [], then undefined, but it does display the data. But I have removed the setState as per request

webapp/src/pages/profileList/components/ProfileListComponent.js
mthembbu: I understand. Sorted.

mthembbu:
@avishkar58 vplease review

webapp/src/App.js
avishkar58: @mthembbu please see Kale-ab's comment here - we don't think this item should have been removed.

mthembbu: I don't recall making the changes. Will be placing the lines back. Thanks a lot for making me be aware.

PR #288, opened 2019-06-04
Create Registration Form Controllers

soginga:


api/app/registration/api.py
```diff
@@ -0,0 +1,296 @@
...
+        'order': fields.Integer,
+        'registration_questions': fields.List(fields.Nested(registration_question_fields))
+    }
+
+    registration_form_fields = {
+        'event_id': fields.Integer,
+        'registration_sections': fields.List(fields.Nested(registration_section_fields))
+    }
+
+    @marshal_with(registration_form_fields)
```
avishkar58: Since we're reading the offer, this method should have the `@auth_required` decorator

api/app/registration/api.py
```diff
@@ -0,0 +1,296 @@
...
+        try:
+            offer = db.session.query(Offer).filter(Offer.id == offer_id).first()
+            if offer and offer.expiry_date >= datetime.now():
+                registration_form = db.session.query(RegistrationForm).filter(
+                    RegistrationForm.event_id == event_id).first()
+
+                if not registration_form:
+                    return errors.REGISTRATION_FORM_NOT_FOUND
+
+                sections = db.session.query(RegistrationSection).filter(
```
avishkar58: We need to filter the sections according to the flags on the offer. I see this wasn't included in the issue - sorry about that! The relevant text from the spec is:

> Note that the relevant controller that returns a registration form, should compare the show_for_* flags on each RegistrationSection against the relevant Offer. If all of the show_for_* flags are NULL, the RegistrationSection should be included in the response. If any of the show_for_* flags are not null, the RegistrationSection should only be included if the condition `(RegistrationSection.show_for_travel_award AND Offer.travel_award) OR (RegistrationSection.show_for_accomodation_award AND Offer.accomodation_award) OR (RegistrationSection.show_for_payment_required AND Offer.payment_required) is TRUE`.

api/app/registration/api.py
```diff
@@ -0,0 +1,296 @@
...
+                return errors.OFFER_EXPIRED
+
+        except SQLAlchemyError as e:
+            LOGGER.error("Database error encountered: {}".format(e))
+            return errors.DB_NOT_AVAILABLE
+        except:
+            LOGGER.error("Encountered unknown error: {}".format(traceback.format_exc()))
+            return errors.DB_NOT_AVAILABLE
+
+    def post(self):
```
avishkar58: This should have the `@admin_required` decorator :)

api/app/registration/api.py
```diff
@@ -0,0 +1,296 @@
...
+                return question, 201
+
+        except SQLAlchemyError as e:
+            LOGGER.error("Database error encountered: {}".format(e))
+            return errors.DB_NOT_AVAILABLE
+        except:
+            LOGGER.error("Encountered unknown error: {}".format(traceback.format_exc()))
+            return errors.DB_NOT_AVAILABLE
+
+    def post(self):
```
avishkar58: Also need `@admin_required` here

api/app/registration/api.py
```diff
@@ -0,0 +1,296 @@
...
+                return registration_section, 201
+
+        except SQLAlchemyError as e:
+            LOGGER.error("Database error encountered: {}".format(e))
+            return errors.DB_NOT_AVAILABLE
+        except:
+            LOGGER.error("Encountered unknown error: {}".format(traceback.format_exc()))
+            return errors.DB_NOT_AVAILABLE
+
+    def post(self):
```
avishkar58: Also need `@admin_required` here

api/app/registration/api.py
avishkar58: Please also add a check that the logged in user id is the same as the user_id on the offer - if not, return a 403 status

soginga:
@avishkar58 i've made the changes you've requested...

PR #292, opened 2019-06-09
Registration answers init

ilzeLourens:


ilzeLourens:
having trouble with offer table,
Says payment_required is undefined column

PR #297, opened 2019-06-11
Create offer frontend

LindanyRetroRabbit:


webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+              addedSucess: false,
+              conflict: true
+            });
+          }
+        })
+      }
+    
+    displayOfferContent = (e) => {
+      const {offerList,userProfile} = this.state;
+      console.log("<<< List  Function >>>>",offerList," userProfile ",userProfile)
```
jaderabbit: clean plz

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+     return(   
+      <div className="container"  align="center" >
+      <p className="h5 pt-5">We are pleased to offer you a place at the Deep Learning Indaba 2019. Please see the details of this offer below </p>
+
+      <form class="pt-5 ">
+        <p class="font-weight-bold">You have been accepted as a  {userProfile!=null ? userProfile.user_category: "<Category>"} </p>
```
jaderabbit: @avishkar58 Should we even be showing this here?

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+          <div class="col-8 font-weight-bold">
+            Payment Required:
+          </div>
+          <div class="col-2">
+          {offerList!=null
+          ? offerList.payment_required
+          : "Not available"}
+          </div>
+        </div>
+        <p class="font-weight-bold">Please accept or reject this offer by {offerList!=null ? offerList.expiry_date : "<Expiry Date>"} </p>
```
jaderabbit: If there is no expiry date being pulled from backend, front-end should error, not just display "<Expiry Date>"

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+        <div class="row">
+         <div class="col">
+          <button type="button" class="btn btn-danger" id="reject"onClick={
+            ()=>{
+              this.setState({
+                rejected:true
```
jaderabbit: @avishkar58 Assumedly we'd update this to a redirect once we have the registration url?

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+                    message: "Failed",
+                    error: (error.response && error.response.data)
+                    ? error.response.data.message
+                    : error.message
+                }
+            }
+        })
+}
+
+function updateOfferList(offer_id, event_id, accepted, rejected, rejected_reason){
```
jaderabbit: It's not an offer list? It's just a single offer?

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+            offer_id, event_id, accepted, rejected, rejected_reason),{ headers: authHeader() })
+        .then(function(response){
+            return{
+                message:"succeeded",
+                response:response
+            }
+        })
+        .catch(function(error){
+            if (error.response.status==="404") {
+                return {message:"404"}
```
jaderabbit: Please look at how other errors are handled on the system. We don't look at "404", we set an error message

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+            }else{
+                return{
+                message:"Failed",
+                    error:(error.response && error.response.data)
+                    ? error.response.data.message
+                    : error.message
```
jaderabbit: This is better

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+        offerServices.updateOfferList(offer_id, DEFAULT_EVENT_ID, accepted, rejected, rejected_reason)
+        .then(response=>{
+          if (response.msg === "succeeded") {
+            this.getofferList();
+            this.setState({
+              addedSucess: true,
+              conflict: false,
+              notFound: false
+            });
+          } else if (response.msg === "404") {
```
jaderabbit: You should check on the response.error

Please look at: ReviewForm.js as an example:

```
               reviewService
                    .submit(this.state.form.response.id, this.state.form.review_form.id, scores, shouldUpdate)
                    .then(response => {
                        if (response.error) {
                            this.setState({
                                error: response.error,
                                isSubmitting: false
                            });
                        }
                        else {
                            if (this.state.form.review_response) {
                                this.props.history.push(`/reviewHistory`)
                            }
                            else {
                                this.loadForm();
                            }
                        }
                    });
```

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+    return axios
+        .post(baseUrl + 'api/v1/offer',data,{headers:authHeader()})
+        .then(function (response){
+            return{
+                message:"succeeded",
+                response: response
+            }
+        })
+        .catch(function(error){
+            if (error.response.status===404){
```
jaderabbit: same here

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+    updateOfferList
+}
+
+function getOfferList(event_id){
+    return axios.get(baseUrl + "/api/v1/offer?event_id=" +event_id,{
+        headers: authHeader()
+      })
+      .then(function(response){
+        return{
+            offerList: response.data,
```
jaderabbit: is it a list of offers or a single offer?

webapp/src/pages/offer/components/OfferComponent.js
avishkar58: Yes

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+              Reject
+          </button>
+          </div>
+          <div class="col">
+          <button type="button" class="btn btn-success" id="accept" onClick={()=>{
+              this.setState({
+                accepted:true,
+                rejected:false,
+              });
+              this.buttonSubmit();
```
avishkar58: This can cause problems because setState is asynchronous - there is no guarantee that the state has updated by the time you read it in this.buttonSubmit - see [here](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) for more information. To get around this, you can put the this.buttonSubmit inside the setState callback like this: 
```
this.setState({
                accepted:true,
                rejected:false,
              }, ()=>this.buttonSubmit());
```

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+          <div class="form-group mr-5  ml-5 pt-5" >
+          <textarea class="form-control pr-5 pl-10" id="exampleFormControlTextarea3" onChange={()=>{this.handleChange(this.rejected_reason)}}  placeholder="Enter rejection message"></textarea>
+          <button type="button" class="btn" id="apply" onClick={
+            ()=>{
+              this.setState({
+                rejected:false
+              });
+              this.buttonSubmit();
+            }
+            }>apply</button>
```
avishkar58: You already have an onChange handler on the textarea - so no need for an apply button? And why is the apply button setting rejected to false??

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+        :""}
+        </div>
+      
+      </form>
+  </div>
+      )}
+
+    componentDidMount() {
+          this.getOfferList();
+          console.log("<<< List >>>>",this.state.offerList)
```
avishkar58: Clean up

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+          profileService.getProfilesList(DEFAULT_EVENT_ID)
```
avishkar58: This method returns a **list** of profiles - ie everyone who applied for the event! We only need one piece of information (the user_category) from the user whose offer is being viewed. Either use the UserProfile controller (`/api/v1/userprofile`) or better, update the Offer controller to return the user_category along with the offer.

webapp/src/services/offer/offer.service.js
```diff
@@ -0,0 +1,101 @@
...
+    let data = {
+        offer_id:offer_id, 
+        event_id:event_id,
+        accepted:accepted,
+        rejected:rejected,
+        rejected_reason:rejected_reason
+    };
+
+    return axios
+        .put(baseUrl + "/api/v1/offer?offer_id=&event_id=&accepted=&rejected=&rejected_reason=" % (
```
avishkar58: The data should be send with the request data, like how it's done in the method above, not in the query string.

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,210 @@
...
+class Offer extends Component {
+  
+  constructor(props) {
+        super(props);
+  
+        this.state = {
+          userProfile : [],
+          loading: true,
+          error: "",
+          offer_id: 1,
```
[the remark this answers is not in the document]
jaderabbit: oh lol, missed that

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,200 @@
...
+        :""}
+        </div>
+      
+      </form>
+  </div>
+      )}
+
+    componentDidMount() {
+          this.getOffer();
+          profileService.getUserProfile(DEFAULT_EVENT_ID)
```
avishkar58: getUserProfile takes a user_id, not event_id...

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: For hiding the rejected_reason textfeild after submitting rejection reason. I just needed to change the variable to showReasonBox variable instead of using reject variable itself.

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: Interesting, i've checked it

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: I wasn't sure about loggedin user_id, so I pulled it from App.js componentDidMount()

jaderabbit: @LindanyRetroRabbit  What do you mean here?

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: @LindanyRetroRabbit Even if you've tested it, that's only in your circumstances locally. Please follow avishkar's proposal here

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: @LindanyRetroRabbit I'm going to suggest returning user_category along with the offer

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: Yes, I mean't I have read the content on the... about setStaes callback and asynchronous. I also iimplemented on my code after going through the content

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: getUserProfile takes user_id parameter instead of event_id, so I should pass user_id as well.

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -2,10 +2,12 @@ import React, {Component} from "react";
 import {withRouter} from "react-router";
 import TextArea from "react";
 import {Chart} from "react-google-charts";
+import { userService } from "../../../services/user";
 import { offerServices  } from "../../../services/offer/offer.service";
 import { profileService} from "../../../services/profilelist/profilelist.service";
 
 const DEFAULT_EVENT_ID = process.env.REACT_APP_DEFAULT_EVENT_ID || 1;
+let DEFAULT_USER_ID = process.env.REACT_APP_DEFAULT_USER_ID || 1;
```
jaderabbit: There is no such thing as a default user ID? 
Also REACT_APP_DEFAULT_USER_ID.
Please remove this.
What you're saying, is that if a user isn't return, we randomly default to user 1? That makes no sense

The app has a default EVENT_ID because the "Deep Learning Indaba" and it has ID 1

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -143,14 +145,26 @@ class Offer extends Component {
...
       )}
 
+    componentWillMount(){
+      userService.get().then(result=> {
+        this.setState({
+            user:{
+              user_id:result.user_id
+            }
+        });
+        DEFAULT_USER_ID = result.user_id;
```
jaderabbit: please remove this too. This makes no sense?

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -143,14 +145,26 @@ class Offer extends Component {
...
+
     componentDidMount() {
           this.getOffer();
           this.setState({
             user: JSON.parse(localStorage.getItem("user")),
             loading: false,
             buttonLoading: false
           });
-          profileService.getUserProfile(user.user_id)
+          profileService.getUserProfile(DEFAULT_USER_ID)
```
jaderabbit: Do not use this detault user ID. It makes no sense.  
You need to user the one in local storage that you set above.
surely it's `state.user.id`?  
If it ends up being null then chances are you have a async issue.

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -7,15 +7,14 @@ import { offerServices  } from "../../../services/offer/offer.service";
...
 const DEFAULT_EVENT_ID = process.env.REACT_APP_DEFAULT_EVENT_ID || 1;
-let DEFAULT_USER_ID = process.env.REACT_APP_DEFAULT_USER_ID || 1;
 
 class Offer extends Component {
   
   constructor(props) {
         super(props);
   
         this.state = {
-          user: {},
```
jaderabbit: why is it 0?

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -129,51 +128,49 @@ class Offer extends Component {
...
+    </div>
       )}
 
-    componentWillMount(){
+    componentDidMount(){
       userService.get().then(result=> {
         this.setState({
             user:{
-              user_id:result.user_id
+              user_id:result.response.data.app_user_id
```
jaderabbit: where is this `response` variable?
Why are we setting a `user_id` object? 
Look at how this was done. Look at authenticate, look at the user service. 

I'm going to have to finish this since we're just not getting there

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -7,15 +7,14 @@ import { offerServices  } from "../../../services/offer/offer.service";
...
-let DEFAULT_USER_ID = process.env.REACT_APP_DEFAULT_USER_ID || 1;
 
 class Offer extends Component {
   
   constructor(props) {
         super(props);
   
         this.state = {
-          user: {},
+          user:{user_id:0},
```
jaderabbit: Why is it 0?

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,214 @@
...
+   
+        offerServices.updateOffer(offer_id, DEFAULT_EVENT_ID, accepted, rejected, rejected_reason)
+        .then(response=>{
+          if (response.msg === "succeeded") {
+            this.setState({
+              addedSucess: true,
+              notFound: false
+            });
+          } else if (response.error) {
+            this.setState({
```
jaderabbit: you need do display the error here. But get the functional part right. I'll fix this tomorrow morning

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: @LindanyRetroRabbit You still haven't adjusted this as requested

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,214 @@
...
+            user:{
+              user_id:result.response.data.app_user_id
+            }
+        });
+      });
+
+    }
+
+    componentWillMount() {
+      const {user_id} = this.state.user;
```
jaderabbit: Please check this works at all

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,214 @@
...
+          this.getOffer();
+          this.setState({
+            user: JSON.parse(localStorage.getItem("user")),
```
jaderabbit: here you load the user from storage. 
You should actually just use this. the user.id is here already?

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -143,26 +140,17 @@ class Offer extends Component {
...
           this.getOffer();
           this.setState({
-            user: JSON.parse(localStorage.getItem("user")),
+            
             loading: false,
             buttonLoading: false
           });
-          profileService.getUserProfile(user_id)
+
+          let currentUser = JSON.parse(localStorage.getItem("user"));
```
jaderabbit: Do you not want to store the user on the state?

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: @LindanyRetroRabbit Reminder to remove this

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: reminder

webapp/.env.development
```diff
@@ -1,2 +1,3 @@
 REACT_APP_API_URL=http://localhost:5000
-REACT_APP_DEFAULT_EVENT_ID=1
\ No newline at end of file
+REACT_APP_DEFAULT_EVENT_ID=1
+REACT_APP_DEFAULT_USER_ID=1
```
jaderabbit: reminder to remove

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,202 @@
...
+          rejected:false,
+          showReasonBox:false,
+          accepted: false,
+          offerList: [],
+          category:""
+        };
+      }
+    
+      
+      handleChange = field => {
```
jaderabbit: is this even used? I don't think so

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -0,0 +1,202 @@
...
+            buttonLoading: false
+          });
+
+          let currentUser = JSON.parse(localStorage.getItem("user"));
+          profileService.getUserProfile(currentUser.id)
+          .then(results => {
+              this.setState(
+                  {
+                    userProfile: results,
+                      loading: false,
```
jaderabbit: fix your indentation here

webapp/src/pages/offer/components/OfferComponent.js
LindanyRetroRabbit: Yes, during takes takes the value passed when entering rejection reason. event handler.

webapp/src/pages/offer/components/OfferComponent.js
jaderabbit: SInce you seem to have fixed the other things, can we sort this handling of the error properly now (as you've done with the others :)

PR #298, opened 2019-06-11
created a ViewProfile, with readOnly input types.

mthembbu:
Additional info about the application status has been catered for as well
Bootstrap 4.3.1 has been included.

### DIFF
--- a/webapp/src/pages/profileList/components/ProfileListComponent.js
+++ b/webapp/src/pages/profileList/components/ProfileListComponent.js
@@ -33,7 +33,7 @@ class ProfileListComponent extends Component{
                             });
                     });                
  }
-onsub=(user_id)=>{
+onSubmit=(user_id)=>{
     console.log(user_id)
     window.location='/viewprofile/:'+user_id;
 }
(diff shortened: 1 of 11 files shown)


webapp/src/pages/viewprofile/components/ViewProfileComponent.js
```diff
@@ -0,0 +1,295 @@
...
+        super(props);
+        this.state = {
+            user: {},
+            loading: true,
+            error:"",
+            isNull:true
+        }
+    }
+
+isUserNull(user){
```
avishkar58: This is nit, but is this method necessary? I think it's clearer to just "inline" this and have have `result.data === null`  in your setState function directly.

webapp/src/pages/profileList/components/ProfileListComponent.js
```diff
@@ -33,15 +33,20 @@ class ProfileListComponent extends Component{
                             });
                     });                
  }
-
+onsub=(user_id)=>{
```
avishkar58: Can this method have a more descriptive name please?

webapp/src/pages/viewprofile/components/ViewProfileComponent.js
mthembbu: dano....

webapp/src/pages/profileList/components/ProfileListComponent.js
mthembbu: sorted

webapp/src/pages/profileList/components/ProfileListComponent.js
```diff
@@ -33,15 +33,20 @@ class ProfileListComponent extends Component{
                             });
                     });                
  }
-
+onSubmit=(user_id)=>{
+    console.log(user_id)
```
jaderabbit: clean plz

webapp/src/services/profilelist/profilelist.service.js
```diff
@@ -20,4 +21,18 @@ function getProfilesList(event_id){
                       ? error.response.data.message :
                          error.message }
         });
+}
+
+function getUserProfile(user_id){
+    return axios
+            .get(baseUrl+"/api/v1/userprofile?user_id="+user_id.toString().split(":")[1],{headers: authHeader()})
```
jaderabbit: @mthembbu Can you explain your reasoning here? Why do we need to make it a string and split it?

mthembbu: my mistake was not creating a new variable as an int, then parse that int instead.

Issue #302, opened 2019-06-12
Offer Bug: Error when submitting offer & interface does nothing


jaderabbit:
When I view my offer and try to respond with accept or decline, nothing happens. 

In my console, I have the above error:
![image](https://user-images.githubusercontent.com/5547095/59383325-f696d380-8d5f-11e9-8832-1c9940015b42.png)

jaderabbit:
@LindanyRetroRabbit please add extra details about the bug!

@soginga Lindani seems to be fighting his machine. Would you assist with fixing this bug? He says a null "offer_id" is passed to the backend

PR #306, opened 2019-06-12
Think this fixes the offer submission bug

jaderabbit:
hopefully addresses #302


soginga:
prettier extension is responsible for those many file changes... disabling it.

jaderabbit:
> prettier extension is responsible for those many file changes... disabling it.

@soginga Thanks

PR #309, opened 2019-06-13
offer response message to display after accepting or rejecting

LindanyRetroRabbit:
In progress


webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -31,30 +30,43 @@ class Offer extends Component {
...
       });
   }
-
+  
+  displayOfferResponse = e =>{
+       const { offer } = this.state;
+       return (<div className="container">
+           <div className="white-background card form">
+              {offer.candidate_response?
+              "Accepted an offer"
```
jaderabbit: Can we update this to be nice?

"You have accepted the following offer:
insert details here

"You have rejected the following offer:
insert offer details here
reason

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -41,9 +41,11 @@ class Offer extends Component {
         candidate_response? "" : rejected_reason
       )
       .then(response => {
-        if (response.msg === "succeeded") {
+      console.log(response)
+        if (response.status === 201) {
           this.setState({
-            offer: response.data
+            offer: response.data,
```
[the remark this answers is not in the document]
avishkar58: Are you sure it's being called? Try set a breakpoint in the Chrome debugger and see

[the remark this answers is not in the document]
soginga: Great point, thanks! ... turns out that response contained another response inside!  had use 

> response.response.data

webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -31,33 +36,71 @@ class Offer extends Component {
...
 
     offerServices
       .updateOffer(
         offer.id,
         DEFAULT_EVENT_ID,
-        accepted,
-        rejected,
-        rejected_reason
+        candidate_response,
+        candidate_response? "" : rejected_reason
```
KaleabTessera: Consider adding a check for candidate_response being null, I know you call buttonSubmit and haven't passed a null but others might.

soginga: candidate_response having a null isn't a problem is this case,I think... cause if its a null it just means the user hasn't responded?

KaleabTessera: I get you, I am referring to the ```candidate_response? "" : rejected_reason``` line. If candidate_response is null, then rejected_reason will be passed, which would most prob be a null. 

Also, having a check in this function to see if candidate response is null (if it is then we just break out), will prevent sending null responses to updateOffer. It is more of a precautionary thing.

avishkar58:
Merging to not block testing

PR #311, opened 2019-06-14
list-profile-fullname,table cell updated with a cursor

mthembbu:


jaderabbit:
@mthembbu can we not also style it as a hyperlink? :P

Issue #305, opened 2019-06-12
Update Baobab front-page to take user to offer, if one exists


jaderabbit:
Right now, even though I've got an offer waiting for me, the front-page still says this: 

![image](https://user-images.githubusercontent.com/5547095/59384860-783c3080-8d63-11e9-97f3-b6c5b7e6862b.png)

Perhaps it should say, if there exists an offer for the user:

*There is an offer waiting for you, click here to view it*

And the default message should be updated to:

*You are currently on the waiting list for the Deep Learning Indaba 2019. Please await further communication*

@avishkar58  Do we eventually send people rejection emails after a certain date? Maybe we should mention this date somewhere?

avishkar58:
@jaderabbit Yeah I think we should send people rejection emails, especially those who are not based in Nairobi (who we might call on at very short notice to fill last minute places) - will report back with some more details.

PR #318, opened 2019-06-18
check if the offer has been issued on the home page and display appro…

soginga:
…priate message if it doesn't or url to it if does #305

### DIFF
--- a/webapp/src/pages/home/Home.js
+++ b/webapp/src/pages/home/Home.js
@@ -3,6 +3,7 @@ import logo from '../../images/indaba-logo-dark.png';
 import './Home.css';
 import { getEvents } from "../../services/events";
 import { applicationFormService } from "../../services/applicationForm";
+import { offerServices } from "../../services/offer/offer.service";
 import { NavLink } from "react-router-dom";
 
 const DEFAULT_EVENT_ID = process.env.REACT_APP_DEFAULT_EVENT_ID || 1;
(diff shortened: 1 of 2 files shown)


webapp/src/pages/home/Home.js
```diff
@@ -116,26 +124,31 @@ class Home extends Component {
...
                 </h4>
-                
-                {this.state.applicationStatus == "Submitted" && <p>
-                  Thank you! Your application has been received and is under review. You can expect to hear our decision mid-June.
-                </p>}
+
+                {this.state.applicationStatus == "Submitted" &&
+                <div>
+                  <p>
+                    Thank you! Your application has been received and is under review. You can expect to hear our decision mid-June.
```
avishkar58: Lets remove this line - so we just have information about the offer. Alternatively you could add a table with the event, application date and offer status

webapp/src/pages/home/Home.js
```diff
@@ -116,26 +124,31 @@ class Home extends Component {
...
+                  </p>
+                  {this.state.offer !== null ? <p>There is an offer waiting for you, <NavLink to="/offer">click here</NavLink> to view it.</p>
+                  : <p>You are currently on the waiting list for the Deep Learning Indaba 2019. Please await further communication.</p>}
+                </div>}
                 {this.state.applicationStatus == "Withdrawn" && <p>
                   Your application has been withdrawn - you will not be considered for a place at the Indaba.
                 </p>}
                 {this.state.applicationStatus == "NOT Submitted" && <p>
                   You have NOT submitted your application! <NavLink to="/applicationForm">Complete and submit</NavLink> it before 26 May to be considered for a place at the Indaba.
```
avishkar58: Please remove the links to applicationForm as these are no longer relevant :)

soginga: what should be the remaining text there

> You have NOT submitted your application!
?

avishkar58:
Yeah that one can stay, thanks! :)

PR #324, opened 2019-06-19
Accepted Accommodation & Travel been updated for both Front-end and b…

mthembbu:
…ack-end.


webapp/src/pages/offer/components/OfferComponent.js
```diff
@@ -147,6 +164,28 @@ class Offer extends Component {
                   : "Not available"}
               </div>
             </div>
+
+            <div class="row pb-5 ">
+              <div class="col-6 font-weight-bold pr-4"  align="right">
+                <div class="form-check accommodation-container">
```
jaderabbit: This should only show if they've been awarded accomodation.
Is that the case? 

Same with travel_award

api/migrations/versions/539d0567cd6c_.py
```diff
@@ -1,5 +1,6 @@
 """empty message
 
+
```
jaderabbit: Blank migrations?

api/migrations/versions/472bf293e0a1_.py
```diff
@@ -0,0 +1,22 @@
...
+# revision identifiers, used by Alembic.
+revision = '472bf293e0a1'
+down_revision = ('79c61673a487', '7e8bffa88454')
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    pass
```
jaderabbit: Also this one?

mthembbu: Resolving. Thanks @jaderabbit

jaderabbit: DO NOT RESOLVE NOW! IT'S ALREADY DEPLOYED! Eeek

Issue #316, opened 2019-06-16
User profile page should show responses and reviews


avishkar58:
Update the user profile (details) page to display all the answers in the applicant's response. Also display the reviews that reviewed their application along with their final verdicts. This involves updating the backend (api/app/users/api.py) and front-end (webapp/src/pages/viewprofile/components/ViewProfileComponent.js)

LindanyRetroRabbit:
Quick question @avishkar58 , just for clarity

Should we display all user comments, and
List all the reviewers names alongside with their comments and final score (verdicts) of that specific user's application?

In the front-end must we create the new section under the EventAdmin >> ApplicantProfile tab?

Review_response table the bridging table between the App_user(user_id) and the Review_score(value) table

jaderabbit:
@avishkar58 

To confirm:
On a user's profile, we should display the following table:

| Reviewer Name | Reviewer Comments |  Reviewer Final Verdict |
| ----------------------|----------------| -------------|
| Jade Abbott | Excellent submission | 4
| Harry Potter | Magical! | 5
| Voldemort | None of this | 1

avishkar58:
Yes exactly :)

avishkar58:
Ideally converting the final verdict value into its corresponding label (eg
No, Maybe or Yes) which can be obtained by joining to ReviewQuestion and
looking at the options field.

jaderabbit:
@avishkar58 Oh wait, So exactly like we have under review history?

jaderabbit:
@avishkar58 And would it make sense so then have an api called `user-reviews` which would return such a structure?

avishkar58:
If by "exactly like under review history" you mean how the final verdict is
displayed - then yes :)
Sure, a user-reviews api makes sense I think

PR #325, opened 2019-06-20
User profile page should show comments and reviews #316

LindanyRetroRabbit:
Though I am not sure about the relationship between Reviewer, ReviewResponse and the UserComment tbl


api/app/users/api.py
```diff
@@ -545,3 +554,31 @@ def get(self):
...
+
+        current_user = user_repository.get_by_id(g.current_user['id'])
+        if not current_user.is_event_admin(args['event_id']):
+            return FORBIDDEN
+
+        comments = db.session.query(UserComment).filter(
+            UserComment.event_id == args['event_id'],
+            UserComment.user_id == args['user_id']).all()
+        
+        review_by_user_firstname = user_repository.get_by_id(args['user_id']).firstname
```
jaderabbit: So, who is this getting? The current users's firstname? 
Should we not be getting the first name of the reviewers who made the comments?

api/app/users/api.py
```diff
@@ -545,3 +554,31 @@ def get(self):
...
+        
+        reviewer_id = db.session.query(ReviewResponse).filter(
+            ReviewResponse.reviewer_user_id == args['user_id'])
```
jaderabbit: I'm not sure what this code is doing?

LindanyRetroRabbit: image

LindanyRetroRabbit: Creates and  returns an object with 4 attributes.

jaderabbit: But we should not be  filtering by the reviewer_user_id = args['user_id']. 

A specific user creates a specific Response object (application response). (so you'll need to filter on user_id for Response)
ReviewResponse's are assigned to specific Response objects
Those are the reviews you need to pull
Reviewers are assigned to specific ReviewResponse's - reviewer_user_id - get the names of *those* users

And the call to get reviews for a specific user, should perhaps be in the reviews api.py, not under user?

api/app/users/api.py
```diff
@@ -543,5 +548,55 @@ def get(self):
...
+        req_parser.add_argument('user_id', type=int, required=True)
+        args = req_parser.parse_args()
+        user_id = args['user_id']
+        event_id =args['event_id']
+        
+        current_user = user_repository.get_by_id(g.current_user['id'])
+        if not current_user.is_event_admin(args['event_id']):
+            return FORBIDDEN
+
+        comments = db.session.query(UserComment).filter(UserComment.event_id == args['event_id'], UserComment.user_id == args['user_id']).all()
```
jaderabbit: Not every review has a comment, so you can't use it to get the reviewers.
You need to use reviews to get reviewers.

I've advised on this before. Please look at how I explained the logic to work before.

LindanyRetroRabbit:
#316

Issue #329, opened 2019-06-21
Invitation Letter Controller


jaderabbit:
Implement the Invitation Letter Controller as per the [Registration specification](https://docs.google.com/document/d/1dBTfmsBvWtCEw36mKYIgRsvMxE0BhMT-mU9iHVcWTrs/edit#). The invitation letter controller needs to save the invitation letter request, . 

Create `api/app/invitationletter/api.py` controller:

POST method: 
```
{
   registration_id: 1, 
   event_id: 1,
   work_address: "Somewhere over the rainbow",
   addressed_to: "Sir",
   residential address: "Way up high"
   passport_name: "Jane Doe",
   passport_no: "23456565",
   passport_issued_by: "Neverland"
   to_date: datetime,
   from_date: datetime  
}
```

Controller functionality: 
1. Persist the InvitationLetterRequest
2. Go fetch the related *offer*. 

- If the offer has travel & accomodation -> then get the InvitationTemplate, for the correct event, that has `send_for_both_travel_accom` set to true
- Else if the offer has travel - then get the InvitationTemplate, for the correct event, that has `send_for_travel_award_only`
- Else if the offer has accomodation - then get the InvitationTemplate, for the correct event, that has `send_for_accomodation_award_only`

2. Given the correct InvitationTemplate model, please get the related template, which will be a word document. 

3. Fetch the relevant user, as we will need the following details from there:  
- country_of_residence
- nationality
- date_of_birth

4. Fetch the relevant registration and registration_answers, extract from there whether or not the person will be presenting a poster.

5. Call the `generate` method of the Invitation letter generator specified in issue #328 

6. Update the `invitation_letter_sent_at` field if sent correctly

7. Return the invitation letter request ID if correctly sent, else send the appropriate error

jaderabbit:
@avishkar58 @KaleabTessera Another sense check here

jaderabbit:
@soginga If you don't want to be blocked by #328 's ticket, I'd suggest creating a module that has the same method signature as what @mthembbu needs to do.

PR #341, opened 2019-06-24
Invitation letter controller

soginga:
#329


api/app/invitationletter/api.py
```diff
@@ -0,0 +1,115 @@
...
+        global invitation_template
+
+        if offer.accommodation_award and offer.travel_award:
+            invitation_template = db.session.query(InvitationTemplate).filter(
+                InvitationTemplate.event_id == offer.event_id).filter(
+                InvitationTemplate.send_for_both_travel_accommodation).first()
+        elif offer.travel_award:
+            invitation_template = db.session.query(InvitationTemplate).filter(
+                InvitationTemplate.event_id == offer.event_id).filter(
+                InvitationTemplate.send_for_both_travel_accommodation).first()
```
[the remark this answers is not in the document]
soginga: thanks! overlooked those.

api/app/invitationletter/tests.py
```diff
@@ -0,0 +1,108 @@
...
+        data = json.loads(response.data)
+        header = {'Authorization': data['token']}
+        return header
+
+    def test_create_create_invitation_letter(self):
+        self.seed_static_data()
+        response = self.app.post(
+                '/api/v1/invitation-letter', data=INVITATION_LETTER, headers=self.headers)
+        data = json.loads(response.data)
+        LOGGER.debug(
```
jaderabbit: can you assert the fields are saved correctly?

soginga: That's not possible given i'm supposed to return only invitation_letter_request_id from the post request #329 or should i modified what get's returned from the api to InvitationLetterRequest fields?

jaderabbit: @soginga You should go retrieve it from the database and check it's updated. Its a test. This is how one tests :P

api/app/invitationletter/api.py
```diff
@@ -0,0 +1,115 @@
...
+            from_date=from_date,
+        )
+
+        offer = db.session.query(Offer).filter(
+            Offer.user_id == user_id).first()
+
+        if not offer:
+            return errors.OFFER_NOT_FOUND
+
+        global invitation_template
```
avishkar58: Why is this global?

api/app/invitationletter/api.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if offer.accommodation_award and offer.travel_award:
```
avishkar58: Please also check accepted_accommodation_award and accepted_travel_award, so the condition should be something like: 
```if offer.accommodation_award and offer.accepted_accommodation_award and offer.travel_award and offer.accepted_travel_award:```
Same applies for the below two

File: api/app/invitationletter/api.py [hunk 5]
```diff
@@ -0,0 +1,115 @@
...
+                InvitationTemplate.send_for_both_travel_accommodation).first()
+        elif offer.travel_award:
+            invitation_template = db.session.query(InvitationTemplate).filter(
+                InvitationTemplate.event_id == offer.event_id).filter(
+                InvitationTemplate.send_for_travel_award_only).first()
+        elif offer.accommodation_award:
+            invitation_template = db.session.query(InvitationTemplate).filter(
+                InvitationTemplate.event_id == offer.event_id).filter(
+                InvitationTemplate.send_for_accommodation_award_only).first()
+
```
avishkar58: What about the base case for people who have no travel or accommodation award?

api/app/invitationletter/api.py
jaderabbit: OMG YES WHY YOU DO DIS? :scream:

jaderabbit:
@soginga We've updated the issue. 

We ask during registration if they're bringing a poster. 
The template will have %BRING_POSTER%

- *If*, in their registration, they said they're bringing a poster, then we should populate %BRING POSTER% with the following (temporary, to be updated) text:
```
They will be presenting an academic poster of their research at the event, entitled: <Poster Title>
```
These details you can get from registration 

- *Else*, populate %BRING_POSTER% with an empty string

api/app/invitationletter/api.py
soginga: mm🤔 yea it's kinda dumb I'm not sure... I was getting the warning that `"Local variable 'invitation_template' might be referenced before the assignment"` looked it up, they suggested that on stackoverflow

avishkar58:
Better than global would be to use `invitation_template = None`. Better
still would be to think about what the warning is telling you - you might
be using invitation_template without having assigned it a value - ask
yourself whether this is a valid state for your program. In this case, it
isn't, because when it comes time to use invitation_template to generate
the letter, you expect it to have been populated with one of the templates.
So the best solution would be to remove that line entirely and instead
handle the ELSE case to set invitation_template for people who have no
awards :)

api/app/invitationletter/api.py
soginga: I was getting the `Local variable 'invitation_template' might be referenced before assignment`

api/app/invitationletter/api.py
```diff
@@ -57,21 +57,27 @@ def post(self):
...
                 InvitationTemplate.send_for_travel_award_only).first()
-        elif offer.accommodation_award:
+
+        elif offer.accommodation_award and offer.accepted_accommodation_award:
             invitation_template = db.session.query(InvitationTemplate).filter(
                 InvitationTemplate.event_id == offer.event_id).filter(
                 InvitationTemplate.send_for_accommodation_award_only).first()
 
+        elif (not offer.accommodation_award) and (not offer.travel_award):
+            return errors.NO_ACCOMMODATION_TRAVEL_AWARD
```
jaderabbit: It's not an error case. They also get a invitation letter even if they don't get an award. There is a default invitation letter

api/app/utils/errors.py
```diff
@@ -65,3 +65,5 @@
     {'message': 'Offer has not been accepted'}, 409)
 APPLICATIONS_CLOSED = (
     {'message': 'Applications are now closed'}, 403)
+NO_ACCOMMODATION_TRAVEL_AWARD = (
+    {'message': 'No template found because your offer has neither accommodation nor travel award allocation'}, 404)
```
jaderabbit: There is a template if you have neither.
You still get an invitation letter even if they're not sponsoring travel or accomodation :P

api/app/invitationletter/api.py
soginga: so what's the condition to retrieve that template? 

> `....filter( not InvitationTemplate.send_for_both_travel_accommodation).filter( not InvitationTemplate.send_for_travel_award_only).filter( not InvitationTemplate.send_for_accommodation_award_only)` ?

jaderabbit: All things false :)

api/app/invitationletter/api.py
```diff
@@ -0,0 +1,124 @@
...
+                               passport_issued_by=passport_issued_by,
+                               invitation_letter_sent_at=invitation_letter_request.invitation_letter_sent_at,
+                               to_date=to_date,
+                               from_date=from_date,
+                               country_of_residence=country_of_residence,
+                               nationality=nationality,
+                               date_of_birth=date_of_birth,
+                               email=user.email)
+
+            if is_sent:
```
jaderabbit: if it's not sent, then we also need a failcase. 
"Failed to send Invitation Email"

PR #342, opened 2019-06-24
Offer & Registration Bug fixes

avishkar58:


api/app/registration/api.py
```diff
@@ -143,30 +143,29 @@ def post(self):
...
+
         offer_entity = Offer(
             user_id=user_id,
             event_id=event_id,
             offer_date=offer_date,
             expiry_date=expiry_date,
             payment_required=payment_required,
             travel_award=travel_award,
-            accommodation_award=accommodation_award,
-            accepted_accommodation_award=accepted_accommodation_award,
```
jaderabbit: Is this not working properly?

avishkar58: It works - but we shouldn't be setting it when creating an offer...

PR #354, opened 2019-06-27
Guest registration

ilzeLourens:


webapp/src/pages/registration/Registration.js
```diff
@@ -1,11 +1,39 @@
...
+        this.setState({
+          GuestRegistration: false
+        });
+      }
+    })
+  }
+  render() {
+    return (
+      <div>
+        {this.state.GuestRegistration == true ? <GuestRegistrationComponent></GuestRegistrationComponent> : this.state.GuestRegistration == true ? <RegistrationComponent></RegistrationComponent> : ""}
```
jaderabbit: these are both checking the same value to be true

webapp/src/pages/registration/components/GuestRegistrationComponent.js
```diff
@@ -0,0 +1,485 @@
...
+        });
+    }
+
+    handleChangeDropdown = (id, dropdown) => {
+        let value = dropdown.value.toString();
+        this.onChange(id, value);
+    };
+
+    componentDidMount() {
+                    registrationService.getGuestRegistration(DEFAULT_EVENT_ID).then(result => {
```
jaderabbit: fix indentation here

ilzeLourens:
> Only a few things from me.

@jaderabbit It's determined by if the user is a guest or normal user and shows the appropriate component eg. if user is a guest it shows  <GuestRegistrationComponent></GuestRegistrationComponent> which is the guest registration page

jaderabbit:
> > Only a few things from me.
> 
> @jaderabbit It's determined by if the user is a guest or normal user and shows the appropriate component eg. if user is a guest it shows which is the guest registration page

I just couldn't find the code, but also I'm a terrible js + html dev :P I'll trust you if you say it is there

webapp/src/pages/registration/Registration.js
```diff
@@ -1,11 +1,39 @@
...
+    this.state = {
+      GuestRegistration: null
     }
-  }
\ No newline at end of file
+  }
+
+  componentDidMount() {
+    registrationService.determineIfGuest(DEFAULT_EVENT_ID).then(response => {
+      if (response !== "404") {
```
KaleabTessera: Rather specifically check for ```response === 200```, in this case a response of 400 set would ```GuestRegistration: true```.

api/app/guestRegistrations/api.py
```diff
@@ -0,0 +1,307 @@
...
+                                   question.description + "\nAnswer :" + _get_answer_value(
+                            answer, question) + "\n"
+
+            subject = event_name + ' Registration'
+            greeting = strings.build_response_email_greeting(user.user_title, user.firstname, user.lastname)
+            if len(summary) <= 0:
+                summary = '\nNo valid questions were answered'
+            body_text = greeting + self.get_confirmed_message(confirmed) + '\n\n' + summary
+
+            emailer.send_mail(user.email, subject, body_text=body_text)
```
KaleabTessera: Relating to the comment above in the `api/app/guestRegistrations/api.py` file. We need this function to return an indication of successful mail. If this doesn't throw an exception, does it mean that the mail was send? In that case we can just check if this throws an exception and if it doesn't, we assume it was successful.

jaderabbit: Sorry, this was my fault. I gave them a method signature and my brain was in golang mode. 
golang has a bizarre error system

api/app/guestRegistrations/api.py
```diff
@@ -0,0 +1,307 @@
...
+        # Update an existing response for the logged-in user.
+        req_parser = reqparse.RequestParser()
+        args = self.req_parser.parse_args()
+        try:
+            user_id = verify_token(request.headers.get('Authorization'))['id']
+            registration = db.session.query(GuestRegistration).filter(
+                GuestRegistration.id == args['guest_registration_id']).first()
+            if registration is None:
+                return 'Registration not found', 404
+            registration.registration_form_id = args['registration_form_id']
```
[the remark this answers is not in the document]
ilzeLourens: This is the put, so updating the existing registration questions,
If you somehow manager to execute put on front end instead of the post before a registration exists i presume this shouldn't be done, however i doubt this will be the case, according to how the front end is calling the api, n

api/app/guestRegistrations/api.py
```diff
@@ -0,0 +1,307 @@
...
+
+            if not registration_form:
+                return errors.REGISTRATION_FORM_NOT_FOUND
+
+            registration = GuestRegistration(
+                registration_form_id=args['registration_form_id'],
+                user_id=user_id,
+                confirmed=True,
+                created_at=date.today(),
+                confirmation_email_sent_at=date.today()
```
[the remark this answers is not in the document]
ilzeLourens: image
image

The above is my proposed solution
however even though nullable=False in the models,
i get the below error for violating null constraint
image

api/app/guestRegistrations/api.py
ilzeLourens: Returned as required

webapp/src/pages/registration/Registration.js
ilzeLourens: Changed

api/app/guestRegistrations/api.py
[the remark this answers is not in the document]
ilzeLourens: I have committed the proposed solution, 
But commented out,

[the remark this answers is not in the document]
KaleabTessera: Try nullable=True and rerun the migration. If that doesn't work then I can help you dig deeper into the issue.

api/app/guestRegistrations/api.py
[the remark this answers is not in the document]
KaleabTessera: Okay cool.

Issue #328, opened 2019-06-21
Invitation Letter Generator


jaderabbit:
This module needs to generate Invitation Letters for a particular user, as per the [Registration spec](https://docs.google.com/document/d/1dBTfmsBvWtCEw36mKYIgRsvMxE0BhMT-mU9iHVcWTrs/edit#)

Create a `api/app/invitationletter/generator.py` module with a `generate` method that takes the following inputs:

- template_path
- event_id 
- work_address
- addressed_to
- residential address
- passport_name
- passport_no
- passport_issued_by
- invitation_letter_sent_at
- to_date
- from_date  
- country_of_residence 
- nationality
- date_of_birth 
- email

It must do the following: 
1. Fetch the word template from google cloud storage. [Here](https://docs.google.com/document/d/1OA-ER_Oscb_P3YETj_rfpql6Txiqg1cYCa0QbB01ZX0/edit?usp=sharing) is an example of a template

2. Replace the tags in the invitation template, with the correct fields . You can use the following package: https://pypi.org/project/docx-mailmerge/ 

3. Convert the word doc to a PDF

4. Email the the user the newly generated invitation letter

5. Return true if email was successful

**Definition of Done**
This task is done when:

- A new module exists in api/app/invitationletter/generator.py implementing method described above 
- Unit tests exist for the module

jaderabbit:
@avishkar58 @KaleabTessera Can you do a sense check here plz

avishkar58:
Looks good to me! Only thing I'd add is that the letter should be converted/saved as PDF before emailing it.

avishkar58:
Note: Please also add whether or not the person is presenting a poster to the fields.

jaderabbit:
> Note: Please also add whether or not the person is presenting a poster to the fields.

Does this need to go on the invitation letter? :P

avishkar58:
Yes, apparently this is useful for people using the letter to seek funding
from their university or company etc :)

jaderabbit:
I figure, we could just have %BRINGING_POSTER% on the templates, and then mailmerge that with text if they're bringing one. I'm trying to avoid another permutation of invitation templates :P

avishkar58:
Yes definitely, much better than doubling the templates :)

PR #357, opened 2019-07-02
Invitation letter generator

mthembbu:
#328


api/app/invitationletter/generator.py
```diff
@@ -1,9 +1,105 @@
...
+
+os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = json.dumps(GCP_CREDENTIALS_DICT)
+
+
+OFFER_EMAIL_BODY = """
+Dear {user_title} {first_name} {last_name},
+
+Congratulations! You've been selected to attend the {event_name}!
+
+Please see the attached document below for your acceptance of offer: {host}/offer
```
jaderabbit: I'm not sure if this is github being weird, but I gave specific copy for this
It's not an offer email. it's an invitation letter

api/app/invitationletter/generator.py
```diff
@@ -1,9 +1,105 @@
...
+from app.utils import emailer
+from app.utils import pdfconvertor
+from app.events.models import Event
+from app import db
+from app.utils import errors
+
+os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = json.dumps(GCP_CREDENTIALS_DICT)
+
+
+OFFER_EMAIL_BODY = """
```
jaderabbit: it's not an offer email, it's an invitation email

api/app/invitationletter/generator.py
```diff
@@ -1,9 +1,107 @@
...
+        LOGGER.debug("merge-fields.... {} .".format(document.get_merge_fields()))
+
+        document.merge(
+            TITLE=user_title,
+            FIRSTNAME=firstname,
+            SURNAME=lastname,
+            WORK_ADDRESS=work_address,
+            ADDRESSED_TO=addressed_to,
+            RESIDENTIAL_ADDRESS=residential_address,
+            PASSPORT_NAME=passport_name,
```
KaleabTessera: I think having ```Firstname```, ```Surname``` and ```Passport_Name``` is redundant.  I think we should just use ```Passport_Name```.

jaderabbit: The templates are currently using firstname, lastname. We don't even use passport name, but we should be

KaleabTessera: Can't we just replace firstname and lastname, with Passport_name? We just need their full name as  shown on their passport. 

Even on the templates we can just change it to use to Passport_name. I can't think of a use case for having the names split and having seperate fields for firstname,lastname and passport name.

jaderabbit: Agreed. Can we merge for now though? Testing the download from gcp is proving to be nightmare, so would like to test on staging if we can.

KaleabTessera: Cool.

PR #360, opened 2019-07-02
Registration admin front end

### LINKED ISSUE #333: Registration Administration Front End
A registration admin is in charge of tracking un-confirmed registrations and handling any processes that occur outside of Baobab to confirm these registrations (for example, liasing with the events organiser about registration fee payments for the Indaba 2019). They will perform their duties using a page in Baobab that lists the candidates who have registered for an event and have the confirmed flag set to FALSE (fields to display: user title, firstname, lastname, email, user_category, registration.created_at) There should be a button next to unconfirmed candidates that the registration admin can use to mark the registration as confirmed.

A user who is a registration admin should have access to a "Registration Administration" page. 

*On this page*:
1. A table of unconfirmed attendees for the event should be displayed. The following details for each attendee should be displayed: 
   - user title,
   - firstname, 
   - lastname, 
   - email,
   - user_category, 
   - registration.created_at

2. Additionally, a "Confirm" button must exist for each attendee in the table. The confirm button must hit the API described in #332 to confirm the attendee's registration.

3. The table must be searchable by attendee first name, last name and email

Dependant on #332

avishkar58:
Resolves #333

### DIFF
--- a/api/app/registrationResponse/api.py
+++ b/api/app/registrationResponse/api.py
@@ -1,24 +1,36 @@
-from datetime import date
+import abc
+from abc import ABCMeta
+from datetime import date, datetime
 import traceback
 from flask_restful import reqparse, fields, marshal_with, marshal
 import flask_restful as restful
 from flask import g, request
 from sqlalchemy.exc import SQLAlchemyError
 from app.utils.auth import verify_token
-from app.registrationResponse.mixins import RegistrationResponseMixin
+from app.registrationResponse.mixins import RegistrationResponseMixin, RegistrationAdminMixin, RegistrationConfirmMixin
 from app.registration.models import Offer, Registration, RegistrationAnswer, RegistrationForm, RegistrationQuestion
 from app.users.models import AppUser
 from app.events.models import Event
-from app.utils.auth import auth_required
+from app.utils.auth import auth_required, admin_required
+from app.users.repository import UserRepository
+from app.events.repository import EventRepository
 from app.utils import errors, emailer, strings
 from app import LOGGER
 from app.users.repository import UserRepository as user_repository
+from app.registrationResponse.repository import RegistrationRepository
 
 
 from app import db
 
 
 REGISTRATION_MESSAGE = 'Thank you for completing our registration form.'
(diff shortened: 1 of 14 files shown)


api/app/registrationResponse/api.py
```diff
@@ -1,24 +1,36 @@
-from datetime import date
+import abc
```
jaderabbit: just this :P

webapp/src/pages/registrationAdmin/components/RegistrationAdminComponent.js
```diff
@@ -0,0 +1,150 @@
...
+      let value = field.target.value.toLowerCase();
+      this.setState({
+        searchTerm: value
+      }, ()=>this.filterList());      
+    }
+
+    filterList = () => {
+      let value = this.state.searchTerm;
+      let filteredList = this.state.unconfirmedList.filter(
+        u => u.firstname.toLowerCase().indexOf(value) > -1 || u.lastname.toLowerCase().indexOf(value) > -1 || u.email.toLowerCase().indexOf(value) > -1);
```
KaleabTessera: Might need to rework this a little. If for example, your are searching for someone with the surname `John`, and you have people with the name `John`, then the filter will not include people with the surname `John`. You might need to filter on each field independently.

avishkar58: @KaleabTessera not sure I follow? There is a single search term that is applied as an OR to all 3 fields of each record - so if you search for John it will include people with the name OR surname "John"  ?

KaleabTessera: @avishkar58 So if there is match for people with the name John, then it won't filter people with the surname John, because the `or` condition will have already been met. This means the filtered list will only filter out people with the name John and won't even look at the surname.

Does that make sense?

avishkar58:
Yeah, but I don't see why that's a problem?

webapp/src/pages/registrationAdmin/components/RegistrationAdminComponent.js
KaleabTessera: This is applied per record and not a search for a list of records so it is not a prob.

KaleabTessera:
Confusion clarified.

PR #372, opened 2019-07-04
Let's hope we're nearly there

jaderabbit:


api/app/invitationletter/api.py
```diff
@@ -61,9 +61,12 @@ def post(self):
             if not registration:
                 return errors.OFFER_NOT_FOUND
         else:
-            # Normal registration
+            # Normal registratRegistrationion
```
avishkar58: ?

jaderabbit: I'm tired okay :P

Issue #336, opened 2019-06-22
Attendance Front End


jaderabbit:
On the day of an event, a registration admin should be able to access an event attendance page as per [Registration spec](https://docs.google.com/document/d/1dBTfmsBvWtCEw36mKYIgRsvMxE0BhMT-mU9iHVcWTrs/edit#)

The attendance page should list all of the people who have confirmed registrations for the event in a table (showing email, firstname, lastname, affiliation, user_category). (See #332 )

- The table should be sortable (sorted by lastname by default) and pageable  
- The table should have a search box that searches in any of the fields.
- A final column in the table should contain a button saying “Mark attendance” which will create a new attendance record and send a “welcome” email notification to the attendee (see #335 ).
- The registration_volunteer should get a visual notification when they mark someone’s attendance along with the ability to UNDO their action in the event that they mistakenly clicked the wrong button (also in #335 ).

KaleabTessera:
Done.

PR #415, opened 2019-08-16
Feature/attendance list front end

KaleabTessera:
Implementation of https://github.com/deep-learning-indaba/Baobab/issues/336. 

This will need quite a bit of testing on test-baboab, because there were quite a few situations that were hard to test locally.


api/app/attendance/api.py
```diff
@@ -85,8 +110,48 @@ def post(self):
...
+                RegistrationForm.event_id == event_id).first()
+            if(registration_form):
+                registration = db.session.query(GuestRegistration).filter(
+                    GuestRegistration.user_id == user_id).filter(GuestRegistration.registration_form_id == registration_form.id).first()
+
+            if not registration:
+                return OFFER_NOT_FOUND
+        else:
+            # Normal registration
+            registration = db.session.query(Registration).filter(
```
avishkar58: Some people may have not filled in their registration form, but we still want to be able to check them in - rather just return blanks or unknowns or something for the fields that come from the registration form, rather than an error.

api/app/registrationResponse/repository.py
```diff
@@ -45,3 +49,18 @@ def get_confirmed_for_event(event_id, confirmed):
         ).filter(
             Offer.event_id == event_id
         ).all()
+
+    @staticmethod
+    def get_confirmed_for_event_exclude_completed_registrations(event_id, confirmed):
+        """Get registrations for an event according to confirmed status, excluding those who have already registered."""
```
avishkar58: This is a little confusing because here I think you're using registered to mean "showed up at the event", but elsewhere registered just means "have confirmed they're coming and agreed to the code of conduct and filled in some details"

webapp/src/pages/attendance/components/AttendanceTable.js
```diff
@@ -0,0 +1,293 @@
...
+      }
+    }
+
+    let userInfoDiv = null;
+    if (confirmResult && confirmResult.confirmed) {
+      userInfoDiv = (
+        <div>
+          <h1>
+            {selectedUser.firstname} {selectedUser.lastname}
+          </h1>
```
avishkar58: I think there's too much information here, ideally this should be short and to the point, maybe just show:
The name and t-shirt size
and add like those bootstrap badge things for accommodation award or invited guest (showing the role) and shirt size, if that information is available, otherwise just don't display anything...

api/app/attendance/api.py
KaleabTessera: Done that.

webapp/src/pages/attendance/components/AttendanceTable.js
KaleabTessera: Updated this.

api/app/registrationResponse/repository.py
KaleabTessera: I have changed the name and made it more descriptive. This just gets people who have confirmed they are attending, but not signed in yet.

api/app/attendance/api.py
```diff
@@ -85,8 +113,53 @@ def post(self):
...
+
+        else:
+            # Normal registration
+            registration = db.session.query(Registration).filter(
+                Registration.offer_id == offer.id).first()
+
+        unavilable_response = None
+        if(registration is None):
+            is_guest_registration = unavilable_response
+            has_accepted_accom_award = unavilable_response
```
avishkar58: has_accepted_accom_award comes from offer though?

api/app/attendance/api.py
```diff
@@ -85,8 +113,53 @@ def post(self):
...
+            role = unavilable_response
+        else:
+            is_guest_registration = (not offer and registration)
+            event_role = db.session.query(EventRole).filter_by(
```
avishkar58: Sorry, this was meant to be the invited guest role (which you can find on the invited_guest table) - like "Speaker" or "Tutor", "Organiser" etc - this is mostly because Tutors and Organisers receive different T-shirts

api/app/attendance/api.py
KaleabTessera: Thanks for spotting this! Fixed.

api/app/attendance/api.py
KaleabTessera: Cool, I have changed this as well. Just confirming, are all tutors invited guests?

avishkar58: Thanks, looks good! 
So, one complication is that some people have offers AND are invited guests - most of the tutors fall in this category. So we should try to read the invited guest role irrespective of whether they have an offer or not.

PR #416, opened 2019-08-20
Feature/attendance list front end

KaleabTessera:
Things Done:
- Fixed shirt value.
- Allowed undo of signing in, even after the confirm page has been left. This is by going to 
by [BASE_URL]/eventAttendance?exclude_already_signed_in=false page and then clicking on the confirm for the  attendee. This will state that the user has been confirmed, but you can undo this.


KaleabTessera:
Fixing tests.

Issue #450, opened 2020-01-15
Create Organisation API


avishkar58:
Depends on #448 

Create an organisation API that allows clients to retrieve the current organisation name, small_logo, large_logo. The routes and GET methods should be set up in such a way that it's easy for the client to use the image in a natural way like `<img src="/api/app/organisation/small_logo"/>`

avishkar58:
Change to a get method that has the logo names that can be retrieved from front-end. @KaleabTessera let me know if you think there's a better approach?

Issue #448, opened 2020-01-15
Organisation middleware


avishkar58:
Create some middleware that runs before the controller on every route to intercept the request URL, retrieve the current organisation based on the domain and set it on the Flask g context for use by controllers. 

The organisation data should be cached rather than querying the database on every request. Create something to do this (eg a @before_first_request function, or something better that can also handle expiry)

Depends on #446

PR #455, opened 2020-01-19
Organisation Model, Middleware, API

### LINKED ISSUE #446: Create Organisation Model
Create a new model called Organisation with the following fields:
- id (int, primary key)
- name (string)
- small_logo (blob or string pointing to a file
- large_logo (blob or string pointing to a file) 
- domain (string)
- is_default (bool) 

Populate it with data for the Deep Learning Indaba:
- name: Deep Learning Indaba
- small_logo: Use the small logo we have in the top left corner of the menu bar
- large_logo: Use the large logo we have on the home page
- domain: deeplearningindaba.com
- is_default: True

Create a migration to apply the changes to the database.

avishkar58:
Resolves #446 #448 #450


api/migrations/versions/016571f41a20_add_organisation.py
```diff
@@ -0,0 +1,60 @@
...
+            'large_logo': 'logo_eeml_black.png',
+            'domain': 'eeml',
+            'is_default': False
+        },
+        {
+            'id': 3,
+            'name': 'Deep Learning Indaba',
+            'small_logo': 'logo-32x32-white.png',
+            'large_logo': 'indaba-logo-dark.png',
+            'domain': 'localhost',
```
jaderabbit: this for testing I assume?

avishkar58: Yeah, to stop the log spam during development... was going to do something with the app.yaml, but decided it's easier to just remove this record from the prod database :) Let me know if you prefer another approach though

api/app/organisation/resolver.py
```diff
@@ -0,0 +1,23 @@
...
+            organisations = OrganisationRepository.get_all()
+            cls._cache = {}
+            for org in organisations:
+                cls._cache[org.domain] = org
+                if org.is_default:
+                    cls._default = org
+        
+        if domain not in cls._cache:
+            LOGGER.warning('Could not find domain {} in cache, returning default organisation'.format(domain))
+            return cls._default
```
amritpurshotam: Should this return the default org? If for some reason an organisation isn't found, my feeling is that the site should break or at the very least return some user friendly error page (and the log should be at an error level since obviously something has gone really wrong). With the way it is now, the Deep Learning Indaba site will display potentially causing a lot of confusion to any users that might be there expecting some other organisation.

avishkar58: Yeah that's a good point! Have updated to remove the default option and make failure to resolve the organisation and error. Thanks!

avishkar58: Ah damn, now without the default organisation, all the tests are broken because they don't set the origin header 😂

api/app/__init__.py
```diff
@@ -27,15 +27,41 @@
...
 migrate = Migrate(app, db)
 
 manager = Manager(app)
 manager.add_command('db', MigrateCommand)
 
+from organisation.resolver import OrganisationResolver
+
+def get_domain():
+    # TODO: Remove this test-related hack!
+    if app.config['TESTING'] and 'HTTP_ORIGIN' not in request.environ and 'HTTP_REFERER' not in request.environ:
```
avishkar58: Very hacky solution. @KaleabTessera @jaderabbit @amritpurshotam - let me know if I'm maybe missing some obvious better solution? Otherwise will submit this PR in the interests of time and create an issue to improve this

amritpurshotam:
@avishkar58 I know this closed already but decided to look anyway. Just didn't get a chance to look before since I'm still fighting with my docker.
Otherwise the changes look good! Don't quite get the test hack bit though but looks like they all passing now

Issue #462, opened 2020-01-24
Resolve all Eslint warnings


bamtak:
When I run the program for the first time, I notice some Eslint warnings. 

> '===' and instead saw '==' , 
 'value' is defined but never used
Useless constructor
Arrow function expected a return value (some map and filter function returns null)

I have actually resolved all except the 

> href in < a />

 that I need access to some to see the styling before I make any update

bamtak:
Please can I have a dev username and password that I can use to test locally? I can't seem to get am email if I register locally.

avishkar58:
Hi Tajudeen,

Thanks for taking this!
When registering locally, keep an eye on your console to pick up your
verification link. The email gets logged there in development mode.

Thanks!

bamtak:
Thank you.
I have completed the issue and make a pull request

PR #463, opened 2020-01-24
Resolve all Eslint warnings #462

bamtak:


webapp/src/pages/home/Home.js
```diff
@@ -81,49 +80,42 @@ class Home extends Component {
...
-      table = this.state.rows ? (
-        <table align="center" className="Table">
-          <thead>{theadMarkup}</thead>
-          <tbody>{tbodyMarkup}</tbody>
-        </table>
-      ) : (<div></div>)
-    }
 
-    let statusClass = this.state.applicationStatus == "Submitted" 
+    // if (this.state.rows && this.state.rows.length > 0) {
```
KaleabTessera: Hey man. Why was this commented out?

webapp/src/pages/invitedGuests/components/InvitedGuestComponent.js
```diff
@@ -152,25 +152,16 @@ class InvitedGuests extends Component {
...
-        {
-          if( guest.role === roleSearch)
-          {
-            return guest;
-          }
-        }
-        else{
-          return guest;
-        }
+      return (fullname.toLowerCase().indexOf(value) > -1) && (roleSearch === "all" || guest.role === roleSearch)
```
KaleabTessera: This part needs to return a guest, this way it would only return a true/false value.

webapp/src/pages/registration/components/RegistrationComponent.js
```diff
@@ -509,20 +509,14 @@ class RegistrationComponent extends Component {
 
                 {section.registration_questions
                   .sort((a, b) => a.order - b.order)
-                  .map(question => {
+                  .filter(question => {
```
KaleabTessera: Hey man, what was the reason the filter was changed to filter + map?

webapp/src/pages/home/Home.js
bamtak: It was never used => decided to comment it out because maybe someone is working on it and plan to finish it out later. I comment on it with 'unused'

webapp/src/pages/invitedGuests/components/InvitedGuestComponent.js
bamtak: The callback function in filter is for testing if the element should be included or not
From mozila

> filter() calls a provided callback function once for each element in an array and constructs a new array of all the values for which callback returns a value that coerces to true

webapp/src/pages/home/Home.js
avishkar58: Yeah, the setState on rows was commented out a while ago, so the table didn't render.

webapp/src/pages/registration/components/RegistrationComponent.js
bamtak: It wasn't 'filter' that was changed but 'map' was change to 'filter' + 'map'

> map calls a provided callback function but based on the logic implemented on the code the function only returns  value (map) if a certain condition is satisfied (filter) else nothing (Eslint warnings => map is expected to return a value)

Issue #440, opened 2020-01-14
Add Email Templates


avishkar58:
Add an email template model and migration that has the following fields:
- id (primary key)
- key (string)
- event_id (int foreign key references event.id)
- template (string)

Add a unique constraint on (key, event_id)

Create a repository to enable easy querying of EventTemplate by key and event_id

PR #465, opened 2020-01-24
Add email templates table

amritpurshotam:
Resolves issue #440


amritpurshotam:
Looks like the test failed due to this error. I think this is fixed by re-running the build. Don't think I have access to do that 
```
Pulling db (orchardup/postgresql:la****)...
ERROR: unexpected EOF
```

api/app/email_template/repository.py
```diff
@@ -0,0 +1,13 @@
...
+from app.email_template.models import EmailTemplate
+
+class EmailRepository():
+
+    @staticmethod
+    def get_by_id(email_id):
+        return db.session.query(EmailTemplate).get(email_id)
+
+    @staticmethod
+    def get_by_key(key):
```
avishkar58: Should be by key and event_id

amritpurshotam: Oh yes of course 🙈
Fixed now.

amritpurshotam:
Regarding the null event_id, I think that's a great idea. Gives the best of both worlds then. The event organisers can use the existing templates if they think it's suitable or they can use their own custom ones.
Also just so we're on the same page, is the key essentially an email type? So would example keys look like `PasswordReset`, `EventOffer`, `InvitationLetter`, etc. And in code at the appropriate place, let's say in the password reset flow, I'll search for the `event_id `and `PasswordReset`? And then send the email that's retrieved.

avishkar58:
Yes exactly, although the specific example of `PasswordReset` isn't tied to an event, so we can leave that one (and the email verification ones in the user API) hard-coded. Everything else you'd retrieve by key and event_id when needed :)

api/app/email_template/repository.py
```diff
@@ -0,0 +1,9 @@
+from app import db
+
+from app.email_template.models import EmailTemplate
+
+class EmailRepository():
+
+    @staticmethod
+    def get(event_id, key):
+        return db.session.query(EmailTemplate).filter_by(event_id=event_id, key=key).first()
```
avishkar58: Think it's best to implement the fall-back logic here

amritpurshotam: Done

amritpurshotam:
Cool. I saw your comment in the other email template issue. Makes sense 👍

PR #466, opened 2020-01-24
Rich Text Input  

NileshArnaiya:
Markdown with Draft WYSIWYG editor


webapp/src/components/form/FormRichText.js
```diff
@@ -0,0 +1,61 @@
...
+
+class EditorContainer extends React.Component{
+  constructor(props){
+    super(props);
+    this.state = {
+      editorState: EditorState.createEmpty(),
+    };
+  }
+  onEditorStateChange = (editorState) => {
+    // console.log(editorState)
```
avishkar58: Remove

avishkar58:
Looks good to me, but let's wait for @KaleabTessera to review.

avishkar58:
Suggestion: try add this to applicationForm to test that it works as a component in the form.

NileshArnaiya:
So, I tried replacing FormTextArea with FormRichText and it looks good on the UI part but I'm not sure how to move ahead with the validation part. How do I add the same validation that is for FormTextArea to FormRichText? Thanks for guiding me.

KaleabTessera:
@NileshArnaiya Can you try having FormRichText exntend FormTextArea (inheritance) and then overwrite some of functionality? That way you can just use the validation alreadty built into FormTextArea.

NileshArnaiya:
Copy pasted functions from FormTextArea turns out those same functions wouldn't work for the EditorState FormRicH in DraftJS, same for showError and errorText in ApplicationForm, also wordcount needs to be fetched differently for it.

Issue #473, opened 2020-01-25
Add POST and PUT to event api


avishkar58:
Add post and put methods to the event api (api/app/events/api/py) that allow updating the fields of a single event (everything on the Event model in api/app/events/models.py). 

If the current user is not an event_admin for the event they are trying to modify, return a FORBIDDEN error.

s-mawjee:
@avishkar58  I'm keen to take this on.

avishkar58:
Great thanks! Let me know if anything is unclear :) at the moment there is
an EventsAPI which is more like an "Event List" API, so you might need to
create a new EventAPI for single events

s-mawjee:
Am I correct to assume that any user that `is_admin` can create an event? And on creation, the user is assigned as event admin?

avishkar58:
Hey, yes that's correct :)

PR #478, opened 2020-01-26
Issue 472

arunavkonwar:
1. Added following date fields to the event model:

application_open
application_close
review_open
review_close
selection_open
selection_close
offer_open
offer_close
registration_open
registration_close

2. Added flags for each process stage


api/app/events/models.py
```diff
@@ -12,14 +12,50 @@ class Event(db.Model):
     start_date = db.Column(db.DateTime(), nullable=False)
     end_date = db.Column(db.DateTime(), nullable=False)
 
+    application_open = db.Column(db.DateTime(), nullable=False)
+    appcliation_close = db.Column(db.DateTime(), nullable=False)
```
amritpurshotam: Typo in `application_close`

api/app/events/models.py
```diff
@@ -12,14 +12,50 @@ class Event(db.Model):
...
+                 offer_close,
+                 registration_open,
+                 registration_close
+                 ):
         self.name = name
         self.description = description
         self.start_date = start_date
         self.end_date = end_date
+        self.appcliation_open = application_open
+        self.appcliation_close = application_close
```
amritpurshotam: Typo in `self.application_open` and `self.application_close`

api/app/events/models.py
```diff
@@ -12,14 +12,50 @@ class Event(db.Model):
...
+    appcliation_close = db.Column(db.DateTime(), nullable=False)
+    review_open = db.Column(db.DateTime(), nullable=False)
+    review_close = db.Column(db.DateTime(), nullable=False)
+    selection_open = db.Column(db.DateTime(), nullable=False)
+    selection_close = db.Column(db.DateTime(), nullable=False)
+    offer_open = db.Column(db.DateTime(), nullable=False)
+    offer_close = db.Column(db.DateTime(), nullable=False)
+    registration_open = db.Column(db.DateTime(), nullable=False)
+    registration_close = db.Column(db.DateTime(), nullable=False)
+
```
amritpurshotam: These fields all look good however we're still missing the migration that will generate the script that will add these columns to the database when it's time for us to deploy. So once you fix the typos, run the following command

`docker-compose run web python ./api/run.py db migrate --directory api/migrations`

Assuming no errors occurred (generally this will happen if there's a syntax error in your code), this should generate the script. If you look in your project directory, you will see a new file has been added to `./app/migrations/versions`. You'll also see all the other files that have been generated previously.

Open the file and verify that in the `upgrade()` method the columns have been added and in the `downgrade()` method, the columns have been removed. The` upgrade()` is for when we deploy and the `downgrade()` is for if we decide to rollback a change on prod.

Once you're happy with the script, run the following command,

`docker-compose run web python ./api/run.py db upgrade --directory api/migrations`

This should now actually run the script that was generated and apply the changes to your local instance of your database. Pesonally I use [pgAdmin ](https://www.pgadmin.org/download/) as a sanity check to see in the database itself in a nice GUI if the changes I expected are actually there. 

Also in case you need to reference these commands again in the future, have a look at the `./api/README.md`. I know I still look at it all the time :)

api/app/events/api.py
```diff
@@ -26,7 +26,12 @@ def event_info(user_id, event):
         'description': event.description,
         'start_date': event.start_date.strftime("%d %B %Y"),
         'end_date': event.end_date.strftime("%d %B %Y"),
-        'status': get_user_event_response_status(user_id, event.id)
+        'status': get_user_event_response_status(user_id, event.id),
+        'is_application_open': True if (datetime.now() < event.end_date and datetime.now() >= event.start_date) else False,
+        'is_review_open': True if (datetime.now() < event.review_close and datetime.now() >= event.review_open) else False,
+        'is_selection_open': True if (datetime.now() < event.selection_close and datetime.now() >= event.selection_open) else False,
+        'is_offer_open': True if (datetime.now() < event.offer_close and datetime.now() >= event.offer_open) else False,
+        'is_registration_open': True if (datetime.now() < event.registration_close and datetime.now() >= event.registration_open) else False
```
amritpurshotam: Can we move this logic into the Event model? The reason is because I think these calculated fields may also be useful later down the line for future checks and I'd like to avoid the logic being duplicated (note we're not adding these fields to the db though). So the way I would imagine you could do it would be by using `@property`. The code would look like the following then

```
@property
def is_application_open(self):
    now = datetime.now()
    return now >= self.start_date and now < self.end_date
```
Then you will be able to call this calculated field like any other _actual_ field like so `event.is_application_open` (note no brackets)

To then pass these fields down to the front end, you can make the API return these fields like all the others, and the front end can just accept those fields as accurate without having to recalculate them again.

api/app/events/models.py
arunavkonwar: Thank you very much for this detailed review. Appreciate it. I have made the changes. :)

api/app/events/api.py
arunavkonwar: This is another incredible piece of new information that I was never aware of. Thank you for taking the time to explain it to me.

api/app/events/models.py
arunavkonwar: @amritpurshotam  I believe a lot of your above replies can be incorporated in the development docs for newcomers and everyone else. @avishkar58

arunavkonwar:
@amritpurshotam resolved merge conflicts from issue #473 (pull request #479 ). Please have a look at this PR

api/migrations/versions/b7006cbdd923_.py
```diff
@@ -0,0 +1,54 @@
...
+    op.add_column('event', sa.Column('offer_open', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('registration_close', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('registration_open', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('review_close', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('review_open', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('selection_close', sa.DateTime(), nullable=False))
+    op.add_column('event', sa.Column('selection_open', sa.DateTime(), nullable=False))
+    op.alter_column('organisation', 'system_name',
+               existing_type=sa.VARCHAR(length=50),
+               nullable=False)
```
amritpurshotam: This part doesn't look quite right. I think your local database might be slightly out of date. I think when these scripts are generated, it compares what your database looks like currently and what the Python models look like and then does the diff. So I think when you pulled the latest changes, and ran the migration, it added this extra field. In the future, I think make sure to run upgrades whenever you pull latest changes, then run your migration, then the upgrade again. This can get tricky if you've already applied your changes which means you'll first have to downgrade them, delete the script, upgrade, migrate, then upgrade again. But don't worry about this for now. I'll note this down as something to explain in the docs like you mentioned in another comment. 

For now just backspace this line

api/migrations/versions/b7006cbdd923_.py
```diff
@@ -0,0 +1,54 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('organisation', 'system_name',
+               existing_type=sa.VARCHAR(length=50),
+               nullable=True)
```
amritpurshotam: And backspace this line too

amritpurshotam:
> @amritpurshotam resolved merge conflicts from issue #473 (pull request #479 ). Please have a look at this PR

I don't see those changes here. In fact Shahil abandoned his fork and is going to work directly off the Baobab repo now (and so that branch isn't going to be merged at all now). So not sure now. Did you merge his changes?

arunavkonwar:
@amritpurshotam Ah I think that was the PR/Issue that was causing merge conflicts with my last PR. Initially, it was fine then some changes by someone else got merged in the main branch and mine started having conflicts.

Anyhow, I have made the requested changes. Hopefully, this resolves this issue. And thank you again for explaining the underlying issues that might have affected the migration file.

amritpurshotam:
@arunavkonwar Sure you welcome!

amritpurshotam:
@arunavkonwar Also I just realised something now. It doesn't look like our tests run on your fork, I think for subsequent issues can you clone and work from the Baobab repo directly? You should have access to create your own branches and push to that.

arunavkonwar:
@amritpurshotam Yes! I'll work from the Baobab repo directly from now.

Issue #472, opened 2020-01-25
Add process stage dates to event 


avishkar58:
Add the following date fields to the event model (api/app/events/models.py)

application_open
application_close
review_open
review_close
selection_open
selection_close
offer_open
offer_close
registration_open
registration_close

You can normalise this further if you think that's better (eg using a ProcessStage entity with event_id, name, open,  close)

Update the event api get method to return these dates along with flags corresponding to whether the current date lies between the open and close date for each stage. For example, an is_application_open flag that is true if the current date is between application_open and application_close dates.

This will be used to toggle pages/menu items etc on the front end (which is why it's better to resolve the flags on the backend)

PR #479, opened 2020-01-26
#473 Event API post and put

s-mawjee:
Added POST and PUT in EventAPI


api/app/events/api.py
```diff
@@ -96,13 +99,86 @@ def get(self):
...
+        end_date = datetime.strptime(
+            (args['end_date']), '%Y-%m-%dT%H:%M:%S.%fZ')
+
+        event = Event(name=name,
+                      description=description,
+                      start_date=start_date,
+                      end_date=end_date)
+
+        db.session.add(event)
+        db.session.commit()
```
amritpurshotam: Can we add a new method called `add(event: Event)` to the ./api/event/repository? And then import the method in. Personally I prefer to push lower level operations down and away from an API level. I think it looks it makes the logic clearer and allows re-usability of the methods. What do you think?

api/app/events/api.py
```diff
@@ -96,13 +99,86 @@ def get(self):
...
+
+        try:
+
+            admin_event_role = EventRole('admin', user_id, event.id)
+            db.session.add(admin_event_role)
+            db.session.commit()
+        except Exception as e:
+            LOGGER.error(
+                'Failed to add event role for user id {} due to: {}'.format(user_id, e))
+            return errors.ADD_EVENT_ROLE_FAILED
```
amritpurshotam: Similar reasons here. I think there's two places this code can be pushed down closer to it's domain objects. 

The first I think can be into a generic `add_role(event_id, role)` method in the AppUser model. It looks like the `event_roles` field is already there so you would create the `EventRole` object and append it to `current_user.event_roles`. A db.session.commit should then save it (I think, just test and double check)?

The second place you could do it would be doing something similar but on the Event object you created above (but you would modify it before you saved). It looks like you might still need to add the `event` field and `event_roles` field (use `AppUser` and `EventRoles` as an example). Again here you would append `EventRole` to the `event_roles` field. And then when you save to the db, the `Event` and `EventRole` should both appear with one call. 

Let me know what you think and if I made sense with my explanation.

api/app/events/api.py
```diff
@@ -96,13 +99,86 @@ def get(self):
...
+        end_date = datetime.strptime(
+            (args['end_date']), '%Y-%m-%dT%H:%M:%S.%fZ')
+
+        event = db.session.query(Event).filter(
+            Event.id == event_id).first()
+
+        event.name = name
+        event.description = description
+        event.start_date = start_date
+        event.end_date = end_date
```
amritpurshotam: Can we have an update() method on the Event model with these fields that we're updating? Again just to move this logic out of the API.

amritpurshotam:
Also just a heads up @s-mawjee, looks like there's a dependency on PR #478 (issue #472 ). There's a lot of new fields being added there so not sure if you should wait for that one to be closed first? cc @arunavkonwar

api/app/events/api.py
s-mawjee: Sounds good. I like this design pattern. I see that this might need to be applied to the other APIs, such as users.

api/app/events/api.py
s-mawjee: I will investigate this and get back to you.

api/app/events/api.py
s-mawjee: I will add an update method.

PR #482, opened 2020-01-26
Fix import error and update tests files

bamtak:


api/app/events/api.py
```diff
@@ -7,26 +7,33 @@
...
+from app.utils.errors import EVENT_NOT_FOUND, FORBIDDEN, EVENT_WITH_KEY_NOT_FOUND
 
 from app.utils.auth import auth_optional, auth_required
 from app.utils.emailer import send_mail
+from app.events.repository import EventRepository as event_repository
+from app.organisation.models import Organisation
 
 
-def event_info(user_id, event):
+def event_info(user_id, response):
```
avishkar58: Would change the name here to avoid confusion - took me a while to figure out it wasn't an HTTP response, but rather one of the named tuples that come back from the repository :D Maybe `event_org` ?

bamtak: Yes, that's better. I have changed the name

avishkar58:
Ah I see you're developing in a fork, that's why the tests aren't running on your PRs... Fine for now, but maybe work on a branch inside this repository for future changes :) 

Have you run all tests locally and confirmed they're passing?

bamtak:
Aah! Yes, thank you for pointing that out.
I have run all the test locally and they all passed

PR #490, opened 2020-01-27
Fix migration

amritpurshotam:
Changed down_revision of script so that the alembic graph is linear again.

So I did this manually by tracing the path of the revisions and just setting this one script to be the last one in the path but I wanted to use the commands in this [article ](https://blog.jerrycodes.com/multiple-heads-in-alembic-migrations/)but I couldn't get them to run. @avishkar58 Do you know how to go about doing this? I see you did a successful merge in file `32c1ae6dd23a.py`

Also updated the tests that need an Event constructor. Just did a quick and dirty solution now so everyone elses PRs can go green again. I think I'll make a separate issue to clean up these tests.


avishkar58:
I was able to do a merge migration of the two latest heads using:
`docker-compose run web python ./api/run.py db merge --directory api/migrations heads`

amritpurshotam:
Ah that makes sense! Thanks. Something else I'll add to the docs :)

PR #496, opened 2020-01-28
#473 - EventAPI POST and PUT

s-mawjee:


api/app/events/api.py
```diff
@@ -78,16 +80,173 @@ def _log_application_status(context):
...
+            LOGGER.error("Event with KEY: {} already exists".format(key))
+            return EVENT_KEY_IN_USE
+
+        try:
+            admin_event_role = event_role_repository.add(
+                'admin', user_id, event.id)
+        except Exception as e:
+            LOGGER.error(
+                'Failed to add event role for user id {} due to: {}'.format(user_id, e))
+            return ADD_EVENT_ROLE_FAILED
```
amritpurshotam: So I want to try to avoid this separate save to the db and rather try to save the event and event role at once. Can you try this? Setup the relationship between `Event` and `EventRole` by going to the `Event` model and adding a field called `event_roles = db.relationship('EventRole')` and in `EventRole` adding `event = db.relationship('Event'. foreign_keys=[event_id])`.

In the Event model, add the following method (you may need to initialise `event_roles` if it's `None` here or in the constructor)
```
def add_event_role(event_role: EventRole):
    self.event_roles.append(event_role)
```
Back in this file you can create your `EventRole`, and add it to the `Event` by running the `add_event_role()` method and then run `event_repository.add(event)`

api/app/events/repository.py
```diff
@@ -19,12 +19,28 @@ def exists_by_key(event_key):
...
+
+
+class EventRoleRepository():
+
+    @staticmethod
+    def add(role, user_id, event_id):
+        event_role = EventRole(role, user_id, event_id)
+        db.session.add(event_role)
+        db.session.commit()
+        return event_role
```
amritpurshotam: If the other code above works, this can then come out

api/app/events/api.py
s-mawjee: This change works.
Just a small adjustment I made was, the inputs to `add_event_role` are just the `role` and `user_id` 

```
def add_event_role(self, role, user_id):
    event_role = EventRole(role, user_id, self.id)
    self.event_roles.append(event_role)
```

amritpurshotam: Oh yes of course. This is right. The event_id should just auto-populate (and yeah it doesn't event exist at this point)

PR #500, opened 2020-01-28
Event frontend

### LINKED ISSUE #453: Organisation and Event front-end changes
Depends on organisation and event backend work

In order to support both multi-tenancy (through organisations) and multi-event (through events linked to organisations), we need to make the following changes to the structure of the front-end. 
 
- Move the existing front-end home component to be an "event-home"
- Update routes so that an event_key identifies and event (for example, baobab.deeplearningindaba.com/indaba2020 should go to the home page for the Deep Learning Indaba 2020 event)  
- Update other routes like "apply" so that <event_key>/apply identifies the apply page for the Indaba 2020 event
- Create a new organisation home page. This page should have no top-level menus which are now all event-based. 
- The organisation home page should query the event_api to get a list of upcoming events, with links to the home pages for those events.

avishkar58:
Resolves #453 
Handling multi-event on front end

The application statuses on the event home page aren't quite right, but want to get this out to review and submitted asap and will create another issue to look into that.


webapp/src/pages/home/Home.js
```diff
@@ -1,207 +1,88 @@
...
+                }
+                this.setState({
+                    events: response.events,
+                });
+            });
         }
-        else {
-          applicationStatus = "NOT Submitted";
+        
+        if (this.props.setEvent) {
```
KaleabTessera: Why is this being done?

avishkar58: The check is so the test doesn't break when setEvent is null. The calling setEvent with nulls is to "reset" the navbar in App.js when the user navigates away from a specific event to the organisation home page.

webapp/src/pages/eventHome/EventHome.js
```diff
@@ -0,0 +1,340 @@
...
+  
+                    {this.state.applicationStatus === "Submitted" && this.state.offer && <span>
+                      Your application was successful! 
+                    </span>}
+  
+                    {this.state.applicationStatus === "Submitted" && !this.state.offer &&  this.state.invitedGuest !== true && <span>
+                      Waiting List
+                    </span>}
+  
+                    {this.state.applicationStatus !== "Submitted" && this.state.invitedGuest === true && <span>
```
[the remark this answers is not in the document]
avishkar58: Quite Possibly! Haven't gone through this logic though - just copied it from the previous Home.js - will create a new issue to check it properly :)

PR #504, opened 2020-01-28
Add ReferenceRequest entity and API 497

bamtak:


bamtak:
Please note that I have not done the migration for ReferenceRequest.

amritpurshotam:
Hi @bamtak. Any particular reason you haven't been able to do the migration? 
If you're not sure of what to do, have a look at this [link](https://github.com/deep-learning-indaba/Baobab/pull/478#discussion_r371015441) which explains the process. Just be sure to actually run the upgrade command first to make sure your local db is up to date, then the migrate, then the upgrade again. Then test if everything is working as expected

bamtak:
@amritpurshotam Thank you. MIgration is now created for the reference request entity.

api/app/references/api.py
```diff
@@ -0,0 +1,101 @@
...
+        LOGGER.debug('Received get request for reference-request')
+        args = self.req_parser.parse_args()
+        return reference_request_repository.get_all_by_response_id(args['response_id']), 200
+
+
+
+class ReferenceRequestFormAPI(ReferenceRequestsFormMixin, restful.Resource):
+
+    @auth_required
+    def post(self):
```
avishkar58: Why not put this post method on the ReferenceRequestAPI? That way the routing can also be consistent with REST and with other APIs (ie you can GET, POST, PUT etc on an entity which is the reference-request)

api/app/references/api.py
```diff
@@ -0,0 +1,101 @@
...
+        LOGGER.debug('Received get request for reference-request')
+        args = self.req_parser.parse_args()
+        return reference_request_repository.get_by_id(args['id']), 200
+
+    
+class ReferenceRequestListAPI(ReferenceRequestsListMixin, restful.Resource):
+
+    @marshal_with(reference_request_fields)
+    @auth_required
+    def get(self):
```
avishkar58: Let's add some validation here to check that the response belongs to the authenticated user, otherwise we could have users querying for other people's reference request data.

api/app/references/repository.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ReferenceRequestRepository():
+    
+    @staticmethod
+    def get_by_id(reference_request_id):
+        return db.session.query(ReferenceRequest).get(reference_request_id)
+
+    @staticmethod
+    def get_event_by_response_id(response_id):
```
avishkar58: This seems a bit out of place, should be on the event or response repository..

api/app/references/repository.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @staticmethod
+    def create(reference_request):
+        reference_request.set_token(str(uuid.uuid4()))
+        db.session.add(reference_request)
+        db.session.commit()
+    
+    @staticmethod
+    def update(reference_request):
+        db.session.add(reference_request)
```
amritpurshotam: Are we sure this doesn't add another row to the table?

bamtak: Well, that depends on the entity passed to the function. If the entity has no id, it will definitely create a new row but if there is id and exist it will merge. 
But I also added an assertion on POST request for that.

bamtak:
@avishkar58 Please can you take a look at this, I can't seem to merge the PR

PR #508, opened 2020-01-29
Move emails related to applications

### LINKED ISSUE #441: Replace hard-coded email templates
Search the back-end code for hard-coded email templates. Add these templates to the emailTemplate table, assigning a key to the template. Replace the template with a query to the table using the current event_id and new key.

Depends on #440

amritpurshotam:
So I undid the moving of the emails that weren't related to applications. Realised it was easier to fix the tests one at a time. 

Partially resolves #441. Will continue moving the other emails after this closing this PR so it's not blocking any other work.


api/app/responses/tests.py
```diff
@@ -85,270 +94,268 @@ def _seed_data(self):
...
+        self.assertEqual(data['application_form_id'], self.test_form.id)
+        self.assertEqual(data['user_id'], self.other_user_data['id'])
+        self.assertIsNone(data['submitted_timestamp'])
+        self.assertFalse(data['is_withdrawn'])
+        self.assertTrue(data['answers'])
+
+        answer = data['answers'][0]
+        self.assertEqual(answer['id'], self.test_answer1.id)
+        self.assertEqual(answer['value'], self.test_answer1.value)
+        self.assertEqual(answer['question_id'], 1)
```
amritpurshotam: So I kept getting an error along the lines of
`DetachedInstanceError: Instance <Event at 0x7fdb85c18710> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/bhk3)` but only if I ran the tests all at once. Running a test one at a time, they didn't break.

This is why I changed this whole file. Seems it's something to do with the objects not being loaded correctly anymore. Removing the app.context() and hard coding some of the values fixed the issue. Probably not the best way of fixing it but was a bit stumped so if someone knows the proper fix, let me know.

api/app/events/api.py
```diff
@@ -388,12 +366,21 @@ def post(self):
             firstname = user.firstname
             lastname = user.lastname
             event_name = event.name
+            organisation_name = event.organisation.name
+            system_name = event.organisation.name
```
avishkar58: There is a system_name property on organisation :)

api/app/responses/tests.py
avishkar58: I had the same issue the other day, and also couldn't figure out what was causing it :( @jaderabbit any idea?

api/app/events/api.py
amritpurshotam: Oh yes I saw that 🙈 Think I was on autopilot. Good catch!

PR #509, opened 2020-01-30
Updated organisation email from

avishkar58:


api/app/utils/emailer.py
```diff
@@ -25,13 +24,16 @@ def send_mail(recipient, subject, body_text='', body_html='', charset='UTF-8', m
     Raises:
         e -- [description]
     '''
+    sender_name = sender_name or SMTP_SENDER_NAME
+    sender_email = sender_email or SMTP_SENDER_EMAIL
+
```
amritpurshotam: Is it possible that the email in the organisation table will differ from the one in the config? Because if so, won't that cause confusion if event level emails are sent from a different address?

avishkar58:
So ideally the sender_email and sender_name should be set for every email
that gets sent. In particular for EEML it should always be contact@eeml.eu
and for Indaba it should always be baobab@deeplearningindaba.com ... The
default was really just a quick solution to prevent errors and all the
tests failing...

PR #516, opened 2020-02-03
EEML Event and Application Form Config

### LINKED ISSUE #503: Populate EEML Event and Application form configuration
Create a new migration to add the following configuration to the database:

Event:
- organisation_id: 2   (EEML)
- name: Eastern European Machine Learning Summer School 2020
- description: Eastern European Machine Learning Summer School 2020, 6-11 July 2020, Krakow, Poland
- start_date: 6 July 2020
- end_date: 11 July 2020
- email_from: contact@eeml.eu
- key: eeml2020
- url: http://eeml.eu
- application_open: 28 January 2020
- application_close: 20 March 2020
Add any future dates for the other open and close dates for now

Application Form:
Create a new application form linked to the previously created event, set is_open to True and the deadline to 20 March 2020

Add sections and questions to the application form according to [this document](https://docs.google.com/document/d/1ZOL4E3sQCS8vCCHCLKQZUWwPHpepxW1NHQyPRWnr6kQ/edit?usp=sharing)

Take note of the following:
- The options field on the Question model should be a list of python dictionaries, with a label and value. So for example, to insert `"Yes"; "No"`  as options for a question, you would use `[{"value": "yes", "label": "Yes"}, {"value": "no", "label": "No"}]`
- The yellow highlighted items in the question tables are instructions to you, for example, for the country list, you'll need to create a list of countries to insert into the options field and for the validation you'll need to write a little **regex** to apply the validation. Everything else can be interpreted as a literal.

TIPS:
- To generate a new blank migration, run `docker-compose run web python ./api/run.py db migrate --directory api/migrations` in your terminal. This will create a new file in the api/migrations/versions folder. 
- Have a look at the [migration used to configure the Indaba 2019 application form](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/migrations/versions/b3f7a73c98f3_.py) to see how it was done before. But keep in mind that the [models](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/applicationModel/models.py) have been updated since then!

avishkar58:
Resolves #503

### DIFF
--- a/api/migrations/versions/87213d612eaf_eeml_application_form.py
+++ b/api/migrations/versions/87213d612eaf_eeml_application_form.py
@@ -20,6 +20,31 @@ import datetime
 
 Base = declarative_base()
 
+class Organisation(Base):
+
+    __tablename__ = "organisation"
+    __table_args__ = {'extend_existing': True}
+
+    id = db.Column(db.Integer(), primary_key=True)
+    name = db.Column(db.String(50), nullable=False)
+    system_name = db.Column(db.String(50), nullable=False)
+    small_logo = db.Column(db.String(100), nullable=False)
+    large_logo = db.Column(db.String(100), nullable=False)
+    domain = db.Column(db.String(100), nullable=False)
+    url = db.Column(db.String(100), nullable=False)
+    email_from = db.Column(db.String(100), nullable=True)
+    system_url = db.Column(db.String(100), nullable=False)
+
+    def __init__(self, name, system_name, small_logo, large_logo, domain, url, email_from, system_url):
+        self.name = name
+        self.small_logo = small_logo
+        self.large_logo = large_logo
+        self.domain = domain
+        self.system_name = system_name
+        self.url = url
+        self.email_from = email_from
+        self.system_url = system_url
+
 class Country(Base):
     __tablename__ = "country"
     __table_args__ = {'extend_existing': True}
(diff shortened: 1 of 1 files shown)


api/migrations/versions/87213d612eaf_eeml_application_form.py
```diff
@@ -0,0 +1,327 @@
...
+    session.commit()
+
+    app_form = ApplicationForm(eeml2020.id, True, datetime.date(2020, 3, 20))
+    session.add(app_form)
+    session.commit()
+
+    main_section = Section(app_form.id, 'EEML 2020 Application Form', """
+This is the official application form to apply for participation in EEML 2020, 6-11 July, Krakow, Poland. 
+
+Do not forget to press Submit once you fill in the application form. It is possible (and recommended!) to submit partially completed forms early on to get familiar with the form, and come back at a later date before the application deadline (20 March 2020, 23:59 Eastern European Time) to edit/complete your responses.
```
amritpurshotam: With all the dates we're setting now, something I noticed in the code base is that all the checks do datetime.now(). I think it's mostly fine since most of our applicants are in the same time zone where the code is deployed (which is Europe right?) but for Khipu this might become an issue. So we might have to start thinking about introducing time zones? Or I suppose a low tech solution would be to figure out the time in Europe for midnight of the regions the event is taking place.

api/migrations/versions/87213d612eaf_eeml_application_form.py
```diff
@@ -0,0 +1,327 @@
...
+    eeml2020 = Event('EEML 2020', 'Eastern European Machine Learning Summer School 2020, 6-11 July 2020, Krakow, Poland',
+        datetime.date(2020, 7, 6), datetime.date(2020, 7, 11), 'eeml2020', 2, 'contact@eeml.eu',
+        'http://eeml.eu', datetime.date(2020, 2, 2), datetime.date(2020, 3, 20), datetime.date(2020, 3, 20), 
+        datetime.date(2020, 4, 20), datetime.date(2020, 4, 20), datetime.date(2020, 5, 20), datetime.date(2020, 5, 20),
+        datetime.date(2020, 6, 20), datetime.date(2020, 6, 20), datetime.date(2020, 6, 30))
+
+    session.add(eeml2020)
+    session.commit()
+
+    app_form = ApplicationForm(eeml2020.id, True, datetime.date(2020, 3, 20))
```
amritpurshotam: Seems we have some duplication now in the ApplicationForm deadline field. This is the same as the application_close field in Event. Should we clean this up? We can maybe make a refactor issue?

avishkar58: Have made an issue to clean this up (#518) - thanks :)

api/migrations/versions/87213d612eaf_eeml_application_form.py
avishkar58: yeah good point... timezones are always such a rabbit hole though! :D

PR #536, opened 2020-02-06
Add reference entity and api

bamtak:


api/app/references/api.py
```diff
@@ -105,3 +114,41 @@ def get(self):
...
+
+
+class ReferenceAPI(ReferenceMixin, restful.Resource):
+
+    @marshal_with(reference_fields)
+    @auth_required
+    def get(self):
+        LOGGER.debug('Received get request for reference')
+        user = user_repository.get_by_id(g.current_user['id'])
+        if not (user or user.is_admin):
```
avishkar58: The `is_admin` flag indicates the user is a system admin. To determine if a user is an event admin, use the `is_event_admin` method instead (you'll also need an event_id parameter from the request)

api/app/references/api.py
```diff
@@ -105,3 +114,41 @@ def get(self):
...
+            return FORBIDDEN
+
+        args = self.get_req_parser.parse_args()
+        response = response_repository.get_by_id_and_user_id(args['response_id'], user.id)
```
avishkar58: Not sure why the response repository has a user_id parameter when the response_id is enough to unique identify the record... Can you add another method that's just get_by_id and use that?

api/app/references/models.py
```diff
@@ -34,9 +35,29 @@ def __init__(self,
         self.lastname = lastname
         self.relation = relation
         self.email = email
+        self.reference_submitted = False
```
avishkar58: I understand why you've added this, but I'm not the biggest fan of having a flag like this as it can easily become out of sync with the "truth" which is that a reference exists if and only if a record exists in the Reference table with a corresponding `reference_request_id`, so would prefer if the logic checked explicitly for that, even though it's another DB query.  What do you think  @amritpurshotam  ?

api/app/references/api.py
bamtak: Gotcha, I agree that is totally not necessary.

amritpurshotam: @avishkar58 @bamtak Is it not to prevent a user from looking up some other users response?

api/app/references/models.py
amritpurshotam: I agree. @bamtak I think setup a relationship between `Reference` and `ReferenceRequest` and then you will be able to check if a reference exists. So on `ReferenceRequest` add the following line 
`references = db.relationship('Reference')` 
and on `Reference` add 
`reference_request = db.relationship('ReferenceRequest', foreign_keys=[reference_request_id])`

api/app/references/api.py
```diff
@@ -105,3 +114,41 @@ def get(self):
...
+        args = self.post_req_parser.parse_args()
+        token = args['token']
+        uploaded_document = args['uploaded_document']
+        reference_request = reference_request_repository.get_by_token(token)
+        if not reference_request:
+            return REFRERENCE_REQUEST_WITH_TOKEN_NOT_FOUND
+
+        reference = reference_request_repository.get_reference_by_reference_request_id(reference_request.id)
+        if reference:
+            return DUPLICATE_REFERENCE_SUBMISSION
```
amritpurshotam: With the relationships setup between ReferenceRequest and Reference, on reference_request you will be able to check if an item exists in `reference_request.references` and then return the duplicate error. Even better you can perhaps wrap this check in a method in the ReferenceRequest class called `has_reference()`

api/app/references/models.py
```diff
@@ -34,9 +35,29 @@ def __init__(self,
...
+
+
+
+class Reference(db.Model):
+
+    __tablename__ = "reference"
+
+    id = db.Column(db.Integer(), primary_key=True)
+    reference_request_id = db.Column(db.Integer(), db.ForeignKey('reference_request.id'), nullable=False)
+    uploaded_document = db.Column(db.String(128))
```
amritpurshotam: Set `nullable=False` unless is there a reason it should be nullable?

api/app/references/tests.py
```diff
@@ -66,6 +66,29 @@ def test_get_reference_request_by_response_id(self):
         reference_request_repository.create(reference_req2)
         response = self.app.get(
             '/api/v1/reference-request/list', data={'response_id':1}, headers=self.headers)
+        LOGGER.debug(response.data)
```
amritpurshotam: Same here. Remove.

api/app/references/tests.py
```diff
@@ -66,6 +66,29 @@ def test_get_reference_request_by_response_id(self):
...
+        assert response.status_code == 201
+
+        response = self.app.get(
+            '/api/v1/reference', data={'response_id':1}, headers=self.headers)
+        LOGGER.debug(response.data)
+        data = json.loads(response.data)
+        assert response.status_code == 200
+        assert len(data) == 1
+        reference_request = reference_request_repository.get_by_id(1)
+        assert reference_request.reference_submitted == True
```
amritpurshotam: For all the asserts can we replace them with `self.assertEqual(first, second)`. The reason is that when the tests break the message is more helpful i.e. this method will show each of the values. `assert` just says the value is wrong which isn't as helpful.

api/app/references/models.py
amritpurshotam: Also make sure to run another migration to make sure this column gets dropped

api/app/references/models.py
amritpurshotam: Also remember to run another migration if you do set this to not nullable

api/app/references/api.py
bamtak: I am leaving this part @avishkar58 what do you think.

avishkar58: Sure, think that's fine for now, thanks @bamtak !

bamtak:
@avishkar58 @amritpurshotam Any update on this pull request.

amritpurshotam:
@bamtak Sorry about the delay. Just got a bit busy this weekend. I'll have a look tomorrow morning.

amritpurshotam:
Thanks, nice work @bamtak

bamtak:
@avishkar58 Please check this

PR #538, opened 2020-02-06
Fix #179 Create Indexes on selected tables

### LINKED ISSUE #179: Add database indexes
- Look at the queries we're doing on some of the larger tables in our controllers (AppUser, Response, Answer in particular) and determine whether any DB indexes would be useful for these queries (eg. (application_form_id, user_id) for the Response table) 
- Create a migration to add these indexes to the database

**Definition of done**
- The new migration has been created with upgrade and downgrade support
- The migration can be applied (upgrade and downgrade) successfully to the test database

espoirMur:
This fixes the issue [179](https://github.com/deep-learning-indaba/Baobab/issues/179) where we were required to create indexes on some tables.

For the first attempt, indexes were added on foreign keys. According to this [So Answer](https://stackoverflow.com/a/11985423/4683950)

Performance of the index is not tested yet if you know how we can test it let me know!


api/README.md
```diff
@@ -77,9 +80,21 @@ This API uses token-based authentication. A token is obtained by registering a n
...
+    "nationality_country_id":"243",
+    "residence_country_id":"243",
+    "user_gender":"M",
+    "department":"Computer Science",
+    "user_disability":"none",
+    "user_category_id":1,
+    "affiliation":"Home",
+    "user_primaryLanguage":"english", 
+    "user_dateOfBirth":"2020-02-06T14:49:16.725Z",
+    "password": "***********"
```
KaleabTessera: Currently this only requires the following fields -title, firstname, lastname, email address - please remove the rest.

amritpurshotam:
So I think the stack over flow post you linked does contain sound advice although I disagree with the order in which you should carry out the advice. The last line about over indexing is the most critical part to be aware of since performance degradation is a factor but also the more indexes you have the faster the size of the database grows (increasing storage costs, making backups take longer, etc).

The next most critical piece is to observe the performance. I haven't done this in postgres before but database engines do have ways of tracking stats such as the longest running query overall or even the query that takes up the most amount of time after factoring in how often it's used (for example there might be a query that takes 10 mins to run once a month but a query that runs 100000 times with each run taking 1s, thats the one you would want to optimise). Once you have these stats then you tackle each bottleneck one by one. In my experience once you optimise the first one or two biggest issues, you're golden. 

So with that being said, @avishkar58 @KaleabTessera do we have stats on what has been the main bottlenecks in the database?

api/README.md
espoirMur: Thanks, @KaleabTessera for the feedback, The reason why I put the rest it's because when I was trying the endpoint it was. refusing and returning a 400 error without those fields!

KaleabTessera:
@amritpurshotam  Good point. I will see if I can get some query stats to guide where to add indexes.

api/README.md
KaleabTessera: Cool, please let me know if this is still an issue. I will fix it if it is.

KaleabTessera:
@amritpurshotam @avishkar58 So I did a bit of research and it seems like logging slow queries for "cloud sql for Postgres" is not yet supported (https://issuetracker.google.com/issues/74578509). 
Until that is supported, it will be hard to evualate what queries are slow. 

Personally, I think adding indexes to most tables will be fine. Addding indexs only really hurt when you have instensive writes to a db, which I don't really think we will have. 

For this PR specifically, I think the added indexs are fine. We will still probably have to adjust the tables to cater for our discussions on reviewing as per yesterday though.

amritpurshotam:
@KaleabTessera @avishkar58 @espoirMur Hmm okay I suppose then we don't really have a choice. But I still don't think adding indexes anyway is the answer. We should only optimise if we see some slowness in the application

Perhaps we can profile the application as a whole? I'll look to see if there's something we can use. I'm thinking of something along the lines of MS Application Insights but a python/google cloud equivalent.

amritpurshotam:
And sorry @espoirMur 🙈 I know you've done the work already as stated on the issue and I'm only now questioning whether this is the right solution.

KaleabTessera:
Yeah sorry @espoirMur .

We could try and profile API calls and see which ones are slow, then dig down into those. @amritpurshotam

avishkar58:
Ah, this is my fault, didn't think it through fully when I created the issue. Agree with @amritpurshotam . Sorry for wasting your time on this @espoirMur !

PR #539, opened 2020-02-06
Bug fixes

### LINKED ISSUE #527: Confirmation email no longer has a hyperlink
The verification email link has to be copied and pasted into browser and does not have a hyperlink

### LINKED ISSUE #528: No indication of preferred file format for uploads
- When uploading the CV or research abstract, the file upload doesn't accept a Word document. 
- This results in the following error:

image

- Clicking on “get in touch” results in a redirect to:
 baobab@deeplearningindaba.com with the error code Invariant Violation: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31

### LINKED ISSUE #531: Make privacy policy organisation-specific
Different organisations will need different privacy policies, due to differences in who and how they share data. 

- Add a field to the organisation model called "privacy_policy". 
- Create a migration to apply the change to the database
- Update the organisation API to return the new field
- Update the footer in App.js to set the link to the privacy policy using the link in the organisation model.

### LINKED ISSUE #537: Change EEML application form character limits to word limits
This fits better with the Baobab model and is more intuitive for users.

avishkar58:
Resolves #537 
Resolves #531 
Resolves #528 
Resolves #527

### DIFF
--- a/api/migrations/versions/74f1f1476873_update_eeml_form.py
+++ b/api/migrations/versions/74f1f1476873_update_eeml_form.py
@@ -1,0 +1,34 @@
+"""empty message
+
+Revision ID: 74f1f1476873
+Revises: 87213d612eaf
+Create Date: 2020-02-06 17:21:37.083255
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '74f1f1476873'
+down_revision = '87213d612eaf'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.execute(r"""UPDATE question SET validation_regex='^\W*(\w+(\W+|$)){150,500}$', validation_text='You must enter between 150 and 500 words' WHERE headline='Research Interests' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET validation_regex='^\W*(\w+(\W+|$)){0,150}$', validation_text='You must enter fewer than 150 words' WHERE headline='Motivation' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET description='Upload your extended abstract (PDF, max size 10MB)' where headline='Extended abstract' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET description='Upload your CV (PDF, max size 10MB)' where headline='CV' and application_form_id=2""")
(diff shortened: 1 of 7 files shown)


api/migrations/versions/74f1f1476873_update_eeml_form.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def upgrade():
+    op.execute(r"""UPDATE question SET validation_regex='^\W*(\w+(\W+|$)){150,500}$', validation_text='You must enter between 150 and 500 words' WHERE headline='Research Interests' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET validation_regex='^\W*(\w+(\W+|$)){0,150}$', validation_text='You must enter fewer than 150 words' WHERE headline='Motivation' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET description='Upload your extended abstract (PDF, max size 10MB)' where headline='Extended abstract' and application_form_id=2""")
+    op.execute(r"""UPDATE question SET description='Upload your CV (PDF, max size 10MB)' where headline='CV' and application_form_id=2""")
+    op.execute(r"""UPDATE organisation SET system_url='https://test-baobab.deeplearningindaba.com' where id=1""")
+    op.execute(r"""UPDATE organisation SET system_url='https://test-portal.eeml.eu' where id=2""")
+    op.execute(r"""UPDATE organisation SET system_url='http://localhost' where id=3""")
```
amritpurshotam: I'm assuming these URLs are temporary? When we do have to go to prod will we these need to be changed?

avishkar58:
> I don't see the fix for issue #527 though

Oh ya, our emails aren't html anyway, so they were never explicitly hyperlinks, but before it had an http:// prefix which was making Gmail (at least) convert it to a hyperlink. So all I changed was adding the http to the system url in organisation table :D

api/migrations/versions/74f1f1476873_update_eeml_form.py
avishkar58: Yeah will update them directly in the prod DB, the other alternatives like having a system url table with a mode column and reading the mode from the app config file seemed like overkill

amritpurshotam: Cool makes sense

PR #540, opened 2020-02-06
Feature/policy agreement

### LINKED ISSUE #488: Add privacy policy agreed date to AppUser
The follow changes are required for GDPR compliance.

Add a new field called policy_agreed_date to the AppUser model (api/app/users/models.py).

Create a migration to apply the above change to the database 

Update the User API (api/app/users/api.py) post methods to set the policy_agreed_date to the current date when creating a new user 

Create a new PrivacyPolicyAPI with a put method that updates the policy_agreed_date to the current date for the logged in user.

amritpurshotam:
Resolves #488


api/app/users/models.py
```diff
@@ -71,6 +72,12 @@ def deactivate(self):
     def verify(self):
         self.verified_email = True
 
+    def agree_to_policy(self):
+        self.policy_agreed_datetime = datetime.now()
+    
+    def has_agreed(self):
+        return not self.policy_agreed_datetime
```
amritpurshotam: This should actually be `return self.policy_agreed_datetime not None`

amritpurshotam:
@avishkar58 Tests are definitely all passing locally on my side now. But looks like it's that EOF error again on Circle CI

avishkar58:
@amritpurshotam - kicked it off again, but now it's failing with the good old multiple heads issue :)

amritpurshotam:
@avishkar58 Thanks. Did the merge now but EOF error again :/

avishkar58:
@amritpurshotam Ok kicked it off again, hopefully it works this time! Do you not have permission to rerun stuff on CircleCI?

amritpurshotam:
@avishkar58 Looks like it got passed the EOF error but seems now my column isn't being added for some reason 🙈 Will have to look into that later.

And I just created an account now on Circle CI and seems like I can re-run the build now. Not sure if that's all I needed to do 😅

amritpurshotam:
@avishkar58 And sorry I just saw your comment now about the front end needing to be updated. I'm trying to make the change but I'm getting the following error when I navigate to the home page now

```
./src/App.js
Module not found: Can't resolve 'react-cookie-consent' in '/webapp/src'
```

image

amritpurshotam:
@avishkar58 So I tried to update the front end to send the privacy policy agreed field and failed miserably 🙈

avishkar58:
@amritpurshotam what was the issue? Happy to help, would be great to get you more involved in front-end stuff ;)

amritpurshotam:
I think I just don't know enough React 🙈 So I saw the checkbox is there already in `webapp\src\pages\createAccount\components\CreateAccountForm.js` with a field called `agreePrivacyPolicy`. I also see there's the `create` method in `webapp\src\services\user\user.service.js`. I'm not sure now how to get the `agreePrivacyField` field across to the `create` method.

avishkar58:
So there is a user object in the component [state](https://github.com/deep-learning-indaba/Baobab/blob/7c0ed0568e404382891c0ca81fd453189e9a9f44/webapp/src/pages/createAccount/components/CreateAccountForm.js#L30) which gets passed to the create method of the service [here](https://github.com/deep-learning-indaba/Baobab/blob/7c0ed0568e404382891c0ca81fd453189e9a9f44/webapp/src/pages/createAccount/components/CreateAccountForm.js#L138). Best thing to do would be to move the [`agreePrivacyPolicy`](https://github.com/deep-learning-indaba/Baobab/blob/7c0ed0568e404382891c0ca81fd453189e9a9f44/webapp/src/pages/createAccount/components/CreateAccountForm.js#L43) into the user object, then just make sure to update all the usages. Let me know if you're still stuck :)

amritpurshotam:
@avishkar58 So I moved `agreePrivacyPolicy` into user but whenever I try to update anything else relating to `agreePrivacyPolicy` in the file I get syntax errors or the checkbox breaks in that, depending on which values I fiddle with, it won't get ticked or it will always stay ticked.. As it is now, nothing is breaking but the value that's posted is always `false`.

avishkar58:
Cool, so that's because you still need to update the usages of `agreePrivacyPolicy`. For example, in [this function](https://github.com/deep-learning-indaba/Baobab/blob/4547a01ba2343acd8de2a47a40ee3feb80342e5b/webapp/src/pages/createAccount/components/CreateAccountForm.js#L116) it needs to use `this.state.user.agreePrivacyPolicy`. The state update on the next line is a little tricky (remember everything in React is immutable), so you will need to do something like (actually exactly) this: 

`this.setState(currentState => { user: { ...currentState.user, agreePrivacyPolicy: !currentPrivacyPolicy}});`

(the ... is the Javascript *spread* operator, it works similar to the python ** operator, but conveniently in Javascript you can treat an object like a dictionary)

Next, inside the render method, where it "unpacks" the state object, you'll need to move the `agreePrivacyPolicy` from [here](https://github.com/deep-learning-indaba/Baobab/blob/4547a01ba2343acd8de2a47a40ee3feb80342e5b/webapp/src/pages/createAccount/components/CreateAccountForm.js#L192) to [here](https://github.com/deep-learning-indaba/Baobab/blob/4547a01ba2343acd8de2a47a40ee3feb80342e5b/webapp/src/pages/createAccount/components/CreateAccountForm.js#L182)

Looking forward to seeing the next "@amritpurshotam PR" for a front-end issue now ;)

amritpurshotam:
@avishkar58 Awesome thanks! I had to make a minor adjustment to the line you gave me but it seems to be working now.

I've also rerun the migrations now after merging develop so hopefully the tests should pass. 
And phew, 27 commits later, I've added a new column to the db 🤣

amritpurshotam:
@avishkar58 Agreed. Well I think some of the issues that popped up were legitimate. Since we changed the logic of the user creation it did have a knock on effects like breaking the invited guests which uses the user creation. Then I was also doing a bit of fixing migrations and merge migrations a few times due to all the rapid changes being made to the app. So I think the number of commits here are a lot higher than should be normal but yeah definitely something to keep an eye on

Issue #522, opened 2020-02-03
Word count is incorrect


avishkar58:
Enter "Hello world Hello world Hello world Hello world Hello world Hello world " in a box in the application form and the word count shows 2.

avishkar58:
Update: seems like the problem is that the word-count only updates when you press "Space" in the box.

arunavkonwar:
@avishkar58 For some reason, I managed to mess up my local database, is there any way I can reset the database and then rerun all the migrations? Any help with this would be appreciated :)

avishkar58:
Hey @arunavkonwar - you should be able to delete the docker image and then it will be recreated next time you run `docker-compose build`,  see [here](https://www.digitalocean.com/community/tutorials/how-to-remove-docker-images-containers-and-volumes) for some more information

arunavkonwar:
@avishkar58 thanks for the pointer. I got it working again. I just realized that things have changed a bit since I last contributed, now I cannot access any registration/application form and being shown a message "No event exists with that KEY". How do I get any of these pages in the development mode?

avishkar58:
Hey Arunav,

This is a result of the recent organisation white labelling and multi event
changes. The problem is that there are no events attached to the
organisation that gets resolved for "localhost". I'll add a dummy event and
application form to make it easier for contributors, but for now the
easiest thing to do would be to update the organisation table in your local
DB and set the domain on EEML (organisation id 2) to "localhost" and then
update the one that's currently localhost (id 3) to something else. You'll
then see the EEML event and application form.

We've also hidden the menu items for the steps that are not open yet for a
given event (thanks to your date changes :) )

Hope this makes sense

arunavkonwar:
@avishkar58 I have previously never tried accessing a DB running in a container, is there any particular tool you use to manipulate the DB data?

amritpurshotam:
@arunavkonwar I use [pgAdmin ](https://www.pgadmin.org/download/). I think I set the server to localhost and it should pick up the db running in the container.

arunavkonwar:
@amritpurshotam is there a DB username/password? I installed pgAdmin but it requires credentials to login to the DB, where can I find it?

amritpurshotam:
@arunavkonwar So the username, password, and database are all `docker` 😅

PR #544, opened 2020-02-07
fixed word count bug issue #522

arunavkonwar:


arunavkonwar:
@avishkar58 not sure why the build is failing. I only updated the frontend.

avishkar58:
Hey @arunavkonwar - if you merge develop in now it should fix the build problem

webapp/src/components/form/FormTextArea.js
```diff
@@ -59,6 +59,7 @@ class FormTextArea extends React.Component {
             value={this.props.value || ""}
             onChange={this.props.onChange}
             onKeyPress={this.getWordCount}
```
avishkar58: Can you remove the onKeyPress handler please, this event was deprecated: see https://developer.mozilla.org/en-US/docs/Web/API/Document/keypress_event

arunavkonwar:
@avishkar58 still the same issue with the build even after merge

Issue #442, opened 2020-01-14
File API Improvement


avishkar58:
The current file API relies on guids to identify files in Blob storage, and when users retrieve files using the file API, they get a file with a GUID name. This is confusing for users and particularly problematic on windows where the file extension is important. 

Update this by introducing a File table that stores metadata about a file, including:
- id
- filename
- mime_type
- guid 

We need to maintain backward compatibility, as some clients have links to CVs using the GUID-based API. In order to do this, we will leave the GET method on the current API intact and remove the POST method. 

Introduce a new API (can be in the same file, api/app/files/api.py, but with a different route) that implements a POST and GET. The POST should save the file in Blob storage and add a record to the File table to store the metadata and return a file ID that identifies the file in the table. The GET method should take a fileID, query the File table, retrieve the document from blob storage, rename the file and set the appropriate mime_type before returning to the client.

amritpurshotam:
Hi @BrainTheos. Let me know if you have any questions or need assistance with anything. Always happy to help :) 
And if you have some code in progress already and stuck, feel free to push it and open a PR anyway and then we can point you in the right direction.

BrainTheos:
Hi @amritpurshotam and sorry for the delay. Here is what i tried to do : i created a File table in a new models.py at api/app/files/ which i import in api/app/files/api.py. In this api.py, i created a new API class with two methods : get and post as specified in the issue description. Now i'm a bit stacked on how to get file metadata argument to feed in our File() object (table) in order to save it !  Any idea ?

amritpurshotam:
@BrainTheos I've updated the PR you submitted with my comments

BrainTheos:
Well received with thanks.
I'm really happy to learn from your comments, and I highly appreciate your
help !
Hope I'll be done by wednesday !
Best regards!

Le lun. 10 févr. 2020 à 22:07, amritpurshotam <notifications@github.com> a
écrit :

avishkar58:
This can be done in an easier way, by adding an optional `filename` parameter to the existing file API. The API then uses this name to rename the file before returning it to the user. The answer value for files can then be made into a JSON object that has the GUID and filename (similar to how the multi-file type works) and the front-end extracts this name and passes it to the backend.

PR #549, opened 2020-02-07
File api improvement 442

BrainTheos:
Tried to solve the issue #442 according to the specs given in the description. But still stucked on how to actually get the metadata file to be saved in the File table !


api/app/files/models.py
```diff
@@ -0,0 +1,9 @@
+from app import db
+
+class File(db.Model):
+	__tablename__ = "file"
+	id = db.Column(db.Integer, primary_key=True)
+	file_name = db.Column(db.String(50), unique=True, nullable=False)
+	mime_type = db.Column(db.String(20), nullable=False)
+	guid = db.Column(db.String(50), unique=True, nullable=False)
```
amritpurshotam: I think this model is fine. Perhaps add a constructor though. It just results in cleaner code when want to add a new row in the database and allows us to have a consistent way of creating a file object.

```
def __init__(self, file_name, mime_type):
  self.file_name = file_name
  self.mime_type = mime_type
  self.guid = str(uuid.uuid4().hex)
```

Then have a look at the [Database Migrations](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/README.md) section of the docs. Basically these classes (that inherit from `db.Model`) represent a table in the database so we actually need to make sure a table exists in the database. The migration will create the script that will actually go and create this table then. Let me know if the docs aren't clear. They're a work in progress so it definitely helps to know if something doesn't make sense.

api/app/files/api.py
```diff
@@ -45,9 +46,33 @@ def get(self):
             return send_file(temp.name, as_attachment=True, attachment_filename=args['filename'], mimetype='application/pdf')
 
 
+
+class FileAPIImproved(FileUploadMixin, restful.Resource):
+    def get(self, file_id):
```
amritpurshotam: Remove the `file_id` argument. We use the request parser to extract the arguments from the HTTP request query.

api/app/files/api.py
```diff
@@ -45,9 +46,33 @@ def get(self):
             return send_file(temp.name, as_attachment=True, attachment_filename=args['filename'], mimetype='application/pdf')
 
 
+
+class FileAPIImproved(FileUploadMixin, restful.Resource):
+    def get(self, file_id):
+        # TO DO: Query the File table
+
+        req_parser = reqparse.RequestParser()
+        req_parser.add_argument('filename', type=str, required=True)
```
amritpurshotam: So over here this should be `req_parser.add_argument('file_id', type=int, required=True)`

api/app/files/api.py
```diff
@@ -45,9 +46,33 @@ def get(self):
...
+
+class FileAPIImproved(FileUploadMixin, restful.Resource):
+    def get(self, file_id):
+        # TO DO: Query the File table
+
+        req_parser = reqparse.RequestParser()
+        req_parser.add_argument('filename', type=str, required=True)
+        args = req_parser.parse_args()
+        LOGGER.debug("FileUpload GET args: {}".format(args))
+
```
amritpurshotam: Then to retrieve the file object, look at some of the other repository files in the project and implement a `get_by_id `method. And then use that method to retrieve the relevant file object. There's plenty of examples in the project to see how to do this. If you're still struggling let me know.

api/app/files/api.py
```diff
@@ -45,9 +46,33 @@ def get(self):
...
+        bucket = _get_storage_bucket()
+
+        blob = bucket.blob(args['filename'])
+
+        # TO DO : Rename the file and set the appropriate mime_type
+
+        with tempfile.NamedTemporaryFile() as temp:
+            blob.download_to_filename(temp.name)
+            return send_file(temp.name, as_attachment=True, attachment_filename=args['filename'], mimetype='application/pdf')
```
amritpurshotam: Then instead of `args`, we will use the `file` object we just retrieved from the database to set the `attachment_filename` and `mimetype`.

api/app/files/api.py
```diff
@@ -45,9 +46,33 @@ def get(self):
...
+
+
     @auth_required
     def post(self):
         args = self.req_parser.parse_args()
+        file_name = args['filename']
+        mime_ = args['event_id']
+        email_template = args['email_template']
```
amritpurshotam: So I think look at the existing `FileUploadAPI` `post` method and do something similar. You can probably then get the file name by doing `file_name = file.filename` and them mime_type by doing `mime_type = file.content_type` (this is what's happening currently). Then create your file object with `new_file = File(file_name, mime_type)`. Then again make a new repository method to save this file object. I'll leave this to you to see how we do this else where in the project (if you're still struggling let me know)
Email template just remove.

api/app/files/api.py
```diff
@@ -67,6 +92,9 @@ def post(self):
 
         blob.upload_from_string(bytes_file, content_type=content_type)
 
+        # TO DO : Add a record to File table to store metadata
+
         return {
-            'file_id': unique_name,
+            'file_id': unique_name, # TO DO : Return file id
```
amritpurshotam: Maybe return the file_name, mime_type, and the guid

api/app/files/api.py
amritpurshotam: The rest of the saving to blob storage should be the same as what's in the existing post method

api/app/routes.py
```diff
@@ -29,6 +29,7 @@
 rest_api.add_resource(content_api.CountryContentAPI,
                       '/api/v1/content/countries')
 rest_api.add_resource(files_api.FileUploadAPI, '/api/v1/file')
+rest_api.add_resource(files_api.FileAPIImproved, '/api/v1/file')
```
amritpurshotam: URIs (Uniform Resource Identifiers) need to be different. So set this to `/api/v1/fileimproved` or something

amritpurshotam:
@BrainTheos Just checking in. How's the updates going? Feel free to ping me if you have any questions :)

BrainTheos:
Hi @amritpurshotam !
I Think i'm done with the suggested updates. Hope there is no longer git
mistakes ? :) !
 Is another PR as suggested so far still necessary ?.
Thanks a bunch for your help, i really appreciate !


Le ven. 14 févr. 2020 à 13:25, amritpurshotam <notifications@github.com> a
écrit :

api/app/routes.py
BrainTheos: Hi @amritpurshotam ! I Think i'm done with the suggested updates. Hope there is no longer git mistakes ? :) ! Is another PR as suggested so far still necessary ?. Thanks a bunch for your help, i really appreciate !

amritpurshotam:
@BrainTheos Awesome thanks! No git issues anymore it seems. Will go through the PR now

api/app/files/models.py
```diff
@@ -0,0 +1,15 @@
+import uuid
+from app import db
+
+class File(db.Model):
+	__tablename__ = "file"
+	id = db.Column(db.Integer, primary_key=True)
+	file_name = db.Column(db.String(50), unique=True, nullable=False)
+	mime_type = db.Column(db.String(20), nullable=False)
+	guid = db.Column(db.String(50), unique=True, nullable=False)
```
amritpurshotam: I think drop the unique constraint here as well. It's highly unlikely that two guids will ever [collide](https://stackoverflow.com/a/39776).

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
...
+        req_parser.add_argument('file_id', type=int, required=True)
+        args = req_parser.parse_args()
+        LOGGER.debug("FileUpload GET args: {}".format(args))
+
+        bucket = _get_storage_bucket()
+        file = get_by_id(args['file_id'])
+
+        blob = bucket.blob(args['file_id'])
+
+        # TO DO : Rename the file and set the appropriate mime_type
```
amritpurshotam: Can we remove this comment please?

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
...
+        last_save_file = db.session.query(File).last()
+        return last_save_file.file_id
+
+
     @auth_required
     def post(self):
         args = self.req_parser.parse_args()
+        file_name = args['filename']
+        mime_ = args['event_id']
+        #email_template = args['email_template']
```
amritpurshotam: Can we remove all three lines here? They're not used anywhere

api/migrations/versions/54aba10497f4_.py
```diff
@@ -0,0 +1,36 @@
...
+    op.create_table('file',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('file_name', sa.String(length=50), nullable=False),
+    sa.Column('mime_type', sa.String(length=20), nullable=False),
+    sa.Column('guid', sa.String(length=50), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('file_name'),
+    sa.UniqueConstraint('guid')
+    )
+    op.drop_column(u'application_form', 'deadline')
```
amritpurshotam: So when creating a new migration something to be mindful of is to watch for fields appearing that have nothing to do with the changes you've made. For example this one on `application_form`. This field was recently dropped in another PR. I think what happened was that you merged in the latest changes but you never upgraded your local version of your database resulting in it appearing to alembic that the database and the application_form model were out of sync. Normally alembic shouldn't let you make a migration if your local database isn't up to date but seemed like that didn't happen for some reason. So I think for now just back space this line.

api/migrations/versions/54aba10497f4_.py
```diff
@@ -0,0 +1,36 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(u'application_form', sa.Column('deadline', postgresql.TIMESTAMP(), autoincrement=False, nullable=False))
```
amritpurshotam: And backspace this here as well

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
             return send_file(temp.name, as_attachment=True, attachment_filename=args['filename'], mimetype='application/pdf')
 
 
+
+class FileAPIImproved(FileUploadMixin, restful.Resource):
+
+
+    @staticmethod
+    def get_by_id(file_id):
+        return db.session.query(File).get(file_id)
```
amritpurshotam: Can we move this method into a separate repository file in `/api/app/files/repository.py` and into a class called `FileRepository`? Look at the rest of the project for examples. Then once it's in there, import the class in with `from app.files.repository import FileRepository as file_repository`. Then you will be able to use the method with `file_repository.get_by_id()`. The reason for this is that we're trying to separate the database operations from what goes on in the API file which results in cleaner code.

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
...
+
+class FileAPIImproved(FileUploadMixin, restful.Resource):
+
+
+    @staticmethod
+    def get_by_id(file_id):
+        return db.session.query(File).get(file_id)
+
+    def get(self):
+        # TO DO: Query the File table
```
amritpurshotam: Can we remove this comment?

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
...
+
+        req_parser = reqparse.RequestParser()
+        req_parser.add_argument('file_id', type=int, required=True)
+        args = req_parser.parse_args()
+        LOGGER.debug("FileUpload GET args: {}".format(args))
```
amritpurshotam: Can we remove this logging please?

api/app/files/api.py
```diff
@@ -45,9 +46,48 @@ def get(self):
...
+            blob.download_to_filename(temp.name)
+            return send_file(temp.name, as_attachment=True, attachment_filename=file.file_name, mimetype=file.mime_type)
+
+    
+    @staticmethod
+    def save_file_object(file_obj):
+        db.session.add(file_obj)
+        db.session.commit()
+        last_save_file = db.session.query(File).last()
+        return last_save_file.file_id
```
amritpurshotam: As described elsewhere, move this into a repository file and class.

Also the way you're retrieving the file_id is risky. What if two people upload files at almost exactly the same time? It's possible the first person will get back the second persons `file_id` which is incorrect.

Instead I think you might be able to just say return file.file_id. SQL Alchemy should hydrate the `file_id` automatically for you (but I think just test and make sure). Also rename `file_obj` as `file` and `save_file_obj` as `save`. We know it's an object :)

BrainTheos: Hey @amritpurshotam. Finish with new updates. But it seems there is 1 issue. Would you please tell me where it comes from ? Hope it's not a git issue again ? :) !

api/app/files/repository.py
```diff
@@ -0,0 +1,14 @@
+from app import db
+from app.files.models import File 
+
+class FileRepository:
+	@staticmethod
```
amritpurshotam: Looks like some indentation is causing errors. Please correct.

BrainTheos: Hi @amritpurshotam. Get a bit confused with migration issues. I tried to do as you suggested by deleting the current migration file (Right ?) but it still doesn't work.

amritpurshotam: Sorry forgot to mention you should delete your current file table on your local instance of your db. Use [pgAdmin](https://www.pgadmin.org/download/). Connect to `localhost` and use `docker` as both the username and password. Then you'll have a gui to delete the table.

BrainTheos: Hi @amritpurshotam. I've installed pgAdmin but I don't see exactly how to connect to my baoba local instance db. Sould I run docker also the same time ?

amritpurshotam: @BrainTheos Yes `docker-compose up` will spin up the database instance and you should be able to connect

amritpurshotam: @BrainTheos Something I also do to reset my database environment is to run the following query

```
DROP SCHEMA public CASCADE;
CREATE SCHEMA public;
```
Then run the upgrade command to rerun all the existing migrations. Then you can be sure your environment is up to date and synced correctly.

BrainTheos: @amritpurshotam , please look at this error i get when i run docker-compose up after lauching pgAdmin 
ERROR: for db  Cannot start service db: driver failed programming external connectivity on endpoint baobab_db_1 (387b839a1bba7b7a1dcace2243594b756c3206e1f2c85c2fc79755e0a15cb628): Error starting userland proxy: listen tcp 0.0.0.0:5432: bind: address already in use
ERROR: Encountered errors while bringing up the project.

KaleabTessera: @BrainTheos this means you have another instance of the db container running. 
Try:
-  `docker-compose stop` or
 - `docker container ls` and then `docker stop [db_container_id]`

BrainTheos: @KaleabTessera thanks. it looks working. but still unable to see local db instance in pgAdmin (in Servers top left panel). and here is an other error i get though i tried to update my branch : 

Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`
webapp_1   | Failed to compile.
webapp_1   | 
webapp_1   | ./src/App.js
webapp_1   | Module not found: Can't resolve 'react-cookie-consent' in '/webapp/src'

amritpurshotam: I think you may need to rebuild your docker image with `docker-compose build`

I remember I had that a little while back when the PR to add the cookie consent feature was merged in

amritpurshotam: @BrainTheos Did yo manage to resolve the issue by rebuilding your image?

BrainTheos: @amritpurshotam , I didn't manage yet though i've rebuild my image. The problem is that my pgAdmin seems to not detect any database.

amritpurshotam: @BrainTheos Let's discuss on slack on how to get you setup. For now to get things moving along, I've done the migration. This issue is now complete

BrainTheos: Ok @amritpurshotam, Thank you so much for your help and patience. I've really learned a lot from you !

avishkar58:
We ended going a different route using the same File API.

PR #551, opened 2020-02-09
increased the readiness check for app_start_timeout_sec

jaderabbit:
@avishkar58 the parameter is in the app.yaml which is encrypted. I've doubled it

That said, when that fails, it's sometimes a sign of something else going wrong so checking the cloud logs


avishkar58:
Thanks @jaderabbit! :)

Issue #474, opened 2020-01-25
Add event configuration front-end


avishkar58:
Depends on #473 

Create functions in webapp/src/services/events to talk to the event API (api/app/events/api.py) 

Create a front-end page called eventConfig that allows the properties of the current event to be viewed and updated. 

Add a menu item linking to this new page under the "Event Admin" menu.

s-mawjee:
I'll work on this.

Is there a UI layout?

avishkar58:
Hey, there isn't one, if you're taking this issue, would you mind doing a
quick mockup? Thanks!

s-mawjee:
The dropdown will contain the available events (that user has admin role). The user can `select an event from the dropdown` or `add event`. 

If the user selects an event, the `add event` button will be disabled. 
The update button will only be active if there was a change made.

Still need to add the other date-time fields. Will use a date-time-picker, for all the date-time fields. 

<img width="1664" alt="Screenshot 2020-01-30 at 07 55 54" src="https://user-images.githubusercontent.com/21179950/73423849-f884c580-4335-11ea-9689-cbe859c61ad0.png">

s-mawjee:
When creating an event from the UI, is the user limited to a specific organisation? i.e `organisation_id` for event is defined by `g.organisation`

avishkar58:
Hey Shahil,

Yes that's correct :)

PR #564, opened 2020-02-10
#474 Event Config UI

s-mawjee:


s-mawjee:
@avishkar58, please rerun the CI build

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -0,0 +1,418 @@
...
+    });
+  };
+
+  updateEventDetails = (fieldName, e) => {
+    let u = { ...this.state.updatedEvent, [fieldName]: e.target.value };
+
+    this.setState({
+      updatedEvent: u
+    });
+    this.hasBeenEdited();
```
avishkar58: You probably want to call this in the callback of setState, because at the moment there is no guarantee that the state would have updated before hasBeenEdited runs.

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -0,0 +1,418 @@
...
+  };
+
+  updateDateTimeEventDetails = (fieldName, value) => {
+    let u = { ...this.state.updatedEvent, [fieldName]: value };
+
+    this.setState({
+      updatedEvent: u
+    });
+    this.hasBeenEdited();
```
avishkar58: Same here

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -0,0 +1,418 @@
...
+  onClickSubmit = () => {
+    // PUT
+    eventService.update(this.state.updatedEvent).then(result => {
+      this.setState({
+        preEvent: result.event,
+        updatedEvent: result.event,
+        hasBeenUpdated: false,
+        successButtonText: "Update Event",
+        updatingEvent: true,
+        addingNewEvent: false,
```
avishkar58: These 3 properties don't seem to be used?

webapp/src/services/events/events.service.js
```diff
@@ -1,53 +1,156 @@
...
+      // handle success
+      let event = null;
+      if (response) {
+        event = response.data;
+      }
+      return { event: event, error: "" };
+    })
+    .catch(error => {
+      // handle error
+      return { event: null, error: error };
```
avishkar58: I've found before that this way of setting the error doesn't actually return the error messages sent by our backend. See the pattern used in getEvents. Maybe something has changed since then though?

webapp/src/pages/eventConfig/components/EventConfigComponent.js
s-mawjee: Thanks, I didn't know about this.

webapp/src/services/events/events.service.js
s-mawjee: Is there a design pattern? I have updated the return.

Issue #570, opened 2020-02-11
Small devices screen scaling & alert component fix


CilliersN:
Components don't scale correctly on smaller devices.

Alert message box doesn't have a top margin -only bottom. This is a bootstrap component problem and as such, top margins need to be applied wherever these alerts are used.

<img width="702" alt="Screenshot 2020-02-12 at 10 35 08" src="https://user-images.githubusercontent.com/29726916/74317136-7eb3f980-4d83-11ea-8732-84a4d25e3bb3.png">
<img width="424" alt="Screenshot 2020-02-11 at 11 18 23" src="https://user-images.githubusercontent.com/29726916/74317144-81165380-4d83-11ea-9359-8a850ab5eef6.png">

<img width="367" alt="Screenshot 2020-02-11 at 11 22 19" src="https://user-images.githubusercontent.com/29726916/74238335-43132400-4cde-11ea-9501-1e11e1e1c94b.png">

Issue #567, opened 2020-02-11
Add a test event, application form etc to the localhost organisation


avishkar58:
At the moment, when running locally, developers won't see much in the application because the organisation that gets resolves for localhost has no event associated with it. Create a test event and application form (with the close dates of the event stages set far into the future) that people can use for testing.

PR #574, opened 2020-02-12
Scaling alert component fixes

CilliersN:
as per issue #570 
Components don't scale correctly on smaller devices.
Alert message box doesn't have a top margin -only bottom. This is a bootstrap component problem and as such, top margins need to be applied wherever these alerts are used.

Result:
Screenshot 2020-02-12 at 10 26 14

Errors were fixed throughout the repo. Indentation Cleanup was also done on files that were worked on

### DIFF
--- a/webapp/src/pages/createAccount/CreateAccount.css
+++ b/webapp/src/pages/createAccount/CreateAccount.css
@@ -3,7 +3,3 @@
     padding: 60px 0;
   }
 }
-
-.alert-container{
-  margin-top: 16px;
-}
--- a/webapp/src/pages/createAccount/CreateAccount.css
+++ b/webapp/src/pages/createAccount/CreateAccount.css
@@ -3,3 +3,7 @@
     padding: 60px 0;
   }
 }
+
+.alert-container{
+  margin-top: 16px;
+}
--- a/webapp/src/App.css
+++ b/webapp/src/App.css
@@ -69,3 +69,7 @@ footer .btn {
 .navbar-brand .brand-image {
   margin-right: 10px;
 }
+
+.alert-container{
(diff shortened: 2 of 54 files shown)


CilliersN:
Please review but do not merge yet, I just want to confirm marginal changes on alerts

webapp/src/pages/createAccount/CreateAccount.css
```diff
@@ -2,4 +2,8 @@
   .CreateAccount {
     padding: 60px 0;
   }
+}
+
+.alert-container{
```
avishkar58: I think you can get away with defining this style once in App.css, which is available to every page anyway.

avishkar58:
By the way @CilliersN, until #567 lands, if you change the domain and system_url of EEML in the organisation table on your local DB to "localhost", you can see the event home page, application form etc - I'm sure there are a lot more styling issues there!

webapp/src/pages/createAccount/CreateAccount.css
CilliersN: ahhh yes, thats right. 
Sorry dont know how ive missed that :)
still finding my way around your repo structure

CilliersN:
> By the way @CilliersN, until #567 lands, if you change the domain and system_url of EEML in the organisation table on your local DB to "localhost", you can see the event home page, application form etc - I'm sure there are a lot more styling issues there!

Thanx for this... will check it out.

webapp/src/pages/createAccount/CreateAccount.css
CilliersN: I've implemented the new changes.

PR #575, opened 2020-02-12
Code Refactor, remove deadline from Application form Model

### LINKED ISSUE #518: Remove deadline from application_form
Since events now have an application_close date, we should remove the deadline date from the ApplicationForm model and replace all of its usages with the event's application_close date.

jainesh99:
Hi All,

This pull request resolves #518 .

I have removed the deadline field from the application form model and updated the code to use application close from event.

### DIFF
--- a/api/migrations/versions/d7f3adff2c9f_.py
+++ b/api/migrations/versions/d7f3adff2c9f_.py
@@ -1,0 +1,22 @@
+"""empty message
+
+Revision ID: d7f3adff2c9f
+Revises: ('3e7bdff8af80', 'b6a2a69fbb13')
+Create Date: 2020-02-12 11:12:35.018060
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'd7f3adff2c9f'
+down_revision = ('3e7bdff8af80', 'b6a2a69fbb13')
+
+from alembic import op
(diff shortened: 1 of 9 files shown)


api/migrations/versions/d7f3adff2c9f_.py
```diff
@@ -0,0 +1,22 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    pass
```
amritpurshotam: @jainesh99 Seems you did a merge migration but we're still missing the dropping of the column itself?

jainesh99: @amritpurshotam , thank you spotting this out for me. I have made the changes, please have a look. 

I have done the automatic migration script, but I had to tweak it a bit because the build was failing on the run migration step due to the "deadline" column containing null values. I added a server default and then altered that column

amritpurshotam: @jainesh99 Did you push the code? I can't see it? Unless I wonder if it's cos the PR was closed

jainesh99: Just created a PR for it now.

PR #576, opened 2020-02-12
Amended the versions, to drop the deadline columns

jainesh99:
I used the automatic script generation to create the versions file accordingly and had to modify it to add a server default value, then alter the column to take it away.

This was due to the build failing with this error:

sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) column "deadline" contains null values


jainesh99:
Resolves #518

Issue #558, opened 2020-02-10
Add nominations flag to Application Form


avishkar58:
Add a new binary flag to the [ApplicationForm model](https://github.com/deep-learning-indaba/Baobab/blob/f65917877a0675a300af41cef0c5df269c408d40/api/app/applicationModel/models.py#L4) called `nominations` which should be required. 

Generate a migration to apply the change to the database, set the value to False on all existing application forms.

Update the `form_fields` dictionary in the [API]() file to include the new `nominations` field. This will ensure that it gets returned to clients with the get request

Update the tests (in particular the `test_get_form` function) to check that the nominations flag is returned correctly. While doing this, please change all the existing Python asserts to `self.assertEqual(actual, expected)` (this is better than using Python asserts because it gives more sensible error messages when something fails)

**Definition of Done**
- Submit a PR to merge the changes into develop
- All tests pass
- Migration has been created and can be successfully upgraded and downgraded both locally and as part of the PR tests
- Any PR review comments have been addressed

PR #577, opened 2020-02-12
Solves Issue #558

arunavkonwar:


arunavkonwar:
@amritpurshotam the build fails for some reason, not sure what's the issue. I had to edit the `api/migrations/versions/f8ab033698a2_.py` file to set a default value to the column `nominations`. Works fine in local though!

api/app/utils/testing.py
```diff
@@ -120,4 +121,14 @@ def add_to_db(self, obj):
     def tearDown(self):
         db.session.remove()
         db.reflect()
-        db.drop_all()
\ No newline at end of file
+        db.drop_all()
+
+    def create_application_form(self,
+                            is_open,
```
avishkar58: Please set defaults

api/app/utils/testing.py
```diff
@@ -120,4 +121,14 @@ def add_to_db(self, obj):
...
+                            deadline,
+                            nominations):
+                            
+        application_form = ApplicationForm(is_open, deadline, nominations)
```
avishkar58: This doesn't look right?

api/app/applicationModel/models.py
```diff
@@ -9,6 +9,7 @@ class ApplicationForm(db.Model):
     is_open = db.Column(db.Boolean(), nullable=False)
 
     event = db.relationship('Event', foreign_keys=[event_id])
+    nominations = db.Column(db.Boolean(), nullable=False)
 
     def __init__(self, event_id, is_open):
```
avishkar58: Please add nominations to the constructor

avishkar58:
Hey @arunavkonwar - the tests are failing because there are a lot of other tests besides the applicationModel one that create application forms and aren't passing the new nominations flag to the constructor. So you'll have to search for those places in the various tests and update them to use your new `create_application_form` method - a little tedious, which is why we're trying to add these helper methods now to prevent having to do it next time we add a field :D

arunavkonwar:
@avishkar58 this solved it. Thanks.

api/app/responses/tests.py
```diff
@@ -76,7 +76,7 @@ def _seed_data(self):
             datetime.now(), datetime.now(), datetime.now(), datetime.now())
         _add_object_to_db(self.test_event)
         self.test_form = ApplicationForm(
-            self.test_event.id, True)
+            self.test_event.id, True, False)
```
avishkar58: Please change these to all your new method instead: `self.create_application_form(...)` - that way, next time someone makes a change to ApplicationForm, they don't have to update all these tests again...
(same comment applies to the other tests)

arunavkonwar:
@amritpurshotam @avishkar58 can you take a look at this PR. Wondering if there needs any more changes. Thanks :)

PR #608, opened 2020-02-27
Added local event, application form, review form, registration form

### LINKED ISSUE #567: Add a test event, application form etc to the localhost organisation

avishkar58:
Resolves #567


avishkar58:
If you bulk insert into a table then postgres doesn't automatically
increment the primary key sequence, and when setting up Indaba 2019 config
we mostly did bulk inserts...

Issue #613, opened 2020-02-28
Add ReviewConfiguration model


avishkar58:
Add a new model to api/app/reviews/models.py called `ReviewConfiguration` which has the following fields:
- id (int, primary key, not null)
- review_form_id (int, foreign key references review_form.id, not null)
- num_reviews_required (int, not null)
- num_optional_reviews (int, not null)
- drop_optional_question_id (int, foreign key references review_question.id, nullable)
- drop_optional_agreement_values (string, nullable)

Create a migration to apply the change to the database

PR #617, opened 2020-02-29
Added a model for ReviewConfiguration

lzbotha:
Added a ReviewConfiguration model as per:
https://github.com/deep-learning-indaba/Baobab/issues/613


amritpurshotam:
Looking now

api/app/reviews/models.py
```diff
@@ -100,4 +100,16 @@ def __init__(self,
...
-        self.value = value
\ No newline at end of file
+        self.value = value
+
+
+class ReviewConfiguration(db.Model):
+    id = db.Column(db.Integer(), primary_key=True)
+    review_form_id = db.Column(db.Integer(), db.ForeignKey('review_form.id'), nullable=False)
+    num_reviews_required = db.Column(db.Integer())
+    num_optional_reviews = db.Column(db.Integer())
```
amritpurshotam: I think specify `nullable=False` here as well and then do another migration (having 2 will be fine)

api/app/reviews/models.py
```diff
@@ -100,4 +100,16 @@ def __init__(self,
...
+    drop_optional_question_id = db.Column(db.Integer(), db.ForeignKey('review_question.id'), nullable=False)
+    drop_optional_agreement_values = db.Column(db.String(), nullable=False)
```
avishkar58: These two fields should be nullable

amritpurshotam: Oh you got to it first 😅

api/app/reviews/models.py
lzbotha: I think the default is `nullable=False`. Happy to make this explicit though.

lzbotha: Particularly since I did it on all the other fields 😓

amritpurshotam: Na if you look at the migration, they were set to be nullable.
Unless do you know of a way to configure SQL Alchemy to make fields not nullable by default?

lzbotha: You're completely right.

amritpurshotam:
@lzbotha Cool, first closed PR for the day!

PR #619, opened 2020-02-29
Add date component

### LINKED ISSUE #612: Add Date Component
Add a new form component in the file webapp/src/components/form/FormDate.js 

This component should represent a date picker using the [react-datetime-picker](https://www.npmjs.com/package/react-datetime-picker) module
- When the user selects a new date, call `props.onChange` with the string value of the date selected (In [ISO8601 format](https://www.iso.org/iso-8601-date-and-time-format.html))
- Set the state of the date according to the value of `props.value`, note that this will be a `string` in Iso8601 format, so you may need to convert it to a date.
- Add support for this component to the [ApplicationForm](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/webapp/src/pages/applicationForm/components/ApplicationForm.js#L69), [RegistrationForm](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/webapp/src/pages/registration/components/RegistrationComponent.js#L331) and [ReviewForm](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/webapp/src/pages/review/components/ReviewForm.js#L44) for questions of type "date"

TarCode:
Resolves #612

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -1145,6 +1145,11 @@
         "@xtuc/long": "4.2.1"
       }
     },
+    "@wojtekmaj/date-utils": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@wojtekmaj/date-utils/-/date-utils-1.0.2.tgz",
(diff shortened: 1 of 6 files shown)


TarCode:
Thanks @avishkar58 !
Will do :)

PR #621, opened 2020-02-29
feat(event_type): Added event_type to event_info return

scott-glover020:
This is the back end changes that need to be made in order to get the event type.

### DIFF
--- a/api/app/events/tests.py
+++ b/api/app/events/tests.py
@@ -209,6 +209,16 @@ class EventsAPITest(ApiTestCase):
         assert data[0]["description"] == 'Event Description'
         assert data[0]["status"] == 'Application not available'
 
+    def test_get_event_type(self):
+        self.seed_static_data()
+        db.session.delete(self.test_form)
+        db.session.commit()
+
+        response = self.app.get('/api/v1/events')
+        assert response.status_code == 200
+        data = json.loads(response.data)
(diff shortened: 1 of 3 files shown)


api/app/events/tests.py
```diff
@@ -209,6 +209,16 @@ def test_get_events_unauthed_not_available(self):
         assert data[0]["description"] == 'Event Description'
         assert data[0]["status"] == 'Application not available'
 
+    def test_get_event_type(self):
+        self.seed_static_data()
+        db.session.delete(self.test_form)
+        db.session.commit()
+
+        response = self.app.get('/api/v1/events')
+        assert response.status_code == 200
```
avishkar58: Please use `self.assertEqual(response.status_code, 200)` instead - this provides better error messages when the assertion fails (I know the rest of the file does it with Python assert, but we're slowly trying to change this :) )

avishkar58: Bonus points if you'd like to clean up the rest of the asserts 🥇

scott-glover020: Ye Ill give it a go!

scott-glover020:
Refactored the assert cases to self.assertEqual for the events tests

steveskimmel:
Added a commit for the front end change to support this backend update

PR #623, opened 2020-02-29
added a put to the reference

### LINKED ISSUE #592: Create PUT method on Reference API
Add a PUT method to the [ReferenceAPI](https://github.com/deep-learning-indaba/Baobab/blob/bf99909f71ef7cca4923d3a93ddcea9e6d443ccf/api/app/references/api.py#L110) which updates an existing Reference using a `token` and `uploaded_document` parameter. Return a 404 error if the reference doesn't already exist. 

Update the POST and PUT methods to return an error ([APPLICATIONS_CLOSED](https://github.com/deep-learning-indaba/Baobab/blob/bf99909f71ef7cca4923d3a93ddcea9e6d443ccf/api/app/utils/errors.py#L71)) if the `is_applications_open` flag on the relevant event is `False`

OdendaalG:
Fixes #592

### DIFF
--- a/api/app/references/api.py
+++ b/api/app/references/api.py
@@ -7,7 +7,8 @@ from flask_restful import fields, marshal_with
 from app.events.models import Event
 from app.events.repository import EventRepository as event_repository
 from app.utils.errors import EVENT_NOT_FOUND, USER_NOT_FOUND, RESPONSE_NOT_FOUND, FORBIDDEN,\
-                        REFRERENCE_REQUEST_WITH_TOKEN_NOT_FOUND, DUPLICATE_REFERENCE_SUBMISSION
+                        REFRERENCE_REQUEST_WITH_TOKEN_NOT_FOUND, DUPLICATE_REFERENCE_SUBMISSION,\
+                        APPLICATIONS_CLOSED
 
 from app.utils.auth import auth_optional, auth_required
 from app.utils.emailer import send_mail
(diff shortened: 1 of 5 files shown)


api/app/references/api.py
```diff
@@ -130,12 +131,37 @@ def post(self):
...
+    def put(self):
+        args = self.put_req_parser.parse_args()
+        token = args['token']
+        uploaded_document = args['uploaded_document']
+        reference_request = reference_request_repository.get_by_token(token)
+        event = event_repository.get_event_by_response_id(reference_request.response_id)
+        reference = reference_request_repository.get_reference_by_reference_request_id(reference_request.id)
+
+        if not reference_request:
+            return {}, 404
```
amritpurshotam: Perhaps in the api\app\utils\errors.py file add a new error `REFERENCE_REQUEST_NOT_FOUND = ({'message': 'No response found for the given event and user'}, 404)` and then import that and use it here

api/app/references/tests.py
```diff
@@ -94,3 +95,66 @@ def test_reference_api(self):
...
+
+        response = self.app.get(
+            '/api/v1/reference', data={'response_id':1}, headers=self.headers)
+        LOGGER.debug(response.data)
+        data = json.loads(response.data)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(data), 1)
+        reference_request = reference_request_repository.get_by_id(1)
+        self.assertEqual(reference_request.has_reference(), True)
```
amritpurshotam: Can we perhaps do an assert to see if the `uploaded_document` field actually changed to `DOCT-UPLOAD-79000`?

api/app/references/api.py
OdendaalG: Added new error and used it in the reference API.

OdendaalG:
Hi @amritpurshotam , I implemented the changes that you mentioned, but the issue is still open. I don't know if more changes have been requested, please let me know if there is something more I can do.
Regards

OdendaalG:
No problem on the delay! I thought I had messed something up with the git process, thanks for the reply!

Issue #591, opened 2020-02-23
Add a ReferenceRequestDetails API


avishkar58:
## Description
The page the captures a reference from a referee needs some details about the candidate requesting the reference. 

## Tasks
Create a new ReferenceRequestDetailAPI in [this file](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/references/api.py) which accepts a `token` parameter
- Retrieve the relevant ReferenceRequest, Response and Event objects using the token. Attempt to retrieve a Reference associated with the ReferenceRequest. Return the following information: `candidate_title`, `candidate_firstname`, `candidate_lastname` and `candidate_email` from the Response object (see #559); `name`, `description`, `is_application_open` and `email_from` from the Event model; `timestamp` (name it `reference_submitted_timestamp` in the response) from the Reference model if it exists (`None` if it doesn't exist).
- Return a 404 error if no ReferenceRequest is found with the given token

## Definition of Done
- New API created
- Tests created for new API
- Endpoint created in routes.py
- PR submitted to merge into develop, reviewed and review comments addressed.

avishkar58:
Hey @ChrisCurrin - assigned you here because it sounded like you were working on this from our previous discussion - feel free to unassign yourself if that's not the case :)

Issue #560, opened 2020-02-10
Update response API to cater for nominations


avishkar58:
Depends on #559 

We need to cater for certain events (and application forms) allowing people to **nominate** someone else (eg for Indaba awards). This means that we should allow multiple responses when the application form's `nominations` flag set to True. If the application form has `nominations` set to false, then we assume that the response is for the logged in user. 

In the file https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/responses/api.py

## Update `ResponseAPI` the GET method:
- Return a list of all responses instead of just the first one

## Update the `ResponseAPI` POST method:
- Raise a duplicate error if the `nominations` flag is set to false and a Response already exists for the logged in user and event

Cleanup:
- Replace any existing direct queries (`db.session.query(....)`) with calls to the response repository

amritpurshotam:
@avishkar58 Will there be multiple application forms now per event? The normal application and the nomination application? Will need to double check but I think a lot of the logic in the app assumes one application form per event.

avishkar58:
Nah, there will still only be one application form per event, and it will
either be nominations or no nominations (eg the awards will have one
application form with nominations set to true and people can nominate
themselves)

amritpurshotam:
@avishkar58 Okay I think I see what you're doing now. Is there a requirement for a motivation why? I'd imagine that would could be a Question and whatever they say will be the Answer. Just trying to figure out the full flow of how this will work because at some point all the nominations will have to be grouped somehow and the code will have to work no matter the event, application form, or question

avishkar58:
Yeah that will be a question in the application form :) the design of this
is probably quite suboptimal.. do you have any suggestions for improving
it?

amritpurshotam:
I've been mulling over a few designs (all suboptimal) but I realised there's already something there. Sections! I forgot about that until looking at the models again now. So we can have a Nomination section.
Also I think we need a way to generate the ERD from the database directly. I'll look to see if there's an easy way to do it in Postgres, Would have that solution pop out a little more obviously.

amritpurshotam:
@avishkar58 I just had another thought now. Would we even need the response to store the nomination title, firstname, etc and the nominations flag on the ApplicationForm model? Can that not just be configurable with the Nomination Section and Questions?

avishkar58:
Hmmm yeah your design definitely sounds simpler and cleaner than mine and
solves an issue with the award form that I was struggling to think through
[...]

PR #624, opened 2020-02-29
fixes candidate_[x] in Response model

### LINKED ISSUE #559: Update response model to cater for nominations
Depends on #558 

Update the [response model](https://github.com/deep-learning-indaba/Baobab/blob/f65917877a0675a300af41cef0c5df269c408d40/api/app/responses/models.py#L5) to add the following fields:
- nomination_title
- nomination_firstname
- nomination_lastname
- nomination_email

Add derived properties called `candidate_title`, `candidate_firstname`, `candidate_lastname` and `candidate_email` to the Response model object, which each apply the following logic (described for `candidate_title`, the other 3 are analogous):
- If the application form has nominations set to True, return the value of `nomination_title`
- Else return user.user_title

Generate a migration to apply the above changes to the database

ChrisCurrin:
this extends e75ad52 and Resolves #559 by creating test cases for getting a response from the repository and checking properties like `candidate_title`, `candidate_email`, etc. 
A test now exists for a "normal" application form
A test now exists for a "nomination" application form

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,10 @@ venv.bak/
 # VSCode project settings
 .vscode
 
+
+# IntelliJ/Pycharm/Webstorm/etc. project settings
+.iml
+
 # Rope project settings
 .ropeproject
 
--- a/api/app/responses/models.py
+++ b/api/app/responses/models.py
@@ -19,39 +19,42 @@ class Response(db.Model):
     nomination_lastname = db.Column(db.String(100), nullable=True)
     nomination_email = db.Column(db.String(255), nullable=True)
 
-    application_form = db.relationship('ApplicationForm')
+    application_form = db.relationship('ApplicationForm', foreign_keys=[application_form_id])
     user = db.relationship('AppUser', foreign_keys=[user_id])
     answers = db.relationship('Answer')
 
     @property
     def candidate_title(self):
-        if application_form.nominations:
-            return nomination_title
+        if self.application_form.nominations:
+            return self.nomination_title
         else:
-            return user.user_title
-    
+            return self.user.user_title
+
     @property
     def candidate_firstname(self):
-        if application_form.nominations:
-            return nomination_firstname
+        if self.application_form.nominations:
+            return self.nomination_firstname
         else:
-            return user.firstname
-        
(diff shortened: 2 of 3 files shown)


ChrisCurrin:
I've been working on the more abstracted version using a Section with a "Nominations" name and pull the relevant questions/answers from that section. I've needed to build up the repositories for getting info from the database, and test cases too of course.

avishkar58:
Hi Chris,

Can you describe in a bit more detail what you're working on? Just want to
make sure we're all on the same page.

Thanks

ChrisCurrin:
Sure :)

I'm working on resolving Issue #591 (ReferenceRequestDetails API).

To get the candidate properties from the Response (previously explicit properties), they can be retrieved by using the related ApplicationForm. ApplicationForm can have a Nominations Section and multiple questions within that section (perhaps a minimum of `title`, `firstname`, `lastname`, `email`) - as suggested [here](https://github.com/deep-learning-indaba/Baobab/issues/560#issuecomment-592940191). The answers to those questions can be retrieved in the ReferenceRequestDetails API and returned as either individual properties 
`
{
  'candidate_title':'Mr',
  ...
}`

or a subobject

`
{
  'candidate':
  {
    'title':'Mr',
    ...
  }, ...
}`

Is this inline with what you were thinking?

avishkar58:
Awesome, that sounds perfect - thank you!
Have a look at this document which describes how the sections and questions
will be set up for the Maathai award (Kambule awards will be very similar):
https://docs.google.com/document/d/1MIt4F8_MIhYQcNFS-wLfQSCsxEYttvA7EhNLbY5pBZI/edit#heading=h.n8mt72nzmyr9


I wonder whether we should introduce an additional field to the section and
question called "keys" or similar, that tags them with standardised names
we can use for these kinds of lookups - I'm worried that querying based on
section or question headline might be flaky! What do you think?
And @amritpurshotam ?

ChrisCurrin:
closing.
See #628

amritpurshotam:
@avishkar58 Sorry the late reply, been a bit hectic at work. Yeah that makes sense like we're doing on the email templates

PR #630, opened 2020-03-05
Add active flag to ReviewResponse

### LINKED ISSUE #614: Add active flag to ResponseReviewer
Update the [ResponseReviewer model](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/responses/models.py#L98) to add a new field called `active` which is of type Boolean and not null. Default it to True in the constructor. 

Generate a migration to apply the change to the database, setting the field to True on all existing records in the ResponseReviewer table.

Update all the queries in the [review api](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/reviews/api.py) that use the ResponseReviewer model to filter by active == True. EXCEPT in the [`get_response_reviewer`](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/reviews/api.py#L242) function 

FYI: We don't want to filter by active in get_response_reviewer so that we don't return a confusing error message to the reviewers in the event that there is a race condition in the logic that drops their review request (#615).

avishkar58:
Resolves #614


amritpurshotam:
@avishkar58 I'll be able to have a look tomorrow morning

api/app/reviews/tests.py
```diff
@@ -170,17 +174,31 @@ def test_one_reviewer_one_candidate(self):
...
+
+        response = self.app.get('/api/v1/review', headers=header, data=params)
+        data = json.loads(response.data)
+
+        self.assertEqual(data['reviews_remaining_count'], 0)
+        self.assertEqual(data['response']['id'], 0)
+
+    @parameterized.expand([
+        (True,), (False,)
+    ])
```
amritpurshotam: Ah this is useful! Something I'll keep in mind for my own tests

avishkar58:
Thanks! Yes exactly, we will set active to false once the two agreeing
reviews come in, but to prevent confusing the 3rd reviewer in case they
were already busy with their review at the time it got deactivated, we will
still accept it.

PR #643, opened 2020-03-23
Glossary backend

SMHall94:
Hi there !

This is another "tutorial" to help onboarding of new contributors with little experience. Not sure if there are some concepts I completely missed - I would be happy to go add those as well :)

### DIFF
--- a/api/glossary_usefulconcepts.md
+++ b/api/glossary_usefulconcepts.md
@@ -27,11 +27,13 @@ Examples include JavaScript, Python
 - **Classes** are used in OOP in the creation of objects. 
 - When comparing high-level to low-level languages, this refers to the level of abstraction. Each level is a step away 
 from the base machine 
-    - Abstraction: the degree to which the characteristics of an object are hidden in order to reduce complexity and increase efficiency. The remaining object is a representation of the original with unwanted detail omitted
+    - Abstraction: the degree to which the characteristics of an object are hidden in order to reduce complexity and 
+    increase efficiency. The remaining object is a representation of the original with unwanted detail omitted
     - Higher level languages have greater levels of abstraction 
 
 **3. Dynamic typing (i.e. _data types_)**
-- As a point of [comparison](https://www.sitepoint.com/typing-versus-dynamic-typing/), static typed languages are Java, C and C++. Static typic requires the explicit declaration 
+- As a point of [comparison](https://www.sitepoint.com/typing-versus-dynamic-typing/), static typed languages are Java,
+ C and C++. Static typic requires the explicit declaration 
 a variable before it is used: 
 
 ```buildoutcfg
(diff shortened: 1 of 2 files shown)


api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+- Compiled:
+written code is converted directly into machine code that the processor executes
+More control over hardware (memory, CPU usage). Examples are C, C++, Go
+- Interpreted: 
+these languages will run through the programme line by line and execute each.
+Examples include JavaScript, Python
+
+**2. Object-orientated programming**
+- This centres around the creation of objects that are characterised by attributes and behaviour. 
+- **Classes** are used in OOP in the creation of objects. High-level versus low-level languages
```
avishkar58: Was "High level versus [...]" meant to be another heading (or part of the next bullet point)?

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+- This refers to the level of abstraction. Each level is a step away from the base machine 
+    - Abstraction: the degree to which the characteristics of an object are hidden in order to reduce complexity and increase efficiency. The remaining object is a representation of the original with unwanted detail omitted
+    - Higher level languages have greater levels of abstraction 
+
+**3. Dynamic semantics**
+- This defines how and when the various constructs of a programming language should produce a programme behaviour
+- Constructs refer to instances of values contained into so-called constructs which exist at runtime. This helps the flow of execution of the programme.
```
avishkar58: No idea what this means :) I suggest removing it, maybe add "Dynamic Typing" instead which is more practically relevant.

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Flask 
+The REST **API** used is Flask. Flask is a Python micro [web-based framework](https://www.fullstackpython.com/web-frameworks.html) that is flexible and simple to implement when creating web-based applications such as Baobab. The main thing to remember about web-frameworks is that they are essentially code libraries that provide common methods / operations for building reliable, scalable, and maintainable web applications
```
avishkar58: REST should be bold too since you defined it below :)

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Classes
+- What is a class?
+    - Classes are fundamental in object oriented programming and centre around creating reusable patterns of code
+    - Classes are blueprints for an object: the blueprint defines a set of attributes that will characterise any object instantiated from a class
+    - An instance of the class is a realised version of the class. In other words, when the class object is run in your _“run.py”_ file, with your respective input parameters.
+- When should write a function over a class:
+    - Function: single, repetitive task e.g. calculating the sum of an array (not all languages have these built-in)
+    - Class: this is the creation of an idea which requires a collection of properties (attributes) pertaining to the idea. These can include a collection of function(s). Functions are referred to as methods. The class will be instantiated repeatedly - without having to build the object from scratch each time.
```
avishkar58: Clarification: functions *on a class* are referred to as methods.

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+Each decorator will modify an aspect of the original function or a class’s behaviour in a specific way without permanently modifying it. The decorator returns a version of the object that was wrapped.
+One can have more than one decorator, and take the function to be modified as input.
+
+For example:
+```buildoutcfg
+@decorator_1
+@decorator_2
+def function_one():
+    pass
+
```
avishkar58: Maybe clarify that the above code is equivalent to the below?

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+[@auth_required](https://flask-security-too.readthedocs.io/en/stable/patterns.html)
+
+These can all be easily Googled for further clarity and descriptions as to how they modify the respective functions.
+
+## Endpoints
+The [API endpoint](https://smartbear.com/learn/performance-monitoring/api-endpoints/) is the end of a communication channel. When there are two interacting APIs, these interact with each other at the endpoint. This is the location from which APIs can access the network resources they need to carry out their functions.
+APIs work using `requests` and `responses` in that they request information from a web application or web server and then receive a response. Endpoints send the requests and indicate where the resource lives. These are stored in the [routes.py](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/routes.py) file in the [Baobab repository](https://github.com/deep-learning-indaba/Baobab).
+
+## Models
+These are an example of a Python object, specifically a database.
```
avishkar58: Maybe say something like "These are Python object representations of some entity that gets mapped to and stored in the database" And then define entity as something like "any person, item, physical or virtual object that has a set of defining properties which need to be stored and may interact with and have relationships to other entities" . The entity and relationship terminology is also relevant to your descriptions about relational databases below.

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+|**id:** int|
+|**name:** str|
+|**descriptions:** str|
+
+**Relational database:**
+
+|ID:|Name:|Description:|
+|---|---|---|
+|INT|Char|VarChar|
+
```
avishkar58: Maybe add that each row in the database corresponds to a single object or instantiation of the class.

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+### Database keys
+Keys are fundamental in the creation of relationships between data (intra- and inter-tables). Further, they assist in maintaining uniqueness in a table; ensure consistent and valid data as well as allow for efficient data retrieval.
+
+|Candidate keys (multiple, one can be a primary key)||
```
avishkar58: I'm a bit confused by this table - what are the rows and columns? Or were you just doing it  to get a 2 column layout? If that's the case then think it would be better to just put headings and bullet points :)

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+|**Composite keys:**| |
+|a.k.a. Compound, composite keys| |
+| More than one columns used to identify a unique value|
+
+### Data types
+These define the type of value that can be stored in a column. 
+This summary of [PostgreSQL data types](https://www.postgresql.org/docs/9.5/datatype.html) will provide more information.
+- Numeric: 
+    - float, real, integers (no decimal, varying precision), decimal (fixed precision (p,s))
+- Characters:
```
avishkar58: Rather say "strings" which I think will be more familiar to Python programmers..

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+    - char, var_char, text
+- Boolean: 
+    - TRUE / FALSE
+- JSON:
+    - JavaScript Object Notation stores and exchanges data consisting of data types that are arrays or of attribute-value type. 
+- XML:
```
avishkar58: We don't use XML and it's considered a bit old-school now :) Maybe remove and replace with BLOB which might be more relevant for us (we're not using it currently, but might in future)

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+    - Extensible markup language (creates information formats and electronically shares structured data via public and corporate networks)
+
+## Universal Resource Locators
+Protocols to locate and access information. These are used for communication between servers and clients. Examples include: [**HTTP**](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol) (hypertext transfer protocols) and [FTP](https://en.wikipedia.org/wiki/File_Transfer_Protocol) (file transfer protocol for file transfers between client and server)
```
avishkar58: Is this right? I wouldn't consider HTTP to be a URL which is what this is suggesting? Rather the communication protocol for interacting with the resource is specified part of a URL as you mentioned in the example...

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+The general syntax is as you are used to seeing in your browser line:
+
+`initiating protocol://web resource`
+
+#### HTTP error codes
+When there is communication between the **client** and **server**, [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are issued in response to a client’s request made to a server.
```
avishkar58: When there is communication between the client and server *using HTTP*

api/glossary_usefulconcepts.md
```diff
@@ -0,0 +1,156 @@
...
+The 2XX codes refer to a successful communication in that the action was received, accepted and understood by the client
```
avishkar58: by the *server*

api/glossary_usefulconcepts.md
SMHall94: I'm really not sure. This was a point I struggled to understand - I hope the new phrasing is better

api/glossary_usefulconcepts.md
SMHall94: I was trying to show the overlap between different keys, but Markdown is limited in terms of tables so it didn't work out as I had hoped. I've changed it to bullets and added full sentences :)

api/glossary_usefulconcepts.md
```diff
@@ -22,23 +24,40 @@ Examples include JavaScript, Python
 
 **2. Object-orientated programming**
 - This centres around the creation of objects that are characterised by attributes and behaviour. 
-- **Classes** are used in OOP in the creation of objects. High-level versus low-level languages
-- This refers to the level of abstraction. Each level is a step away from the base machine 
+- **Classes** are used in OOP in the creation of objects. 
+- When comparing high-level to low-level languages, this refers to the level of abstraction. Each level is a step away
```
avishkar58: Think you still need the "High-level versus low-level languages" - just in this bullet point?

api/glossary_usefulconcepts.md
```diff
@@ -111,46 +145,69 @@ Relational databases set up relationships between items efficiently, and each da
...
+    - JavaScript Object Notation stores and exchanges data consisting of data types that are arrays or of 
+    attribute-value type. 
+- [BLOB](https://en.wikipedia.org/wiki/Binary_large_object):
+    - Binary Large OBject
+    - Collection of [binary data](https://en.wikipedia.org/wiki/Binary_data) stored as a single entity in a database
+     management system (in the case of Baobab, **PostgreSQL**)
+    - These are typically images, audio or multimedia objects  
+
+## Web protocols
+### Universal Resource Locators (URLs)
```
avishkar58: I'd remove this "Universal Resource Locators" heading, and move the "Two prominent examples are HTTP ... " here - because those are examples of web protocols, not of URLs.

PR #647, opened 2020-03-30
Bugfix/circle ci node version

KaleabTessera:
Upgraded node version to fix builds.


KaleabTessera:
Seems to be 2 issues with the build. A docker and node issue. This will hopefully fix the node issue.

PR #655, opened 2020-04-01
Create lint.yml for python linting

avishkar58:


avishkar58:
Park until we've upgraded to Python 3

PR #656, opened 2020-04-02
Add outcome model

### LINKED ISSUE #651: Add Outcome model
## Motivation
At the moment, Baobab has no way to explicitly indicate the outcome of a candidate's application to an event. We previously relied on the existence of an offer to signal that a candidate was accepted, and assumed everyone else was on the waiting list. This does not give us sufficient granularity to determine whether selection has not completed yet, or someone has no chance of being accepted (ie has been rejected). This series of issues aims to introduce the concept of an application outcome to Baobab.

## Requirements
Add a new back-end model to a new models.py file in the new sub-directory (api/app/outcome/models.py) called Outcome which has the following properties:
- id (int, primary key, not null)
- user_id (int, foreign key references AppUser, not null)
- event_id (int, foreign key references Event, not null)
- status (enum[accepted, rejected, waitlist], not null)
- timestamp (datetime, not null)
- latest (bool, not null)

Create a migration to apply the change to the database.

zakkymustafa:
Resolves #651


api/app/outcome/models.py
```diff
@@ -0,0 +1,35 @@
...
+
+    event = db.relationship('Event', foreign_keys=[event_id])
+    user = db.relationship('AppUser', foreign_keys=[user_id])
+    updated_by_user = db.relationship('AppUser', foreign_keys=[updated_by_user_id])
+
+    def __init__(self,
+                 event_id,
+                 user_id,
+                 status,
+                 latest,
```
avishkar58, line 27: Lets remove latest from the constructor parameters and instead default it to True (`self.latest = True` inside the method body), since whenever someone creates a new Outcome, it is by definition the latest one :)

api/migrations/versions/36235e184d09_.py
```diff
@@ -0,0 +1,32 @@
+"""empty message
```
avishkar58, line 1: Strange... it seems like the migration didn't pick up your new model! I suspect it might be because it can't trace a reference to it... Can you try import your model in [__init__.py](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/__init__.py), then create a new migration?

avishkar58:
Hi @zakkymustafa - are you able to make the changes? Hoping to get the outcomes issues completed soon so that they can be tested and used by EEML - thanks!

zakkymustafa:
@avishkar58 have made the changes, but tried to import the model in the __init.py__ but I got a No module error

PR #660, opened 2020-04-12
Feature/nomination (WIP)

### LINKED ISSUE #560: Update response API to cater for nominations

amritpurshotam:
Just testing out this draft mode thing

Also just realised I've been at this issue on and off for over a month 🙈 😲 but I'm almost there

Will resolve #560


amritpurshotam:
There was overlap with #663 so merged this branch into the one created there.

PR #662, opened 2020-04-14
Update Review History

avishkar58:
Update review history to remove Indaba specific items.
Disable event admin links in nav that don't work.

### DIFF
--- a/api/app/reviews/api.py
+++ b/api/app/reviews/api.py
@@ -363,20 +363,14 @@ class ReviewAssignmentAPI(GetReviewAssignmentMixin, PostReviewAssignmentMixin, r
 
         return random.sample(responses, min(len(responses), num_reviews))
 
-review_fields = {
+_review_history_fields = {
     'review_response_id' : fields.Integer,
     'submitted_timestamp' : fields.DateTime(dt_format='iso8601'),
-    'nationality_country' : fields.String,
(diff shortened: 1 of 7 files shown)


api/app/reviews/api.py
```diff
@@ -387,14 +381,12 @@ def __init__(self, review):
         self.review_response_id = review.id
         self.submitted_timestamp = review.submitted_timestamp
         self.reviewed_user_id  = review.AppUser.id
-        final_verdict = [o for o in review.options if str(o['value']) == review.value]
-        final_verdict = final_verdict[0]['label'] if final_verdict else "Unknown"
-        self.final_verdict = final_verdict
```
amritpurshotam, line 392: Just unsure why the final verdict bit is coming out? I can't find a linked issue about this but is it related to the changes we're making to the review system to only having strong yes, yes, no, strong no option?

avishkar58: It's because EEML doesn't have a Final Verdict question on their review form so it was returning nothing :) This is a quick fix for now to get it working, but ultimately we want to move towards something more generic, referring to questions by key if they exist, rather than hard-coding question headlines.

amritpurshotam: Ah yes of course that makes sense, thanks!

PR #663, opened 2020-04-14
Application form nomination

### LINKED ISSUE #561: Update Application Form to cater for nominations
Update the [application form front-end components](https://github.com/deep-learning-indaba/Baobab/tree/develop/webapp/src/pages/applicationForm) to cater for nominations. 

### Background
Certain event application forms (eg Indaba awards) need to allow a user to nominate someone else for the position/award. An application form will have a `nominations` flag set to True if it allows nominations. A user should be able to nominate multiple people. If an application form doesn't allow nominations, it is assumed that any application is for the user applying and the user should only submit a single response. 

### Changes required
The following additional features should be added for application forms that have the `nominations` flag set to `True`. These features should be turned off if the `nominations` flag is `False`.

The response API will return a list of responses. If the application form's `nominations` flag is set to True and a non-empty list of responses is returned, display these in a list to the user (with the candidate title, firstname, lastname, email), so that they can choose one to view or press a button to create a new nomination.

If the `nominations` flag is set to False, display the application form and populate with the first response from the list received from the response API. (ie. keep the behaviour the same as it currently is, but cater for the response API returning a list instead of a single response, and take the first one)

### Definition of Done
- Manually run through test cases with an application form that has nominations set to False (the behaviour should be the same as it was before this change), and set to True, where the new features should be available.
- Submit a PR (select @avishkar58 and @KaleabTessera as reviewers) and address any review comments
- Ensure all tests pass

avishkar58:
Resolves #561


amritpurshotam:
@avishkar58 I think this has overlapped a bit with issue #560 and PR #660 . I have overhauled the response API quite a bit. Was just adding the final tests. Should be done by tonight. And I know sorry I did take quite a while getting it done 🙈 Let me see, I'll try to merge your changes into my branch and see if that makes sense and then I'll create a PR to merge my changes into this branch.

avishkar58:
@amritpurshotam Ah sorry about that! I didn't make much changes to the responses API, just enough to get the front-end working, so hopefully it won't cause too many merge conflicts. Anyway, lets keep these parked for now, until the EEML changes are released to production

amritpurshotam:
@avishkar58 Yeah I also saw it was minimal so I think merging shouldn't be too much of a hassle. And cool.

PR #665, opened 2020-04-15
Combine nomination branches

### LINKED ISSUE #560: Update response API to cater for nominations

amritpurshotam:
Resolves #560 

Note this is merging into branch `applicationFormNomination` and not `develop`

I think I also just want to give this a once over in the morning. It's late so might have missed something.

### DIFF
--- a/api/app/applicationModel/repository.py
+++ b/api/app/applicationModel/repository.py
@@ -11,9 +11,7 @@ class ApplicationFormRepository():
 
     @staticmethod
     def get_by_id(id):
-        return db.session.query(ApplicationForm)\
-            .filter_by(id=id)\
-            .first()
+        return db.session.query(ApplicationForm).get(id)
 
     @staticmethod
     def get_by_event_id(event_id):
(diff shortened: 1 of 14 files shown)


amritpurshotam:
Yeah I agree I think we should. I think as the application gets used more the chance of errors happening increase so rather start now. Let me see what I can do to update that to use the UnitOfWork

PR #669, opened 2020-04-24
Outcome api

### LINKED ISSUE #652: Add Outcome API
## Dependency
Depends on #651 

## Motivation
At the moment, Baobab has no way to explicitly indicate the outcome of a candidate's application to an event. We previously relied on the existence of an offer to signal that a candidate was accepted, and assumed everyone else was on the waiting list. This does not give us sufficient granularity to determine whether selection has not completed yet, or someone has no chance of being accepted (ie has been rejected). This series of issues aims to introduce the concept of an application outcome to Baobab.

## Requirements
Add a new api.py to api/app/outcome with the following APIs:
#### GET method:
- Requires authentication
- Accepts event_id as parameter
- Retrieve and return the latest outcome (has latest flag set) for the logged in user and specified event via a new repository method

#### POST method
- Requires authentication
- Accepts user_id, event_id and outcome as parameters
- Validate that the user_id and event_id exists and that the logged in user is event_admin for the event
- Validate that the outcome is "accepted", "rejected" or "waitlist"
- Query for any existing Outcome instances for the given user_id and event_id and updated the latest flag to 0
- Create a new Outcome instance and add it to the database via a repository method
- Send an email to the user informing them of the outcome decision

#### GET method:
- Requires authentication
- Accept event_id as parameter
- Validate that an event with the given event_id exists
- Validate that the logged in user is an event_admin for the event
- Retrieve the list of Outcomes (with latest flag set to true) for the given event via a repository method
- Return the following fields: [user.user_title, user.firstname, user.lastname, user.email, outcome.status, outcome.timestamp]

Add appropriate routes to [routes.py](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/routes.py)

### LINKED ISSUE #653: Set outcome when adding Offer
## Dependencies
- Depends on #651 
- Depends on #652 

## Motivation
This series of issues adds the concept of an application outcome to Baobab, so that we can explicitly distinguish between people who have been accepted, rejected or wait-listed for an event. One potential issue is that unsynchronised setting of the outcome and offers may result in inconsistent state (for example, a rejected candidate getting an offer). This requirement aims to address this.

## Requirements
Update the [offer API post method](https://github.com/deep-learning-indaba/Baobab/blob/0f9944c26a21de1d1b258feca0c4c66c3c146fb3/api/app/registration/api.py#L135) to add the following functionality
- Retrieve the latest outcome for the given user and event via the outcome repository
- If the latest outcome status is rejected, return an error saying that the candidate has been rejected
- If the latest outcome is waitlist or if there is no outcome, add a new outcome for the user and event with accepted as the status.

avishkar58:
Resolves #652 
Resolves #653


api/app/outcome/api.py
```diff
@@ -0,0 +1,152 @@
...
+                return errors.OUTCOME_NOT_FOUND
+            
+            return outcome
+
+        except SQLAlchemyError as e:
+            LOGGER.error("Database error encountered: {}".format(e))            
+            return errors.DB_NOT_AVAILABLE
+        except: 
+            LOGGER.error("Encountered unknown error: {}".format(traceback.format_exc()))
+            return errors.DB_NOT_AVAILABLE
```
amritpurshotam, line 72: So I was wondering if we should spend some time investigating how to globally handle errors so then we don't need to have try/except blocks in every API call. 

I did briefly look into it now and found this in the [docs](https://flask-restful.readthedocs.io/en/latest/extending.html#custom-error-handlers) but not entirely sure how the logging hooks into flask and ideally I'd like the custom error message to come from a custom exception class (i.e. we'll define a `OutcomeNotFoundError` class for example ) and if it's raised the error code and message come from within the class instead of having to be manually added to the dictionary like in the docs or returned separately like here. Not sure if something like that exists in Python/Flask. 

I know I've done something similar with a C# WebApi before. It really does clean up get/post methods etc, centralises the error handling logic, and then we can have custom error handling for application errors (say we define an `ApplicationError` from which `OutNotFoundError` inherits from) which is then globally caught, or the db errors like `SQLAlchemyError`, or any other kind of error that we want to handle separately. What do you think?

api/app/outcome/api.py
```diff
@@ -0,0 +1,152 @@
...
+    @marshal_with(outcome_fields)
+    def post(self, event_id):
+        req_parser = reqparse.RequestParser()
+        req_parser.add_argument('user_id', type=int, required=True)
+        req_parser.add_argument('outcome', type=str, required=True)
+        args = req_parser.parse_args()
+
+        event = event_repository.get_by_id(event_id)
+        if not event:
+            return errors.EVENT_NOT_FOUND
```
amritpurshotam, line 84: I think just another example here about how the global error handling can work to clean up the code. If we raise an `EventNotFoundError` in the `get_by_id()` method if the event doesn't exist, then we won't have to check if the event exists since the exception would have been raised if it's not there. The exception handler for my hypothetical `ApplicationError` would then take it, do some custom logic (like logging for example but won't be needed here) and then look at the class to find the appropriate message and status code and return that

api/app/outcome/api.py
```diff
@@ -0,0 +1,152 @@
...
+    @event_admin_required
+    @marshal_with(outcome_list_fields)
+    def get(self, event_id):
+        event = event_repository.get_by_id(event_id)
+        if not event:
+            return errors.EVENT_NOT_FOUND
+
+        try:
+            outcomes = outcome_repository.get_latest_for_event(event_id)
+            return outcomes
```
amritpurshotam, line 142: How would we show if an applicant doesn't have an outcome at all yet?

api/app/outcome/tests.py
```diff
@@ -0,0 +1,171 @@
...
+        result = outcome_repository.get_latest_by_user_for_event(self.test_user2.id, self.event1.id)
+        self.assertEqual(result.id, self.event1_user2_outcome.id)
+        self.assertEqual(result.status, Status.REJECTED)
+
+    def test_get_all_by_user_for_event(self):
+        """Test that repository method gets all outcomes for a user."""
+        self.seed_static_data()
+        result = outcome_repository.get_all_by_user_for_event(self.test_user1.id, self.event1.id)
+        self.assertEqual(len(result), 2)
+        self.assertItemsEqual([o.id for o in result], [self.event1_user1_outcome1.id, self.event1_user1_outcome2.id])
```
amritpurshotam, line 68: Ah I learnt something new here, didn't realise there was a method like this to test lists 😄 Will keep this in mind and do this instead of testing each item individually on a new line

api/app/registration/api.py
```diff
@@ -149,6 +151,20 @@ def post(self):
...
+                return errors.CANDIDATE_REJECTED
+            existing_outcome.reset_latest()
+
+        new_outcome = Outcome(
+            event_id,
+            user_id,
+            Status.ACCEPTED,
+            g.current_user['id']
+        )
+        outcome_repository.add(new_outcome)
```
amritpurshotam, line 166: Is this going to be used to register a user who was waitlisted? So it's almost like a shortcut instead of setting their status to accepted first?

api/app/utils/auth.py
```diff
@@ -57,11 +64,28 @@ def wrapper(*args, **kwargs):
...
+        user = _get_user_from_request()
+        if user:
+            user_info = user_repository.get_by_id(user['id'])
+            if user_info.is_event_admin(args['event_id']):
+                g.current_user = user
+                return func(*args, event_id=args['event_id'], **kwargs)
+        
+        return FORBIDDEN
+
+    return wrapper
```
amritpurshotam, line 91: This is quite cool!

api/app/outcome/api.py
avishkar58: 100% agree! We should definitely do this. Most of our APIs have repeated error handling logic which is a definite code smell!

api/app/registration/api.py
avishkar58: Yes exactly, this ensures that we don't have conflicting state, in one case a user has an offer for a place at the event, but on the other hand the outcome says they're waitlisted. Since we're initially going to be doing this by making direct API calls rather than through a GUI, this will be easier than making a call to outcome API first and then adding the offer :)

PR #676, opened 2020-05-03
Siobhan

SMHall94:
Hi, Leonard!
Looking forward to your comments and feedback!

### DIFF
--- a/api/app/users/tests.py
+++ b/api/app/users/tests.py
@@ -594,7 +594,7 @@ class UserProfileListApiTest(ApiTestCase):
 
         # Assert that request succeeds and no users are returned.
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.json), 1)
+        self.assertEqual(response.json, [])
 
         self.assertTrue(False)
 
@@ -627,6 +627,7 @@ class UserProfileListApiTest(ApiTestCase):
         self.assertEqual(len(response.json), 1)
 
         self.assertTrue(False)
+
 
 class UserProfileApiTest(ApiTestCase):
     def setup_static_data(self):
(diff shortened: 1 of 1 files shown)


api/app/users/tests.py
```diff
@@ -566,7 +566,7 @@ def test_response_and_not_invited(self):
 
         # Assert that request succeeds and that users that responded are returned.
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.json), 1)
+        self.assertEqual(len(response.json), 0)
```
lzbotha, line 569: I think there should be users returned here. We  want to check that users that have responded (but are not invited guests) are returned. And the response should include that they are not invited guests (in the field we added).

api/app/users/tests.py
```diff
@@ -594,9 +594,7 @@ def test_no_response_and_invited(self):
 
         # Assert that request succeeds and no users are returned.
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.json, [])
-
-        self.assertTrue(False)
+        self.assertEqual(len(response.json), 1)
```
lzbotha, line 597: Ideally we should check the contents of the response as well. At least check the uid of the user that is returned to make sure it's the right one.

SMHall94: Okay, great. So, as I have it here, it's just checking that there is a response, and the length thereof?
When you say the contents, do you mean type? I hope that's the right word - but, do you mean check if it's a list / dictionary?

api/app/users/tests.py
SMHall94: How do we specify the response that they are not invited guests?

PR #678, opened 2020-05-22
Clean up event status logic

### LINKED ISSUE #505: Check status logic on EventHome
There is some fairly complicated logic on EventHome.js to determine the correct status to display to the user based on a number of different states they could be in: For example, whether they've applied, been accepted, withdrew their application, rejected their offer, are an invited guest etc. Go through and document this logic and look for any cases that we're not covering. Improve the code by moving the status logic to a function and simplify it as much as possible.

### LINKED ISSUE #670: Update Home and EventHome to display outcome status
Create a service to communicate with the outcome api
Update the event home and organisation home to display the candidate's application status from the outcome api

avishkar58:
Resolves #505 
Resolves #670


api/app/events/status.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, 
+                 invited_guest=None, 
+                 application_status=None,
+                 outcome_status=None,
+                 offer_status=None, 
+                 registration_status=None):
+        if invited_guest is None and registration_status is not None and offer_status is None:
+            raise ValueError('Non-invited guest must have an offer to have a registration status')
+        
+        if invited_guest is None and outcome_status is not None and application_status is None:
```
KaleabTessera, line 21: Can an outcome_status be not None, while an application_status is None?

avishkar58: Logically it shouldn't be (hence the check), but technically there is nothing (eg DB constraints) preventing that from happening.

PR #679, opened 2020-05-24
Lb sh update user profile list

smhall97:


api/app/users/tests.py
```diff
@@ -472,6 +422,264 @@ def test_get_comments(self):
...
+        header = self.get_auth_header_for(self.event1_admin.email)
+        params = {'event_id': self.event1_id}
+
+        response = self.app.get('/api/v1/userprofilelist', headers=header, data=params)
+
+        # Assert that request succeeds and no users are returned.
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json, [])
+
+    def test_response_and_not_invited(self):
```
avishkar58, line 550: This seems to be testing the same thing as `test_response_and_candidate` below, but that's doing it better, so rather remove this one.

PR #697, opened 2020-06-17
Feature/miniconf integration

KaleabTessera:
Added basic config to connect to mini-conf.

### DIFF
--- a/api/app/users/api.py
+++ b/api/app/users/api.py
@@ -12,11 +12,11 @@ from app import LOGGER, bcrypt, db
 from app.events.models import EventRole
 from app.users.mixins import (AuthenticateMixin, PrivacyPolicyMixin,
                               SignupMixin, UserProfileListMixin,
-                              UserProfileMixin)
+                              UserProfileMixin,TokenVerificationMixin)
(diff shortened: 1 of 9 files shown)


api/app/users/api.py
```diff
@@ -552,3 +553,18 @@ def put(self):
...
+
+class TokenVerificationAPI(TokenVerificationMixin, restful.Resource):
+    # Check if User is part of Org.
+    @auth_required
+    def post(self):
+        args = self.req_parser.parse_args()
+        organisation_id = args['organisation_id']
+        
+        user = get_user_from_request()
+        user_for_this_event = user_repository.get_by_email(user['email'],organisation_id)
```
avishkar58, line 566: The naming here is a bit confusing... TokenVerification seems very generic "does the user have a valid token" suggests that it's not expired etc. Then the variable name is very specific "is this user participating in this event", but what's actually being checked is if they're part of an organisation... Is this logic temporary for the prototype?

KaleabTessera: Good points, will make the naming a bit better.

PR #709, opened 2020-07-03
Application Form API POST and PUT

### LINKED ISSUE #586: Add POST and PUT to ApplicationFormAPI
## Description
Add a POST and PUSH method to the [ApplicationFormAPI](https://github.com/deep-learning-indaba/Baobab/blob/bf99909f71ef7cca4923d3a93ddcea9e6d443ccf/api/app/applicationModel/api.py#L17)

### POST Method
The POST method should accept an event_id and an application form definition JSON that looks something like the example at the end (without the IDs filled in)

It should then create the application form, the sections and the questions in each section.

### PUT Method
The PUT method is very similar to the POST method, and may share some code, but the idea is to update an existing application form. This means that the method should accept an id for the application_form being update. Note that some sections and questions may be new, or may be updates to existing ones, so we should check whether the IDs are populated first (and validation that the IDs correspond to the application form being updated). 

## Tasks
- Create POST and PUSH methods in the ApplicationFormAPI as described above
- Create a new repository (see events directory for an example) for any new database related code you write
- Clean-up existing database code in the GET method by moving these to the new repository.
- Validate that the logged in user is an event admin for the event whose application form is being updated.

## Definition of Done
- New API methods have been created according to spec
- Tests have been written for new functionality
- A PR has been created to merge into `develop`, reviewed and all review comments addressed.
- All tests pass on the PR

## Example JSON
```
{
  "id": 2,
  "event_id": 2, 
  "is_open": true, 
  "nominations": false, 
  "sections": [
    {
      "description": "Description of the section", 
      "order": 1, 
      "id": 7, 
      "questions": [
        {
          "validation_regex": null, 
          "options": [
           ], 
          "description": null, 
          "headline": "Question 1", 
          "placeholder": "Select an Option...", 
          "is_required": true, 
          "type": "multi-choice", 
          "id": 20, 
          "validation_text": null, 
          "order": 1
        }
      ], 
      "name": "Section 1"
    }, 
    {
      "description": "Description of the section", 
      "order": 2, 
      "id": 5, 
      "questions": [
        {
          "validation_regex": "^\\W*(\\w+(\\W+|$)){0,150}$", 
          "options": null, 
          "description": "Question description", 
          "headline": "Section 2, question 1", 
          "placeholder": "Some question", 
          "is_required": true, 
          "type": "long-text", 
          "id": 15, 
          "validation_text": "You must enter no more than 150 words", 
          "order": 1
        }
      ], 
      "name": "Section 2"
    }
}
```

smhall97:
Thank you for the help!

I still need to spend time figuring out the application form. I'd just like to be sure I'm on the right track :) 

Resolves #586


api/app/applicationModel/api.py
```diff
@@ -79,18 +88,67 @@ def get(self):
...
+
+        app_form = app_repository.get_by_event_id(event_id)
+        if not app_form:
+            app_form = ApplicationForm, Section, Question
+            app_form = app_repository.add(app_form)
+
+        db.session.commit()
+
+        app_id = app_repository.get_by_id(app_form.id)
+        return app_form(app_id), 201
```
avishkar58, line 129: Not sure what this means?

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,67 @@ def get(self):
...
+
+        app_form = app_repository.get_by_event_id(event_id)
+        try:
+            app_form.update(ApplicationForm, Section, Question)
+
+        except IntegrityError:
+            LOGGER.error("Application with id: {} does not exist".format(app_form.id))
+            return FORM_NOT_FOUND_BY_ID
+
+        return {'new_app_form': len(app_form)}, 201
```
avishkar58, line 154: Why len? :)

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,67 @@ def get(self):
...
+        user_id = g.current_user["id"]
+        current_user = user_repository.get_by_id(user_id)
+        if not current_user.is_admin:
+            return FORBIDDEN
+
+        app_form = app_repository.get_by_event_id(event_id)
+        if not app_form:
+            app_form = ApplicationForm, Section, Question
+            app_form = app_repository.add(app_form)
+
```
avishkar58, line 125: Would be good to return an error in the else branch - or better: invert the condition and return an error if the app_form exists, and proceed if it doesn't, something like this:

```
if app_form:
  return APPLICATION_FORM_EXISTS  # A new error

app_form = ApplicationForm, Section, Question
app_form = app_repository.add(app_form)
```

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,67 @@ def get(self):
...
+        current_user = user_repository.get_by_id(user_id)
+        if not current_user.is_admin:
+            return FORBIDDEN
+
+        app_form = app_repository.get_by_event_id(event_id)
+        try:
+            app_form.update(ApplicationForm, Section, Question)
+
+        except IntegrityError:
+            LOGGER.error("Application with id: {} does not exist".format(app_form.id))
```
avishkar58, line 151: Actually an integrity error would occur if you're trying to insert an ID that *already* exists?

api/app/utils/errors.py
```diff
@@ -19,6 +19,7 @@
     {'message': 'No question exists with that Event ID'}, 404)
 FORM_NOT_FOUND = ({'message': 'No form exists with that Event ID'}, 404)
 FORM_NOT_FOUND_BY_ID = ({'message': 'No application form exists with that Application Form ID'}, 404)
+APPLICATION_FORM_EXISTS = ({'message': 'An application form exists with that Application Form ID'}, 403)
```
avishkar58, line 22: Probably should say an An application form already exists for that **event** id :)

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,101 @@ def get(self):
...
+        else:
+            is_open = args['is_open']
+            nominations = args['nominations']
+
+            app_form = ApplicationForm(
+                event_id,
+                is_open,
+                nominations
+            )
+            db.session.add()
```
avishkar58, line 133: you need to pass the app_form to db.session.add :) 
```db.session.add(app_form)```

File: api/app/applicationModel/api.py [hunk 7]
```diff
@@ -79,18 +88,101 @@ def get(self):
...
+
+        section_args = args['section']
+        section = Section(section_args)
```
avishkar58, line 136: You need to create the section objects inside the loop - so loop over section_args and each item in section_args will be the dictionary of fields for that particular section

api/app/applicationModel/mixins.py
```diff
@@ -5,4 +5,7 @@ class ApplicationFormMixin(object):
 
     req_parser = reqparse.RequestParser()
     req_parser.add_argument('event_id', type=int, required=True, help = 'Invalid event_id requested. Event_id\'s should be of type int.')
-    
\ No newline at end of file
+    req_parser.add_argument('is_open', type=bool, required=True)
+    req_parser.add_argument('nominations', type=bool, required=True)
+    req_parser.add_argument('sections', type=dict, required=True, action='append')
+    req_parser.add_argument('questions', type=dict, required=True, action='append')
```
avishkar58, line 11: questions is nested inside sections, so this line is not needed.

api/app/applicationModel/api.py
```diff
@@ -130,21 +130,19 @@ def post(self):
...
             section = section(app_form.id)
-            db.session.add()
+            db.session.add(section)
 
-        for q in question: # TODO: reassess logic for loop (same question as above)
-            question = q(app_form.id, section.id)
-            db.session.add()
+        question_args = args['question']
+        for q in question_args: # TODO: Confirm this makes sense / works
+            question = Question(q)
```
avishkar58, line 143: This isn't how you construct an object in Python?

SMHall94: Thank you for checking! I can't actually explain what I was trying to achieve - I _think_ I have a better idea today though!

api/app/applicationModel/api.py
```diff
@@ -171,7 +171,7 @@ def put(self):
                 section.update(new_s)
                 db.session.commit()
 
-        question = app_repository.get_questions_by_id(question_id) # Confirm correct ID
+        question = app_repository.get_questions_by_id(section.id, app_form.id) # Confirm correct ID
```
SMHall94, line 174: I'd just like to confirm that I understand - is the question is identified by the corresponding section ID and app form ID?
I'm wondering if it's necessary to add a method to "get" the question based on these factors, or if there's a way to get the existing question ID I am missing.

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,120 @@ def get(self):
...
+                event_id,
+                is_open,
+                nominations
+            )
+            db.session.add(app_form)
+
+        section_args = args['section']
+        for s in section_args:
+            section = Section(s)
+            db.session.add(section)
```
SMHall94, line 138: I realised I wasn't quite understanding section_args - I am not sure if this any better though?

api/app/applicationModel/api.py
```diff
@@ -79,18 +88,120 @@ def get(self):
...
+
+        question_args = args['question']
```
SMHall94, line 140: Is this the correct?
I am not sure of the syntax, as I understand this to be nested in "section"?

api/app/applicationModel/api.py
```diff
@@ -117,14 +117,28 @@ def post(self):
         print('SECTION ARGS: ', section_args)
 
         for s in section_args:
-            section = Section(s)
+            section = Section(
+                app_form.id,
```
SMHall94, line 121: I think I am struggling with the ids - it's returning None:

```
IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: section.application_form_id
[SQL: INSERT INTO section (application_form_id, name, description, "order", depends_on_question_id, "key") VALUES (?, ?, ?, ?, ?, ?)]
[parameters: (None, u'Section 1', u'Description of the section', 1, None, None)]
(Background on this error at: http://sqlalche.me/e/13/gkpj)
```

SMHall94: Issue resolved :) Onto the next error message :)

api/app/applicationModel/api.py
```diff
@@ -56,41 +64,144 @@ def get(self):
...
+                nominations
+            )
+            db.session.add(app_form)
+
+        section_args = args['sections']
+        print('SECTION ARGS: ', section_args)
+
+        for s in section_args:
+            section = Section(
+                app_form.event_id,
```
avishkar58, line 122: Why event_id ?

api/app/applicationModel/api.py
```diff
@@ -118,29 +118,31 @@ def post(self):
         print('SECTION ARGS: ', section_args)
 
         for s in section_args:
+            print("Check s_args:", type(section_args))
             section = Section(
-                app_form.event_id,
+                app_form.id,
```
SMHall94, line 123: @avishkar58 
I have been struggling with this from the beginning, and thought I solved it. I'm quite certain I am missing a step, or not understanding something. 

Using `event_id` is the only way I've been able to avoid this error:
`IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: section.application_form_id
[SQL: INSERT INTO section (application_form_id, name, description, "order", depends_on_question_id, "key") VALUES (?, ?, ?, ?, ?, ?)]
[parameters: (None, u'Section 1', u'Description of the section', 1, None, None)]
(Background on this error at: http://sqlalche.me/e/13/gkpj)`

avishkar58: Ah I see - I think what you're missing is a `db.session.commit()` after you do the `db.session.add(app_form)`. This is because "add" essentially adds to a transaction and only when that transaction is committed do the IDs get assigned. Try that and let me know if it fixes the issue? If it doesn't we can debug it together.

SMHall94: That solved the issue! Thank you :D

api/app/applicationModel/api.py
```diff
@@ -56,41 +64,144 @@ def get(self):
...
+
+        user_id = g.current_user["id"]
+        current_user = user_repository.get_by_id(user_id)
+        if not current_user.is_admin:
+            return FORBIDDEN
+
+        app_form = app_repository.get_by_event_id(event_id)
+        if app_form:
+            return APPLICATION_FORM_EXISTS
+        else:
```
avishkar58, line 106: This else isn't needed

api/app/applicationModel/tests.py
```diff
@@ -221,34 +219,34 @@ def test_app_form_is_created(self):
...
         """
-        Confirm the sections to be updated belong to the respective application form
+        Tests that the application form is being created by a user with the correct permissions
         """
-        pass
+        self._seed_data_create()
+
+        response = self.app.post(
+            '/api/v1/application-form',
+            # data=json.dumps(APPLICATION_FORM_POST_DATA),
```
SMHall94, line 244: Will the test for the put method need a new example json?

avishkar58: Yeah it would be best to create a new example JSON with the ID populated from the application form and then some sections with IDs and some without, some questions with IDs and some without so you can test the "existing stuff gets updated" and "new stuff gets created" logic.

api/app/applicationModel/tests.py
```diff
@@ -85,3 +85,168 @@ def test_repo_get_question(self):
...
+            content_type='application/json',
+            headers=self.get_auth_header_for(self.event_admin.email)
+        )
+
+        self.assertEqual(response.status_code, 201)
```
avishkar58, line 249: A successful PUT request should return status code 200 ("OK"). 201 is for "Created"  :)

api/app/applicationModel/tests.py
SMHall94: Thank you ! I realised I don't fully understand what it means for an ID to be populated when it comes to looking at the JSON - I understand the concept, but not sure how to create a section  / question without an ID :)

avishkar58:
Hey @SMHall94 - I updated the PUT method test, I cheated a little bit with the loops, but at least this way it comprehensively covers updates in any of the fields :)

avishkar58 (approved):

PR #722, opened 2020-07-23
Date Picker now highlighted in red when input field is empty

greeffjean:
- FormDate component is now using its own props to listen for the relevant error that my arise from its parent 

- Assigning or not assigning styling to the "date picker html" based on this.


package-lock.json
```diff
@@ -2,145 +2,16183 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@babel/runtime": {
+    "@babel/code-frame": {
```
avishkar58, line 5: I'm not actually sure what this file does :D What resulted in this huge change to it and what are the implications of it?

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -341,6 +355,7 @@ class Section extends React.Component {
         }
       }
     );
+    console.log(this.props)
```
avishkar58, line 358: Please remove debugging message

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -15,7 +15,9 @@ import FormReferenceRequest from "./ReferenceRequest";
 import Loading from "../../../components/Loading";
 import _ from "lodash";
 
+
 const baseUrl = process.env.REACT_APP_API_URL;
+let errorMessage = false;
```
avishkar58, line 20: The introduction of global state makes me uncomfortable, can you explain the logic behind this and is there an alternative that avoids global state?

package-lock.json
greeffjean: Hi, this was a result of setting up tests my side, subsequently this is not needed anymore on my end to please ignore and apologies for not removing this myself prior to committing

webapp/src/pages/applicationForm/components/ApplicationForm.js
greeffjean: Done

webapp/src/pages/applicationForm/components/ApplicationForm.js
greeffjean: Ive made changes here, I studied the flow of the entire js File more in depth and I found a more conventional approach, please review. Still getting familiar with the codebase but its making more sense everyday

webapp/src/components/form/FormDate.js
```diff
@@ -6,6 +6,12 @@ import "./Style.css";
 import * as moment from 'moment';
 
 class FormDate extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.datePicker = React.createRef();
```
avishkar58, line 12: Is this needed? It doesn't seem to be used anywhere?

webapp/src/pages/resetPassword/__tests__/ResetPassword.test.js
```diff
@@ -1,15 +1,22 @@
...
-  expect(wrapper.length).toEqual(1);
+  it('Check if ResetPassword component renders error message in UI', () => {
+    const wrapper = shallow(
+      <ConfirmPasswordResetForm.WrappedComponent />
+    );
+    wrapper.setState({ error: true });
+
+    let errorDiv = wrapper.find(".alert");
+
+     expect(errorDiv.text());
```
KaleabTessera, line 16: Consider changing this to use `toBeTruthy` like `expect(errorDiv.text()).toBeTruthy()` (https://jestjs.io/docs/en/expect#tobetruthy). It is better way to test .text value exists.

KaleabTessera, line 16: Consider changing this to use `toBeTruthy` like `expect(errorDiv.text()).toBeTruthy()` (https://jestjs.io/docs/en/expect#tobetruthy). It is better way to test .text value exists.

greeffjean:
Hi , I made changes please review, the latest updates are more
comprehensive and tests the state, apiCall functions returned value and the
value of the error div.

greeffjean:
Okay Im on it

greeffjean:
Hi Avishkar. 

You mentioned in an earlier post that my fix would affect all date pickers but as far as my knowledge goes it should only change this one thats rendered by the ApplicationForm component, and if this component is only rendering one FormDate per its own render then we can expect the behaviour we want, am I missing something here? I took a slightly different approach with a conditional props listeners method, it listens for a props.id ass well but if Im right then even my previous fix should of worked just fine

avishkar58:
Hi Jean,
This looks better, thanks! Your previous solution would not have worked because it was setting state on the Section component and passing that as a prop to all of the FieldEditors (which are inside a questionModels.map() in Section.render). A Section can have multiple questions and these questions are configurable, so there could be any number of date pickers.

avishkar58:
I just tested this and the date picker is still not turning red when there is a validation error - looks like your CSS selector is wrong.

greeffjean:
Hi, glad we have solved this problem, my previous solution I mentioned in
the above email was actually a more recent update to the very first
solution I provided, I think there may have been some confusion, Please see
the latest commit on iss715, everything is working my end.. I think the
confusion was that you could of been reviewing older review requests I
don't know..

As soon as we resolve the issue Im ready for the next bug!

greeffjean:
I think I was pushing new updates to the branch but didn't re request review

KaleabTessera:
@greeffjean  What is the difference between this PR and https://github.com/deep-learning-indaba/Baobab/pull/730 ? Also please add better descriptions and names to the PRs (like you did in this one), the other PR has no info. 

Also look at a visual diff tools like gitkraken. It makes finding files that shouldn't be commited easier.

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ redis:
 db:
   image: orchardup/postgresql
   ports:
-    - 5432:5432
+    - 4040:5432
```
KaleabTessera, line 8: Please change back in both PRs.

greeffjean: Thank you noted, the difference is **iss710** has tests and fix for the ResetPassword component and **iss715**  has test and fix for the FormDate component

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -205,7 +208,11 @@ class FieldEditor extends React.Component {
             key={"i_" + key}
             showError={validationError}
             errorText={validationError}
-            required={question.is_required} />
+            required={question.is_required}
+            emptyField={this.props.emptyField}
```
avishkar58, line 212: This isn't being used anymore - please remove

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -257,16 +264,20 @@ class Section extends React.Component {
...
         .sort((a, b) => a.question.order - b.question.order),
       hasValidated: false,
-      validationStale: false
+      validationStale: false,
+
     };
   }
 
+
+  // onchange
```
avishkar58, line 273: These comments don't seem useful since they just contain the name of the function on the next line?

PR #723, opened 2020-07-23
Feature/organisation languages

### LINKED ISSUE #717: Add language list to Organisation model
Add a languages property to the [Organisation model](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/organisation/models.py) containing a list of languages that the organisation supports. This can be a JSON column containing a list of JSON objects with code and description. For example:

```
[
  {"code": "en", "description": "English"},
  {"code": "fr", "description": "French"}
]
```

Here the codes are according to [ISO 639-1](https://www.iso.org/standard/22109.html)

Update the [Organisation API](https://github.com/deep-learning-indaba/Baobab/blob/597eec3fd31a606b47ca82a7911fb01930fc5e08/api/app/organisation/api.py#L8) to return this languages property in the get method.

amritpurshotam:
A quick one to get the ball rolling :)

Resolves #717


api/migrations/versions/742a2a556eb9_.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = 'de12a6f382c1'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('organisation', sa.Column('languages', sa.JSON(), nullable=True))
+    op.execute("""UPDATE organisation SET languages = '[{"code": "en", "description": "English"}]'""")
```
amritpurshotam, line 20: @avishkar58 Is AI4D going to be a separate organisation? If I read the previous migration scripts correctly, it looks like there is an AI4D Event under the Deep Learning Indaba? Just wondering in case we need to set it to English and French.

avishkar58: Yup AI4D will be a separate organisation, the AI4D event under the Indaba will remain English only :)

Issue #719, opened 2020-07-22
Update Event model to support translation


avishkar58:
Update the [event model](https://github.com/deep-learning-indaba/Baobab/blob/597eec3fd31a606b47ca82a7911fb01930fc5e08/api/app/events/models.py#L12) by splitting it into two models, `Event` and `EventTranslation` with a one-to-many relationship. Move the `name` and `description` fields to the `EventTranslation` model and add an additional `language` parameter to `EventTranslation` which will contain the 2-character (ISO 639-1) language code.

Update the [EventsAPI](https://github.com/deep-learning-indaba/Baobab/blob/597eec3fd31a606b47ca82a7911fb01930fc5e08/api/app/events/api.py#L313) and [EventsByKeyAPI](https://github.com/deep-learning-indaba/Baobab/blob/597eec3fd31a606b47ca82a7911fb01930fc5e08/api/app/events/api.py#L418) get methods to use the language query parameter passed from the front-end (see #718, default this parameter to "en" if #718 hasn't landed yet) to select the correct translation for the `name` and `description` fields. NOTE: The JSON returned to the front-end shouldn't change - it should just contain the correct language in the `name` and `description` fields.

Update the [EventAPI](https://github.com/deep-learning-indaba/Baobab/blob/597eec3fd31a606b47ca82a7911fb01930fc5e08/api/app/events/api.py#L151) get method to return a dictionary for the `name` and `description` properties where the key is the language code and the value is the translation in that language for the relevant property. For example, the JSON returned by the EventAPI get method should look something like this:

```
{
  "id": 1,
  "name": {
     "en": "Innovation Grants",
     "fr": "subvention à l'innovation"
  },
  "description": {
    "en": "This is an English description",
    "fr": "Ceci est une description française"
  },
  "key": "blah2020",
  ...
}
```

Update the POST and PUT methods to accept incoming JSONs of the form above and insert/update records in the EventTranslation for each language.

See the [multi-language design doc](https://docs.google.com/document/d/1PktZq6kepUCQjatvDUj_-mgwm_YgUHNN5Gyg2lDzz4A/edit?ts=5f1601a4#) for more context

avishkar58:
Hi @amritpurshotam  -  just checking how this is coming along? Keen to see your solution so I can copy it for the application form / other APIs :joy:

amritpurshotam:
[...]

PR #735, opened 2020-07-28
Event Translations

amritpurshotam:
Resolves: #719


api/app/events/api.py
```diff
@@ -306,22 +306,25 @@ def put(self):
...
+        language = request.args['language']
 
         events = event_repository.get_upcoming_for_organisation(g.organisation.id)
         returnEvents = []
 
         for event in events:
-            status = None if user_id == 0 else event_status.get_event_status(event.Event.id, user_id)
-            returnEvents.append(event_info(user_id, event, status))
+            if not event.has_specific_translation(language):
+                return EVENT_WITH_TRANSLATION_NOT_FOUND
```
amritpurshotam, line 325: @avishkar58 Wasn't sure how to error handle this part specifically since if even just one of the upcoming events doesn't have the specified language, the response will return not found?
So should I keep it like this? Default to English (i.e. will all events going forward always have an English translation)? Skip the event entirely (this doesn't feel right though)? Or return the event with the name and description blank (and log errors so we're alerted)?

avishkar58: Hey @amritpurshotam - good question... I think it's fair to assume for now that all events will have an English translation, so how about we default to English and log an error?

amritpurshotam: @avishkar58 Cool will do that then 👍

avishkar58:
Hey @amritpurshotam - do you think you would be able to finish this one off this weekend?

amritpurshotam:
@avishkar58 Yip I should. Gonna work on it tomorrow morning.
Sorry its taking a while, been a bit of a crazy week 🙈

amritpurshotam:
@avishkar58 Gonna need a bit more time (just tomorrow most likely). Just need to finish fixing up the last few tests (pretty much all broke in the project but still a few outstanding)

avishkar58:
Cool no worries, thank you for all your hard work!! :)

api/app/attendance/api.py
```diff
@@ -106,12 +106,12 @@ def post(self):
 
         send_mail(
             recipient=user.email,
-            subject='Welcome to {}'.format(event.name),
+            subject='Welcome to {}'.format(event.get_name('en')),
```
amritpurshotam, line 109: @avishkar58 So you'll see this a lot where I default to English. We do seem to have context of the user in a lot of these cases but I don't think we allow them specify their preferred language (and that's also assuming their preferred language overlaps with the event languages).
Let me know what you think on how to handle this going forward.

api/app/events/api.py
```diff
@@ -167,161 +176,114 @@ def post(self):
...
+            args['review_close'],
+            args['selection_open'],
+            args['selection_close'],
+            args['offer_open'],
+            args['offer_close'],
+            args['registration_open'],
+            args['registration_close'],
+            EventType[args['event_type'].upper()],
+            args['travel_grant'],
+            args['miniconf_url']
```
amritpurshotam, line 209: @avishkar58 I did the TODO here to add the `travel grant` to the request. Doesn't look like the front end is affected by this so shouldn't break anything I think. While I was here also added the `miniconf_url`. Let me know if this is fine.

api/app/events/api.py
```diff
@@ -419,15 +381,24 @@ def get(self):
         args = self.req_parser.parse_args()
 
         user_id = g.current_user['id']
+        language = args['language']
+        if language is None or len(language) > 2:
```
amritpurshotam, line 385: So I think Cyprus was doing something strange here. Looking at my logs in the console I saw this. It was sending a `language` parameter with `en-US` and that test was failing hence why I'm defaulting to `en` if the length of the language parameter is greater than 2 (i.e. the length allowed by the ISO 639-1 standard). 
`172.17.0.5 - - [10/Aug/2020 11:58:34] "GET /api/v1/events?cache_bust=1&language=en-US HTTP/1.1" 401 -`

api/app/attendance/api.py
avishkar58: Hey Amrit, so we do have their preferred language stored in the user_primaryLanguage parameter - but don't worry to change this now because my email templates PR centralises this so I can do it there :)

api/app/events/api.py
avishkar58: Thanks for upholding the scouts rule! :)

api/app/attendance/api.py
amritpurshotam: I saw the email translations PR so my question might be resolved there

api/app/events/api.py
avishkar58: This is an issue with the front-end (or rather a TODO) - I added a plugin that automatically sets the language to the browser default, but I still need to add some logic to filter it to the list of organisation supported languages

PR #743, opened 2020-08-07
Iss734

greeffjean:
Hi Avishkar. 

Please let me know if you need changes made here.


greeffjean:
Not quite sure if I nailed the errorMessage flow you wanted, please give feedback

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ redis:
 db:
   image: orchardup/postgresql
   ports:
-    - 5432:5432
+    - 4040:5432
```
avishkar58, line 8: Remove

greeffjean:
You need to conceptualise what the upload flow will be, will users be required to give a description or is it optional? At the moment all the magic happens when you hit the upload button, once a user types a description after the upload has been triggered, that's where one would need to graft out more code to make sure it finds its little index in the parents state and updates the description value. 

This may not make too much sense now but I wanted to know whats the flow, and then find the most practical and efficient solution. 

Right now just to make things work and look good I disabled any interaction with the description field and made it 50% opaque when the user hits upload file..

greeffjean:
Hi, I think I need to know first what the upload flow has to be as queried above, to avoid wasting time

avishkar58:
Hi Jean,
I added a bit more details to the issue with the user flow. I realise I made a mistake when I initially wrote the issue with the inclusion of the "description" - this is redundant, there should just be a name which is defaulted to the file name and is editable by the user at any point. 
Remember to test that the component still displays the previous files and that the names are still editable if the form is saved and you come back to it (ie. the component should properly initialise from the contents of answer.value) 
Thanks!

PR #744, opened 2020-08-09
Email translation

### LINKED ISSUE #739: Add language to email templates
Update the [email_template model](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/email_template/models.py) to add a language field.

Update the [repository](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/email_template/repository.py) get method to also query on a language parameter

Update all hard-coded emails in the various APIs to use the email_template instead

Update all instances of querying for an email template via the repository to pass the language of the _user that the email is being sent to_ 

Ideally add a utility that makes using email templates and passing the correct language easy.

avishkar58:
Resolves #739 
Waiting on #735 to land then will add event name translation and fix tests.


amritpurshotam:
@avishkar58 If it's okay, I'll look at this tomorrow morning

avishkar58:
> @avishkar58 If it's okay, I'll look at this tomorrow morning

Of course! I see you were super busy today finishing off the other PR - much appreciated! I need to merge your changes and fix the tests tonight anyway.

avishkar58:
Will add a test for the new `email_user` method tomorrow morning

api/app/utils/emailer.py
```diff
@@ -8,6 +8,54 @@
...
+        template_parameters['title'] = user.user_title
+    if 'firstname' not in template_parameters:
+        template_parameters['firstname'] = user.firstname
+    if 'lastname' not in template_parameters:
+        template_parameters['lastname'] = user.lastname
+    if event is not None and 'event_name' not in template_parameters:
+        template_parameters['event_name'] = event.get_name(language) if event.has_specific_translation(language) else event.get_name('en')
+
+    body_text = email_template.template.format(**template_parameters)
+    send_mail(recipient=user.email, subject=subject, body_text=body_text)
```
amritpurshotam, line 57: I think you need to pass the `file_name` and `file_path` parameters here

api/app/utils/emailer.py
```diff
@@ -8,6 +8,54 @@
...
+from app.users.repository import UserRepository as user_repository
+from app.events.repository import EventRepository as event_repository
+
+def email_user(
+    email_template_key, 
+    template_parameters=None, 
+    event_id=None,
+    event=None,
+    user_id=None, 
+    user=None,
```
amritpurshotam, line 21: I'm wondering if it would be cleaner to just pass in the ids or just the objects. I'm leaning more towards the objects since usually I think we do have the object retrieved already or was there a reason you did it this way?

avishkar58: I thought it would be easier for the caller, but yes it does seem that the caller always has the objects anyway so updated as per your suggestion

api/app/utils/emailer.py
avishkar58: Good catch Thanks!!

api/app/responses/api.py
```diff
@@ -192,23 +191,22 @@ def send_confirmation(self, user, response):
...
-                lastname=user.lastname,
-                event_description=event.get_description('en'),
-                question_answer_summary=question_answer_summary,
-                event_name=event.get_name('en'))
-            emailer.send_mail(
-                user.email, 
-                subject, 
-                body_text=body_text, 
-                sender_name=g.organisation.name,
-                sender_email=g.organisation.email_from)
```
amritpurshotam, line 211: Seems like the `sender_name` and `sender_email` aren't being set anymore. I think this is on of the only usages though from all the other emails so probably best for consistency.

api/app/utils/emailer.py
```diff
@@ -8,6 +8,48 @@
...
+        template_parameters['title'] = user.user_title
+    if 'firstname' not in template_parameters:
+        template_parameters['firstname'] = user.firstname
+    if 'lastname' not in template_parameters:
+        template_parameters['lastname'] = user.lastname
+    if event is not None and 'event_name' not in template_parameters:
+        template_parameters['event_name'] = event.get_name(language) if event.has_specific_translation(language) else event.get_name('en')
+
+    body_text = email_template.template.format(**template_parameters)
+    send_mail(recipient=user.email, subject=subject, body_text=body_text, file_name=file_name, file_path=file_path)
```
amritpurshotam, line 51: `organisation` is a field in `event` so I wonder if we could set `sender_name=event.organisation.name` and `sender_email=event.organisation.email_from` (if `event is not None` of course)

api/app/responses/api.py
avishkar58: Yeah, the send_mail function sets sender_name and sender_email to these values anyway, so the previous code was redundant :)

api/app/utils/emailer.py
avishkar58: Relying on the defaulting logic in send_mail which sets it to g.organisation.name and g.organisation.email_from :)

amritpurshotam: Ah perfect! Completely missed that part 🙈

PR #751, opened 2020-08-23
Iss7340

greeffjean:
99% complete please review, need some extra testers and still need to write tests for this


webapp/src/components/form/FormMultiFile.js
```diff
@@ -0,0 +1,130 @@
...
+
+
+    // add file
+    addFile = () => {
+        // variables
+        let condition = true;
+        let addError = false;
+        let handleList = this.state.fileList;
+        // check against empty fields
+        handleList.map(val => {
```
avishkar58, line 32: No need to change it, just FYI, you can do this in a more readable way by using the `all` function:
`const condition = handleList.all(val=>val.file)`  and then `addError` is just `!condition`

webapp/src/components/form/FormMultiFile.js
```diff
@@ -0,0 +1,130 @@
...
+
+        return (
+            <div>
+                <FormGroup>
+                    {this.state.fileList.map(val => {
+                        return <MultiFileComponent className="multi-file-component"
+                            handleUpload={(file, name, del, filePath) => this.handleUpload(file, name, del, filePath)}
+                            errorText={this.props.errorText}
+                            addError={this.state.addError}
+                            del={(file) => this.del(file)}
```
avishkar58, line 117: `this.del` isn't defined? But seems like it should be instead of overloading handleUpload to do both

avishkar58 (changes requested):
Hi @greeffjean - I added a few comments to start, but I need to spend a bit more time testing that this actually does what I'm expecting it to do. Please merge develop into this branch because there are a bunch of changes that aren't in the branch which is preventing me from running it with my local database that's migrated to a later version. Thanks

webapp/src/components/form/FormMultiFile.js
```diff
@@ -0,0 +1,130 @@
...
+                            value={val}
+                        />
+                    })}
+
+                    <button className="add-file-btn" onClick={(e) => this.addFile(e)}>Add File</button>
```
avishkar58, line 122: Just a note that we've recently (in the last 2/3 weeks) added translation support to the app, so now all hard-coded front-end text needs to wrapped to ensure the correct translation gets loaded. To do this, you can change this to: `this.props.t("Add File")`, then:
- Add this import: `import { withTranslation } from 'react-i18next';`
- Change the `export default FormMultiFile` at the bottom to `export default withTranslation()(FormMultiFile)
- Add the text to webapp/public/locales/en/translation.json and webapp/public/locales/fr/translation.json (you can use Google Translate to get the French text

Please do this for all hard-coded text in the UI

webapp/public/index.html
```diff
@@ -74,6 +74,7 @@
     />
     <link rel="manifest" href="%PUBLIC_URL%/favicon/manifest.json" />
     <link href="https://fonts.googleapis.com/css2?family=Open+Sans&family=Roboto&display=swap" rel="stylesheet">
+    <link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
```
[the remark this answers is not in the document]
greeffjean: Done

avishkar58:
Hi Jean,

I tested this and went through it in some detail. The component looks good and functions well in isolation, however, there were some issues when it came to integration into the rest of the form/system. I went ahead and fixed these due to time constraints - we need this component for a new form which we need to start testing ASAP. 

For your reference, the issues were:
- The keys in the French translation file are meant to be in English (this corresponds to the text passed to the t function in the components) 
- Please stop submitting the docker port change
- The API expects answer.value to be a string, but it was set to an object
- I see you were setting the file value to the base64 encoding of the file. While this technically works, this would result in files being saved in our database. We want to save files in our cloud storage instead, which is why we have the separate API call that returns a file ID, and this is what should be used for the file value instead - same as the Markdown editor

Please have a look at the commits I added on this PR to see the changes. 

Thanks

avishkar58 (approved):

greeffjean:
Hi, Im sorry about the Port issue, I have gone and fixed it now myself.

I will look at the commits, thanks for letting me know.

Have a good weekend.

Regards
Jean

PR #754, opened 2020-08-25
Iss6960

greeffjean:
the original branch iss696 contained work from iss7340 witch was weird so I created a branch iss6960, that sorted the issue.


avishkar58 (approved):
Nice catch - thanks!

PR #761, opened 2020-09-03
Include References in Review API Response

### LINKED ISSUE #598: Include references in review API response
## Description
Some application forms (eg for awards) may have questions asking the candidate to provide some contact details for references. These references will be submitted separately by the "referrer". If a review form links to a question on the application form that is of type "reference", the actual references should be retrieved and returned along with the response. 

## Task
- Update the GET methods of the [ReviewAPI](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/reviews/api.py#L111) and [ReviewResponseAPI](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/reviews/api.py#L163) to return a list of all [Reference models](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/references/models.py#L50) which correspond to the response_id (via the [ReferenceRequest model](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/references/models.py#L6)) if any exist. Note: this new list should be in addition to the items already returned, ie. add it to the [ReviewResponseUser object](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/reviews/api.py#L103)
- The following details should be returned for each Reference: title, firstname, lastname, relation from the ReferenceRequest model, uploaded_document from the Reference model.
- Remove everything except the id from the [user_fields](https://github.com/deep-learning-indaba/Baobab/blob/ed9ef8986eeb5d725ce24cffdbbec15b7224c83a/api/app/reviews/api.py#L71)
- Clean up this file by moving all database queries to a repository

smhall97:
Resolves #598


api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,204 @@ def test_total_number_of_pages_when_zero(self):
...
+            Country('Egypt'),
+            Country('Botswana'),
+            Country('Namibia'),
+            Country('Zimbabwe'),
+            Country('Mozambique'),
+            Country('Ghana'),
+            Country('Nigeria')
+        ]
+        db.session.add_all(countries)
+        db.session.commit()
```
avishkar58, line 1318: The organisations, countries and user categories above this line aren't needed because they now get set up by default in the ApiTestCase setUp method :)

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,204 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        # Nominations
+        self.first_user_data = self.add_user('firstuser@mail.com', 'First', 'User', 'Mx')
```
avishkar58, line 1321: To avoid confusion, rather name this `first_user` because it's actually a User object, rather than just the data (stuff named *data generally refers to dictionaries by our conventions)

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,204 @@ def test_total_number_of_pages_when_zero(self):
...
+        # self.other_headers = self.get_auth_header_for("someuser@mail.com")
+
+        db.session.flush()
+
+    def test_get_reference_request_by_response_id(self):
+
+
+        self.seed_static_data()
+        reference_req = ReferenceRequest(1, 'Mr', 'John', 'Snow', 'Supervisor', 'common@email.com')
+        # reference_req2 = ReferenceRequest(1, 'Mrs', 'John', 'Jones', 'Manager', 'john@email.com')
```
avishkar58, line 1477: At this point you'll also want to set up a Reference that links to the ReferenceRequest - it's the data from Reference that we ultimately expect to see in the response from the review API

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,204 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        params = {'event_id': 1, 'application_form_id': 1, 'user_id:': 5, 'sort_column' : 'review_response_id'} # confirm params, check ids
+        reference_request_repository.create(reference_req)
+        responses = [
+            self.app.get('/api/v1/review', headers=self.get_auth_header_for('r1@r.com'), data=params),
+            self.app.get('/api/v1/reference-request/list', data={'response_id': 1}, headers=self.first_headers)
```
avishkar58, line 1483: This isn't needed, it's already tested elsewhere. For this test you just want to check the response from the review API (the first get in this list)

api/app/reviews/tests.py
avishkar58: You should also want to set up more than one of these (ReferenceRequest, Reference pairs) - since a candidate may request references from more than 1 person - you can do this in a new test method if you want to though, to keep this one simple.

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,204 @@ def test_total_number_of_pages_when_zero(self):
...
+        ]
+
+        # Response(1, 5)
+        # reference_request_repository.create(reference_req2)
+
+        data = json.loads(responses.data)
```
avishkar58, line 1489: As it currently stands, responses is a list, so it doesn't have a `data` property. But based on my previous comment it doesn't need to be a list anyway and then this will work :)

api/app/reviews/tests.py
SMHall94: I was quite confused by this part. Is this because once the test is written, I will update the ReviewAPI to return the respective Reference models (corresponding to response_id) - so the second part of the list will effectively be sorted out with that?

api/app/reviews/tests.py
SMHall94: Thanks for clearing this up for me! I didn't understand it at all.

api/app/reviews/tests.py
avishkar58: The 2nd get in this list is a different API that you're not updating? Unless I'm missing something?

api/app/reviews/tests.py
```diff
@@ -1465,27 +1439,47 @@ def static_seed_data(self):
...
+        REFERENCE_DETAIL = {
+            'token': reference_req.token,
+            'uploaded_document': 'DOCT-UPLOAD-78999',
+        }
+        # response = self.app.get(
+        #     '/api/v1/reference', data=REFERENCE_DETAIL, headers=self.first_headers)
+        # self.assertEqual(response.status_code, 201)
+
+        params = {'event_id': 1, 'application_form_id': 1, 'user_id:': 5, 'sort_column' : 'review_response_id',
+                  'reference_detail': REFERENCE_DETAIL} # confirm params, check ids
```
SMHall94, line 1458: I actually have no idea what I'm doing here. This is a produce of copy paste :) 
I realised I wasn't entirely sure how "params" works usually - if there's flexibility in what can be added?

Does referencereq.token handle the linking of the ReferenceRequest and Reference pairs?

api/app/reviews/tests.py
```diff
@@ -1465,27 +1439,47 @@ def static_seed_data(self):
...
+
+        response = self.app.get('/api/v1/review', headers=self.get_auth_header_for('r1@r.com'), data=params)
```
SMHall94, line 1460: I am still struggling with this line. 

If I understand correctly, this is where we return the `data` as described in your previous comment. This comes from the different models - Review and Reference and will be found using the "response_id" as the foreign keys?

api/app/reviews/tests.py
```diff
@@ -1465,27 +1439,47 @@ def static_seed_data(self):
...
 
+        data = json.loads(response.data)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(data), 2) # Check length
 
+    def test_get_reference_request_with_two_references(self):
+        # Initial draft / planning
```
SMHall94, line 1467: This still needs a lot of work - I think I will be able to improve it once I completely understand the previous test.

api/app/reviews/tests.py
avishkar58: So params is all of the inputs that the particular API you're calling expects. In this case you're calling the review API get method and it uses a [request parser](https://github.com/deep-learning-indaba/Baobab/blob/732dd643a99dc46fe99e7211f1e3c3ee71a9ce14/api/app/reviews/api.py#L114) which is defined [here](https://github.com/deep-learning-indaba/Baobab/blob/732dd643a99dc46fe99e7211f1e3c3ee71a9ce14/api/app/reviews/mixins.py#L3)  and that specifies what inputs the API expects. In this case it's just `event_id` which is required and `skip` which is optional

api/app/reviews/tests.py
avishkar58: That's correct, but remember it's the API's job to do that joining between Review and Reference, not the test... the test is to ensure that the API does it correctly, perhaps its confusing that the API doesn't currently do it, but when you're writing the test you need to pretend that you've finished the task and are now checking that it was done correctly :)

SMHall94: Ah, yes. Okay. It's clicked for me now :D Thank you!

api/app/reviews/tests.py
SMHall94: I completely forgot about the request_parser ... but, I understand better now how it fits in!
That obviously makes way more sense than what I was thinking - everything links back to the event_id

api/app/reviews/tests.py
SMHall94: Good morning,@avishkar58! Would you mind having another look at my test, please?

I've also been wondering, how does one debug / test their test normally?

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,200 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        self.first_headers = self.get_auth_header_for("firstuser@mail.com")
+
+        db.session.flush()
+
+    def test_get_reference_request_by_event_id(self):
+        self.seed_static_data()
+        reference_req = ReferenceRequest(1, 'Mr', 'John', 'Snow', 'Supervisor', 'common@email.com')
+        reference_request_repository.create(reference_req)
+        REFERENCE_DETAIL = {
```
avishkar58, line 1434: You still need to add the reference to the system, like you do in the next test with the call to `post`

avishkar58 (changes requested):
Great job @SMHall94 ! This test is very nearly there :) 
Good question about how to test your tests... the best guidelines I know of are:
- Keep your test code as straightforward as possible
- Make sure your test fails first and that it fails for the reason you expect it to (for example, if you know the API only returns 5 items in its response dictionary, and your test checks that it has 6 items in the dictionary, make sure the error message says expected 6 != actual 5 or something along those lines) 
- You can also run your test through a debugger to see exactly whats happening and this can also help to step in to the code under test (for example the API its calling) to see whats happening

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,200 @@ def test_total_number_of_pages_when_zero(self):
...
+        }
+
+        params = {'event_id': 1}
+        response = self.app.get('/api/v1/review', headers=self.get_auth_header_for('r1@r.com'), data=params)
+
+        data = json.loads(response.data)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(data), 6) # len as counted by number of data (dict) items
+
+        self.assertEqual(data['references'][0], REFERENCE_DETAIL)
```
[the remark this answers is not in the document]
SMHall94: I didn't know this until right now - I checked the docs for [assertEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual) and it seems technically it could work, but advises using [type specific methods](https://docs.python.org/2/library/unittest.html#type-specific-methods) to get more detailed errors. I've changed it to assertDictEqual to be more precise.

SMHall94:
Thank you for the testing-tests tips :D That's really helpful.

api/app/reviews/tests.py
```diff
@@ -1484,6 +1488,3 @@ def test_get_reference_request_with_two_references(self):
         self.assertDictEqual(data['references'][1], REFERENCE_DETAIL_2)
         self.assertEqual(data['references'][1]['token'], reference_req2.token)
         self.assertEqual(data['references'][1]['uploaded_document'], 'DOCT-UPLOAD-78979')
-
```
SMHall94: @avishkar58 I am getting quite a few "foreign key constraint" errors with the test. The current one is for lines 1436 and 1463.
I'm just wondering if it's correct to continue checking the `static_seed_data`? I'm not sure when to move away from fixing the test :)

avishkar58: Ah so the reason for that is because the first constructor parameter of ReferenceRequest is a response_id and the value of 1 that you're passing here doesn't correspond to a Response because of line 1420 which is commented out :) So the test_response that you create inside seed_static_data isn't actually being added to the database. In fact, instead of passing 1 in the ReferenceRequest constructor, it would be better to pass `self.test_response.id`  -  because there's actually no guarantee that the first response you create will have id of 1, using `self.test_response.id` will guarantee that you're referring to the correct response object 

`reference_req = ReferenceRequest(self.test_response.id, 'Mr', 'John', 'Snow', 'Supervisor', 'common@email.com')`

Regarding when to move away from fixing the test, that should only be when the test is failing on one of your assertion lines and for a reason that you expect it to (ie that the feature you're testing hasn't been implemented yet). If the test is failing in the setup code then you definitely need to keep fixing the test :)

api/app/reviews/api.py
```diff
@@ -124,8 +130,10 @@ def get(self):
         skip = self.sanitise_skip(args['skip'], reviews_remaining_count)
 
         response = review_repository.get_response_to_review(skip, g.current_user['id'], review_form.application_form_id)
+        print(response)
```
SMHall94, line 133: Good morning, @avishkar58 ! I've been sitting with this for a few days now and I'm out of ideas. The `response` is returning `None` and I'm not sure where the error is!

avishkar58: Hey @SMHall94 ! So the reason for this is because the test it's requesting a response for "firstuser" (the logged in user) to review, but "firstuser" doesn't have any responses assigned to them. You can add a ResponseReviewer in your seed_static_data method to assign firstuser a response to review - see [here](https://github.com/deep-learning-indaba/Baobab/blob/6c9b8d4d2051273ca23493d7012b18710f6f0839/api/app/reviews/tests.py#L493) for example. The two IDs passed to the constructor are the response_id and the reviewer_user_id, so you can do something like this in your seed_static_data method:

```
response_review = ResponseReviewer(self.test_response.id, self.first_user.id)
db.session.add(response_review)
db.session.commit()
```

api/app/reviews/tests.py
```diff
@@ -1518,6 +1518,50 @@ def test_get_reference_not_yet_submitted(self):
         self.assertEqual(len(data), 6)
         self.assertEqual(len(data['references']), 0)
 
+    def test_get_reference_submitted_later(self):
```
SMHall94, line 1521: @greg-newman 
This is the test :)
This [link](https://docs.sqlalchemy.org/en/13/errors.html#error-bhk3) made a lot of sense, so I'm not sure now if my test is dodgy:) 
This is the error:

`======================================================================
ERROR: In this test, a reference has been requested, but the reference is only submitted after the second check
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/api/app/reviews/tests.py", line 1540, in test_get_reference_submitted_later
    'token': reference_req.token,
  File "/env/local/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 287, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/env/local/lib/python2.7/site-packages/sqlalchemy/orm/attributes.py", line 718, in get
    value = state._load_expired(state, passive)
  File "/env/local/lib/python2.7/site-packages/sqlalchemy/orm/state.py", line 652, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/env/local/lib/python2.7/site-packages/sqlalchemy/orm/loading.py", line 946, in load_scalar_attributes
    "attribute refresh operation cannot proceed" % (state_str(state))
DetachedInstanceError: Instance <ReferenceRequest at 0x7f0df593b990> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/13/bhk3)
`

avishkar58: Hey! Sorry for the drive-by comment, is this error coming from line 1540? We see this error very often in our tests, it's something wrong with our setup and/or a bug in the version of SQLAlchemy we're on. Basically, if you try to access a property of a database-connected object *after* making a call to the API, you get this error. So after you've done a self.app.get (or post or put), you can't access properties on database objects anymore. The work-around is to access the properties before doing any API calls, so if you move lines 1539 to 1542 before line 1530 it should work.

avishkar58: Would be so amazing If you can figure out how to fix it :D

SMHall94: @avishkar58 Thank you! :D 
I definitely remember this issue coming up when we were working together, I just wanted to be sure before falling on that!

avishkar58: I wonder if we could globally disable lazy loading for tests?

SMHall94:
Hey, @avishkar58

@greg-newman and I are ready to submit our PR! 
Thank you for all the help, we're looking forward to your comments and feedback :D

api/app/reviews/tests.py
greg-newman: Thanks @avishkar58!

api/app/reviews/api.py
```diff
@@ -90,21 +94,41 @@
...
     'user': fields.Nested(user_fields),
     'reviews_remaining_count': fields.Integer,
-    'review_response': fields.Nested(review_response_fields)
+    'review_response': fields.Nested(review_response_fields),
+    'references': fields.List(fields.Nested(reference_fields)),
 }
 
+
 class ReviewResponseUser():
-    def __init__(self, review_form, response, reviews_remaining_count, review_response=None):
```
avishkar58, line 102: Why did you remove review_response? Its still used in line 194 and also in the review_fields marshalling

api/app/reviews/api.py
```diff
@@ -90,21 +94,41 @@
...
+    def __init__(self, review_form, response, reviews_remaining_count, reference_response=None):
```
avishkar58, line 116: Nit: this parameter name is singular, but you're actually expecting a list, so would be better to name it `reference_responses` (plural)

api/app/reviews/api.py
```diff
@@ -396,11 +416,11 @@ def get(self):
         limit = args['limit']
         sort_column = args['sort_column']
 
-        reviewer = db.session.query(AppUser).get(user_id).is_reviewer(event_id)
+        reviewer = review_repository.get_reviewer(event_id, user_id)
         if not reviewer:
             return FORBIDDEN
 
-        form_id = db.session.query(ApplicationForm.id).filter_by(event_id = event_id).first()[0]
+        form_id = review_repository.get_form_id(event_id)
```
avishkar58, line 423: I realise you're just replacing the existing code, but it would be better to update the `get_review_history` function and associated query on review_repository to accept an `event_id` rather than a `form_id`, then this additional step of getting the form_id won't be necessary.

api/app/reviews/repository.py
```diff
@@ -213,6 +214,45 @@ def get_review_complete_timeseries_by_event(event_id):
                         .all())
         return timeseries
 
+    @staticmethod
+    def get_candidate_response_ids(event_id, reviewer_user_id, reviews_required):
+        candidate_responses = db.session.query(Response.id) \
```
avishkar58, line 219: Nit: rather call this `candidate_response_ids` so that it's clear to the reader that it's just the IDs and not the response objects.

api/app/reviews/repository.py
```diff
@@ -213,6 +214,45 @@ def get_review_complete_timeseries_by_event(event_id):
...
+
+    @staticmethod
+    def get_already_assigned(reviewer_user_id):
+        already_assigned = db.session.query(ResponseReviewer.response_id) \
+            .filter(ResponseReviewer.reviewer_user_id == reviewer_user_id) \
+            .all()
+        return already_assigned
+
+    @staticmethod
+    def get_reviewer(event_id, user_id):
```
avishkar58, line 239: I don't think this function really belongs in this repository because it's doing more than just querying the database (by calling is_reviewer on the AppUser object). On line 419 of api.py, rather use the user_repository to get the user and call is_reviewer there instead.

api/app/reviews/api.py
```diff
@@ -396,11 +416,11 @@ def get(self):
         limit = args['limit']
         sort_column = args['sort_column']
 
-        reviewer = db.session.query(AppUser).get(user_id).is_reviewer(event_id)
+        reviewer = review_repository.get_reviewer(event_id, user_id)
```
avishkar58, line 419: Rather use user_repository to get the user and call is_reviewer on the result of that instead (see similar comment in repository)

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        reviewer1 = AppUser('r1@r.com', 'reviewer', '1', 'Mr', password='abc', organisation_id=1, )
+        candidate1 = AppUser('c1@c.com', 'candidate', '1', 'Mr', password='abc', organisation_id=1, )
+        system_admin = AppUser('sa@sa.com', 'system_admin', '1', 'Ms', password='abc', organisation_id=1, is_admin=True)
+        event_admin = AppUser('ea@ea.com', 'event_admin', '1', 'Ms', password='abc', organisation_id=1)
+        users = [reviewer1, candidate1, system_admin, event_admin]
+        for user in users:
+            user.verify()
+        db.session.add_all(users)
+        db.session.commit()
```
avishkar58, line 1307: Please use `self.add_user` and/or `self.add_n_users` from testing.py instead of this.

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        event_roles = [
+            EventRole('admin', 4, 1),
+            EventRole('reviewer', 1, 1)
+        ]
+        db.session.add_all(event_roles)
+        db.session.commit()
```
avishkar58, line 1314: Please use `event.add_event_role()` instead (see [here](https://github.com/deep-learning-indaba/Baobab/blob/a5b4fa192dfe86869b7df01bcb5cbb8914b69b24/api/app/outcome/tests.py#L21) for an example)

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        application_form = [
+            self.create_application_form(1, True, False),
+        ]
+        db.session.add_all(application_form)
+        db.session.commit()
```
avishkar58, line 1320: These lines aren't necessary because self.create_application_form calls session.add already.

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+        ]
+        db.session.add_all(review_forms)
+        db.session.commit()
+
+        self.test_response = Response(  # Nominating other
+            1, self.first_user.id)
+        self.test_response.submit()
+
+        db.session.add(self.test_response)
+        db.session.commit()
```
avishkar58, line 1336: Please use `self.add_response` instead

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+
+        response_review = ResponseReviewer(self.test_response.id, self.first_user.id)
```
avishkar58, line 1338: Please add a new method to testing.py called add_response_reviewer() which does this. 
We prefer to centralise the functions that create entities for tests in one place so that if someone adds a constructor parameter to an entity in future then we can just update it in one place in testing.py instead of in lots of different test files :)

api/app/reviews/tests.py
```diff
@@ -1285,3 +1288,246 @@ def test_total_number_of_pages_when_zero(self):
...
+        """
+        self.static_seed_data()
+
+        params = {'event_id': 1}
+        response = self.app.get('/api/v1/review', headers=self.get_auth_header_for('firstuser@mail.com'), data=params)
+
+        data = json.loads(response.data)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(data), 6)
+        self.assertEqual(len(data['references']), 0)
```
avishkar58, line 1533: Very comprehensive tests, NICE! :D

avishkar58 (changes requested):
Hey @SMHall94 and @greg-newman - this looks great thank you very much! :) 
I've left a few comments mostly with suggestions for improvements.
Please also fix the broken tests and merge conflicts 
Thanks!!

avishkar58:
Oh actually I see the build is failing because of this "ImportError: No module named clsregistry" - I'll take a look at that one - its a consequence of us still being on Python 2.7 🙈🙈🙈🙈

avishkar58:
Actually lets see what happens after you merge develop into this branch, I think I did fix that clsregistry error already.

SMHall94:
@greg-newman @avishkar58 
Hi there!

I think I managed to merge develop into this branch okay - it seems it's only the unit tests that are failing now? I still need a bit more time to work through the other changes and fixes, so hopefully that will resolve all the errors! :D

avishkar58:
Nice one @SMHall94 !
The unit test failures are now genuine failures resulting from your changes - so please have a look and fix them and let me know if you need any help with that :)

api/app/reviews/tests.py
SMHall94: @avishkar58, I see someone already added [add_response_reviewer](https://github.com/deep-learning-indaba/Baobab/blob/a5b4fa192dfe86869b7df01bcb5cbb8914b69b24/api/app/utils/testing.py#L355).

I've left a comment in my new commit. I'm struggling with this and the self.add_response :)

api/app/reviews/tests.py
```diff
@@ -1518,17 +1514,9 @@ def static_seed_data(self):
         db.session.add_all(review_forms)
         db.session.commit()
 
-        self.test_response = Response(  # Nominating other
-            1, self.first_user.id)
-        self.test_response.submit()
+        self.test_response = self.add_response(1, self.first_user.id, language='en')
```
SMHall94, line 1517: @avishkar58 
I think I'm not clicking on something, but this and line `1519` are returning a `400` in the tests (not 200). 

Do you have any hints for me by any chance, please? :)

api/app/reviews/api.py
SMHall94: I updated this, and there are no new errors in my test and the following test is still passing:
- `test_review_history_returned (app.reviews.tests.ReviewsApiTest) ... ok`

I'm not marking as resolved yet, because I still need to fix the removal of `review_responses` and reassess the tests

api/app/reviews/tests.py
```diff
@@ -1486,16 +1486,11 @@ def static_seed_data(self):
...
-        for user in users:
-            user.verify()
-        db.session.add_all(users)
-        db.session.commit()
-
+        # TODO: S: confirm this works once 400 / 200 error is resolved
+        self.system_admin = self.add_user('sa@sa.com', is_admin=True)
+        non_admin_users = self.add_n_users(3)
+        users = non_admin_users.append(self.system_admin)
+        # TODO: confirm indices
```
SMHall94, line 1493: This links to this [comment](https://github.com/deep-learning-indaba/Baobab/pull/761#discussion_r573976712)

api/app/reviews/tests.py
SMHall94: I'm resolving this comment, because it has been updated, but not confirmed. The new comment can be found here.

api/app/reviews/api.py
```diff
@@ -121,12 +138,23 @@ def _add_reviewer_role(user_id, event_id):
...
 
 class ReviewResponseUser():
-    def __init__(self, review_form, response, reviews_remaining_count, language, review_response=None):
+    def __init__(self, review_form, response, reviews_remaining_count, language, reference_responses, review_response=None):
         self.review_form = _serialize_review_form(review_form, language)
+
         self.response = response
         self.user = None if response is None else response.user
         self.reviews_remaining_count = reviews_remaining_count
-        self.review_response = review_response
```
avishkar58, line 129: This line also needs to be restored

api/app/reviews/repository.py
```diff
@@ -193,11 +195,11 @@ def get_review_list(reviewer_user_id, event_id):
         return reviews
 
     @staticmethod
-    def get_review_history_count(reviewer_user_id, application_form_id):
+    def get_review_history_count(reviewer_user_id, event_id):
         count = (db.session.query(ReviewResponse)
                         .filter(ReviewResponse.reviewer_user_id == reviewer_user_id)
                         .join(ReviewForm, ReviewForm.id == ReviewResponse.review_form_id)
-                        .filter(ReviewForm.application_form_id == application_form_id)
+                        .filter(ReviewForm.application_form_id == event_id)
```
avishkar58, line 202: You'll need to join to ApplicationForm and then filter that. for example:
```
.join(ApplicationForm, Reviewform.application_form_id == ApplicationForm.id)
.filter(ApplicationForm.event_id == event_id)
```

api/app/reviews/tests.py
avishkar58: Hey Siobhan, can you update which line/test you're referring to here?

avishkar58: Oh nevermind, I see where it's coming from now. So status code 400 means "bad request" and this would be thrown by the request parser if the request was either missing a required parameter or you passed data with the wrong datatype to a parameter (eg it was expecting an integer but you passed a string). In this case, you can see the Review API's request parser is defined [here](https://github.com/deep-learning-indaba/Baobab/blob/a5b4fa192dfe86869b7df01bcb5cbb8914b69b24/api/app/reviews/mixins.py#L4) - and it has 2 required parameters, `event_id` and `language`. In your tests, for example in line 1533, you're only passing the event_id which is why you're getting the 400. Hope that makes sense. Have a look [here](https://github.com/deep-learning-indaba/Baobab/blob/a5b4fa192dfe86869b7df01bcb5cbb8914b69b24/api/app/reviews/tests.py#L195) for an example that passes both the event_id and the language.

smhall97:
Hi, @greg-newman and @avishkar58 !

After Greg fixed the merge conflicts yesterday, I updated the branch today. The merge conflict is resolved, but I really don't understand the error message that is coming up - as it doesn't do it on every run, and sometimes changes tests that bring up the error. I'm copying one of the errors below, as it seems to follow the same pattern:

`ERROR: In this test, two references have been requested, but only one submitted (two ids exists, one reference submitted)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/api/app/reviews/tests.py", line 1661, in test_two_references_only_one_submitted
    self.static_seed_data()
  File "/code/api/app/reviews/tests.py", line 1489, in static_seed_data
    non_admin_users = self.add_n_users(3)
  File "/code/api/app/utils/testing.py", line 127, in add_n_users
    user = self.add_user(email, firstname, lastname, title, password, organisation_id, is_admin, post_create_fn)
  File "/code/api/app/utils/testing.py", line 107, in add_user
    db.session.commit()
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/scoping.py", line 163, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 1042, in commit
    self.transaction.commit()
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 504, in commit
    self._prepare_impl()
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 483, in _prepare_impl
    self.session.flush()
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2536, in flush
    self._flush(objects)
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2678, in _flush
    transaction.rollback(_capture_exception=True)
  File "/env/lib/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 70, in __exit__
    with_traceback=exc_tb,
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2638, in _flush
    flush_context.execute()
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/unitofwork.py", line 422, in execute
    rec.execute(self)
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/unitofwork.py", line 589, in execute
    uow,
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 245, in save_obj
    insert,
  File "/env/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 1136, in _emit_insert_statements
    statement, params
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1011, in execute
    return meth(self, multiparams, params)
  File "/env/lib/python2.7/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1130, in _execute_clauseelement
    distilled_params,
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1317, in _execute_context
    e, statement, parameters, cursor, context
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1511, in _handle_dbapi_exception
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1277, in _execute_context
    cursor, statement, parameters, context
  File "/env/lib/python2.7/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
ProgrammingError: (sqlite3.ProgrammingError) You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings.
[SQL: INSERT INTO app_user (email, firstname, lastname, user_title, nationality_country_id, residence_country_id, user_gender, affiliation, department, user_disability, user_category_id, "user_dateOfBirth", "user_primaryLanguage", password, active, is_admin, is_deleted, deleted_datetime_utc, verified_email, verify_token, policy_agreed_datetime, organisation_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)]
[parameters: ('Raven-Symone.Brooks4@bestemail.com', 'Raven-Symon\xc3\xa9', 'Brooks', 'Rev', None, None, None, None, None, None, None, None, None, '$2a$12$if1xgwBNKIkg9RMQEh5GkOu4ZUH1tZMIiyNRLZD03gnb1VKyNAofW', 1, 0, 0, None, 1, 'ac0234e6-c33a-488e-8f63-98a1e6bfe38e', '2021-03-05 07:34:24.540268', 1)]
(Background on this error at: http://sqlalche.me/e/13/f405)
`

avishkar58 (approved):
Excellent work! Thank you @smhall97 and @greg-newman ! 
Will just merge the latest changes from develop and then merge. 
@smhall97 The reason you're seeing those intermittent test errors seems to be because your local docker container is still running Python 2 - please rebuild your containers by running `docker-compose build` and that should sort it out :)

Issue #768, opened 2020-09-04
Off centre presentation of titles on some browsers (web)


SMHall94:
![Off centre](https://user-images.githubusercontent.com/43542098/92277738-f00bf300-eef3-11ea-930b-ab8648f72141.png)

avishkar58:
Can you confirm which browser(s) this is happening on?

SMHall94:
Yes - Manjaro operating system with a Mozilla Firefox browser. This is the only instance reported out of 8 testers

SMHall94:
I just tested this on Firefox on Ubuntu and it has the same off centre presentation of the "Welcome to Baobab". It works fine in Chrome

avishkar58:
This still occurs on a number of pages, still need to go through all pages in Firefox to see which ones and ideally update some global style to correct it. At the very least, the centered and left aligned components should be consistent between Chrome and Firefox

PR #771, opened 2020-09-07
Iss677

greeffjean:
Hi, Please review new look, colors sheme and new background has been added. 

Look out for new fonts and: 
softer borders,
textured background lightgrey ( its actually a Baobab trunk ).
new fonst colors,
font sizes,
footer sitting at bottom
user dropdown modifications
logo img slightly bigger

I still want to get my hands on the Apllication section of the App.


avishkar58:
Hi @greeffjean 

This is fantastic! Thanks for taking the initiative here! 
One minor thing I noticed is the footer overlaps the last table slightly:

Screenshot 2020-09-07 at 14 53 51

While you're making style changes, I wonder if you can also help with issue #768? A bunch of stuff is not centred correctly on Firefox (definitely Mac and Linux, not sure about Windows) - for example on Home.js:

Screenshot 2020-09-07 at 14 54 39

and EventHome.js:

Screenshot 2020-09-07 at 14 56 31

avishkar58:
One more thing: Can you try a lighter background colour on the status for the event table? (The grey background behind the "Thank you for attending!" and "In Progress" etc) - the reason being that I think the button gets lost (for example in the 2nd screenshot it's hard to tell that "Continue" is a button because the colour is quite similar to the background and is also the same shape as the grey area)

webapp/src/App.js
```diff
@@ -282,7 +282,22 @@ class AppComponent extends Component {
       user: {},
       collapsed: true,
       eventKey: null,
-      currentEvent: null
+      currentEvent: null,
+      styling: {
+        logo: {
```
avishkar58, line 287: Is this just temporary while you're experimenting? (styles in the component state rather than the CSS)

greeffjean:
Hi Avishkar.

I'll get back on some more work later this week my PC crashed and sent it
to the repair guys yikes!

greeffjean:
Resolves #768, Please review here I am not sure I did what exactly what you asked, please review: 

1. Footer overlap 
2. Status Buttons
3. Text alignment issue on iss768 ( Not too sure what the issue is but added some styling rules that may solve this )

PR #790, opened 2020-09-22
Response List Component

### LINKED ISSUE #779: Create ResponseList page and component
### Motivation
As part of the review improvements in Baobab, we want event admins to be able to see a list of responses to their event's application form. Since each application form is unique, they should be able to customise what information from the response appears on this list.

### Dependencies
Depends on the backend APIs created in #774 and #775 but can be implemented using mock data

### Requirements
- Create a new page called ResponseList in webapp/src/pages/responseList
- Add a route to the page in EventHome.js
- Add a menu item pointing to this page under the Event Admin menu in App.js
- Create a new service that talks to the ResponseList API created in #774 (use mock data if that issue hasn't been completed) 
- Create a component to implement the page's functionality
- Add a checkbox labelled "Include un-submitted"
- Add a dropdown list containing questions that the user can select and a "Add to table" button that adds the selected question to the table
  - The list of questions can be obtained by querying the back-end API created in #775
  - Maintain a state of selected `question_ids` (note "add to table" appends to this list) 
- Submit a request to the back-end API on component render, whenever the "Include un-submitted" checkbox changes and whenever a new question is added to the list of selected questions
- The component should render a table containing the details that come back from the API. This table should be sortable by clicking on the column heading
  - See the [InvitedGuest component](https://github.com/deep-learning-indaba/Baobab/blob/d57978b2e55719ea82f2c6e416b5d97994265b5e/webapp/src/pages/invitedGuests/components/InvitedGuestComponent.js#L361) where we use ReactTable
  - Each record in the response from the API will include a list of reviewers, this should be expanded into columns. So, for example, if there are two entries in the reviewers list, there should be 2 columns called "Reviewer 1" and "Reviewer 2" which contain the names of the reviewers (or blank when the entry is null). The length of the list will be the same for all records in a given API response (but note that different events can have different number of reviewers, so you can't hard-code the number of columns!) 
  - Note that the answers corresponding to the selected question_ids will be in a nested list in the response JSON, but should be included as columns in the table
  - Adjust the display of the answer depending on the type:
    - If type is "file" or "multi-file", render links to the files
    - If type is "long-text" display a truncated version of the answer depending on what will fit reasonably in the table. When the user hovers over the cell, it should display the full answer in a tooltip
    - If the type is "multi-choice" or "multi-checkbox", lookup the answer value in the options and render the relevant label
    - See the [AnswerValue](https://github.com/deep-learning-indaba/Baobab/blob/d57978b2e55719ea82f2c6e416b5d97994265b5e/webapp/src/pages/applicationForm/components/ApplicationForm.js#L467) in ApplicationForm.js for inspiration
- Add tests for the functionality

## Example JSON Request to API
```
{
  "event_id": 1,
  "include_unsubmitted": false,
  "question_ids": [3, 5]
}
```

## Example JSON Response from API
```
[
  {
    "response_id": 1,
    "user_title": "Mr",
    "firstname": "Jimmy",
    "lastname": "Fallon",
    "start_date": "2020-06-01 [...]",    // ISO8601 format
    "is_submitted": true,
    "is_withdrawn": false,
    "submitted_date": "2020-06-03 [...]",  // ISO8601 format
    "language": "en",
    "reviewers": [
       {"reviewer_id": 4, "reviewer_name": "Ms Hilary Clinton", "review_response_id": 45},
       {"reviewer_id": 63, "reviewer_name": "Mr Barack Obama", "review_response_id": null},
    ],
    "answers": [
       {"question_id": 3, "value": "Hello world", "type": "short-text", "options": null, "headline": "Question 3"},
       {"question_id": 5, "value": "harry-potter", "multi-choice". "options": [{"label": "Harry Potter", "value": "harry-potter"}, {"label": "X-men", "value": "x-men"}], "headline": "Question 5"}
    ]
  },
  {
    "response_id": 2,
    "user_title": "Ms",
    "firstname": "Halle",
    "lastname": "Berry",
    "start_date": "2020-06-01 [...]",    // ISO8601 format
    "is_submitted": true,
    "is_withdrawn": false,
    "submitted_date": "2020-06-03 [...]",  // ISO8601 format
    "language": "en",
    "reviewers": [
       {"reviewer_id": 63, "reviewer_name": "Mr Barack Obama", "review_response_id": 23},
       null
    ],
    "answers": [
       {"question_id": 3, "value": "Hello world", "type": "short-text", "options": null, "headline": "Question 3"},
       {"question_id": 5, "value": "x-men", "multi-choice". "options": [{"label": "Harry Potter", "value": "harry-potter"}, {"label": "X-men", "value": "x-men"}], "headline": "Question 5"}
    ]
  }
]
```

greeffjean:
Hi, Resolves #779. Response List Component

As per instructions and communications here is issue779 resolved, tests to follow shortly. 
Please Let me know if the UI or naming conventions is not acceptable, I think your going to like the UI!


avishkar58:
This does look really nice at first glance, I especially like the animations! :) Will go through it in more detail later tonight. Thanks @greeffjean !

greeffjean:
Great Im glad

avishkar58:
Hey Jean,
This looks great, I can see you put a lot of effort into this, much appreciated! :) Some general feedback:
- Please show the table by default (with no questions selected), so you don't have to press "Add Table" when you land on the page
- Can you combine the user_title, firstname, lastname columns like you do with Reviewer1 and Reviewer2 to save some space?
- I noticed that if you click on the label of a checkbox in the Question dropdown, it selects/unselects the "Include un-submitted" rather than the question's checkbox
- Please change the "Add Table" button to an "Update" button that's only visible when the question selector is visible and maybe move it underneath the questions, at the moment I think its a bit unintuitive to click on "Add Table" which is above the question list (although the fact that it changes to green does help a lot)
- Some application forms can have quite lot of questions, for example we just added a form with 40 questions. I like the UX you've designed here for select questions, but wondering if it will still work for an order of magnitude more questions?
- After adding some questions, the table width expands outside the bounds of the page (see screenshot), can you constrain the width to the page width with a horizontal scrollbar if necessary?
- For the datetime fields, it displays as "2020-06-01 [...]", can you show the time please?
- The headings "Answer1", "Answer2" etc should be the headline of the corresponding question. So for example it might be "First Question", "Third Question" 
- The question selection doesn't seem to be working quite right - if I select 1 question i see Answer1 in the table, if I select 2 then Answer1 and Answer2, which makes sense, but if I select all 5 questions I still only see Answer1 and Answer2 in the table? 

I'll leave some more comments in the code once I've had a chance to go through it in detail

avishkar58:
Screenshot 2020-09-22 at 22 19 04

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+        function addReviewCol(row, newAnswers) {
+            return row.answers = newAnswers
+        }
+
+
+        fetchResponse().then(response => {
+            let handleAnwsers = [];
+            let handleReviews = [];
+            // Handle Anwsers and Reviews
+            response.map(val => {
```
avishkar58, line 84: Since your lambda function has side-effects and doesn't return anything, you should use `forEach` instead of `map`

avishkar58 (changes requested):
Still busy with this review, the handleAnswers and handleReviews doesn't quite make sense to me so need to spend a bit more time understanding the logic

avishkar58:
> The headings "Answer1", "Answer2" etc should be the headline of the corresponding question. So for example it might be "First Question", "Third Question"

We can make this easier by including the headline in answers like we do with the type and options. I've updated the JSON example in the description

> The question selection doesn't seem to be working quite right - if I select 1 question i see Answer1 in the table, if I select 2 then Answer1 and Answer2, which makes sense, but if I select all 5 questions I still only see Answer1 and Answer2 in the table?

Nevermind, I see it's because there's only 2 answers in the JSON :)

greeffjean:
> Still busy with this review, the handleAnswers and handleReviews doesn't quite make sense to me so need to spend a bit more time understanding the logic

This is just variables that store the extracted data I need from JSON objects like the reviews and answers keyes, since they have alot of stuff in them we don't want displayed, then I pass the variables to functions which create new entries in the JSON response with this variable data as the new value and at the end of its parent function I delete the old entries that are not needed anymore like the entire answer object for example, if Im not horribly mistaken the adjustments made to the response data-JSON is not actually changing the original JSON in the cloud / database.

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+        this.toggleList(true)
+
+        const { selected } = this.state;
+
+        // add Answers as new column 
+        function addAnswersCol(row, newAnswers) {
+            return row.answers = newAnswers
+        }
+        // add Reviews as new column 
+        function addReviewCol(row, newAnswers) {
```
avishkar58, line 75: This function does exactly the same thing as `addAnswersCol` ?

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+                        }
+                    }
+                })
+
+                // extract only the reviewers name
+                val.reviewers.map(review => {
+                    review ? handleReviews.push(review.reviewer_name) : handleReviews.push("")
+                })
+                // envoke and store new columns for Reviews and Answers
+                let addAnwsers = addAnswersCol(val, handleAnwsers);
```
avishkar58, line 116: Please fix spelling `addAnswers`

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+                let addReviews = addReviewCol(val, handleReviews);
+
+                // insert anwsers values as columns
+                if (handleAnwsers.length) {
+                    addAnwsers.map((answer, index) => {
```
avishkar58, line 121: This `map` should also be a `forEach` - same for the one below

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+                    })
+                }
+                // insert new reviews values as columns
+                addReviews.map((review, index) => {
+                    let num = (index) + (1);
+                    let key = "Review" + num;
+                    val[key] = review
+                })
+                // delete original review and answer rows as they don't need to be displayed with all their data
+                handleAnwsers = [];
```
avishkar58, line 134: It's generally good practice to keep a variable's scope as minimal as possible as it improves the code's readability (see [here](https://refactoring.com/catalog/reduceScopeOfVariable.html) for more information). So here it would be better to declare `handleAnswers` and `handleReviews` just after line 84 and then you won't need to reset here.

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+                if (colItem.includes('Review')) {
+                    return 180
+                }
+                else {
+                    return 100
+                }
+            }
+
+            let col = Object.keys(this.state.responseTable[0]);
+            col.map(val => {
```
avishkar58, line 175: The "correct" use of map here would be:
`colFormat = col.map(val => ({id: val, Header: val, accessor: val, style: colStyle, width: widthCalc(val)}))`

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+                responseTable: response,
+                addBtn: { background: "hsl(211deg 100% 50%)" }
+            }
+            )
+
+        })
+    }
+
+
+    generateCol() {
```
avishkar58, line 150: This looks like it should be `generalCols` because it's generating a list of columns and not a single one?

webapp/src/pages/ResponseList/components/ResponseListForm.js
```diff
@@ -0,0 +1,251 @@
...
+        let colFormat = [];
+        let colStyle = {
```
avishkar58, line 152: Please put styles in CSS. I think you can pass className to ReactTable to set the class of the table cells and then you don't have to duplicate the style

avishkar58 (changes requested):

greeffjean:
Hi Avishkar.

I made all the changes you requested: 
Please consider the following: 

- The UX should do fine with many questions, looks good to me when I tested.
- The scrollbar was always working, one simply clicks on the table and adds uses the arrows to engage the scroll, otherwise vertical scroll works as normal ( I found no straight forward solutions online for displaying the scrollbar all the time, if this is required I can spend more time making that happen )
- Im not sure weather I did the right thing with the formdate time, before there was just [...] so I added mock time for display.

Thank you for the pointers, Im really enjoying and appreciating the info.

avishkar58 (changes requested):
Hi Jean,

The changes looks great, thanks! :) Almost there, there's just one more issue which is that the answers in the table are still not being displayed correctly. It's grouping up the answers to different questions in the same column, which will be very confusing for the user. If there are 3 different questions selected (a question is different if it has a different ID) then there needs to be 3 columns. For example, if I select "First Question", "Second Question" and "Third Question", then I should see 3 columns with exactly those headings. Suppose the ID of "First Question" is 1, then I should *only* see answers with a question_id of 1 in that column. If a response doesn't have an answer with a question_id of 1, then that row and column should be blank.

greeffjean:
App.js has a conflict I don't recognise as my contribution, perhaps just check that out?

avishkar58:
Looks good, thanks @greeffjean !

avishkar58 (approved):

PR #800, opened 2020-10-21
Add tags to response page

greeffjean:
Tags added, please review


avishkar58:
Hi @greeffjean - please merge develop into here, resolve any conflicts and then I'll review :)

avishkar58:
This seems to be exactly the same as PR #795 ?

greeffjean:
> This seems to be exactly the same as PR #795 ?

Correct, there are a couple on issues that  rely on each other so what I have done to carry the progress over to a new issue is have made merges, I think where I am confusing you is: I updated an component from a older issue in a newer issue 'I can see how that is not best practice', and then merged the older issues with a newer one afterwards to retrieve the changes

PR #834, opened 2020-11-19
Iss773

greeffjean:
Please advise me on formatting, currently the API response contains preset set dates, when I submit the form with dates unchanged I get this error, please see function commented out in `handleSubmit()` to see my approach? 

`time data '2021-07-06T00:00:00' does not match format '%Y-%m-%dT%H:%M:%SZ'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ It will take a fair bit of time the first time you do it, subsequently it will b
 
 7. The first time you run the app, you may need to run the **migrations** to ensure that all the tables are created in the database. While the app is running (after following the previous step), run the following in **another terminal/command prompt**: ```docker-compose run web python ./api/run.py db upgrade --directory api/migrations```
 
+
 8. Read the respective docs for [backend](./api/README.md) and [frontend](./webapp/README.md). 
 ## Troubleshooting
 Common errors you may get when running the ```docker-compose build``` or ```docker-compose up``` commands.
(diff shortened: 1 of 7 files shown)


avishkar58:
Hi Jean,

I just pushed a change to the API into this branch which ensures that the date format from the GET method is the same as expected by the POST and PUSH methods (so it will return and accept dates that look like  `2020-01-01T00:00:00Z`). It seems to work now, but there are still some other errors. Also please note that the following new fields have been also added to Event since this component was created (and they're all required, so the API will throw a "bad request" error if you don't provide them, I added them to the `update` method of the service, but you'll need to create form inputs for them please and also include them in the service's create method.)
- `event_type` - this can be one of "event", "award" or "call" - please add a dropdown box to choose this
- `travel_grant` - this is a checkbox, please add some text along the lines of "Does this event provide travel grants for participants" 
- `miniconf_url`: string, please add a text description along the lines of "If this is a virtual MiniConf event, enter the URL to the miniconf site"

avishkar58:
Hi Jean,
I forgot to mention in the issue description, but can you also please make sure the users can select the *time* in addition to the date, and also add a note somewhere to say that the times are in UTC. This is because we have users from all over the world and they often like to set deadlines in their local timezone (so for example a deadline of midnight PST would require them to select 8am the next day in UTC) - thanks!

greeffjean:
Please advise redirect behaviour once API call is successful.

At the moment the screen just goes gray.

greeffjean:
Ive discovered an issue with `updateEventDetails()`, I am aware of it and working on a solution

PR #857, opened 2021-01-08
visual warning for users if password don't match

NikolaSiplakova:


avishkar58 (approved):
Fantastic, thank you @NikolaSiplakova !
By the way, I've added you as a contributor on the repo now with write access, so you can create branches on here instead of forking - the advantage is that our automated tests will run on these branches and we'll get that feedback on the PRs.

PR #858, opened 2021-01-20
Basic components for application form builder #749

### LINKED ISSUE #749: Create basic page and components for Application Form Builder
## Background

## Requirements
- Create a new application form builder page 
  - Add a route to it in [EventHome.js](https://github.com/deep-learning-indaba/Baobab/blob/7dcc44b008f4291854079db1e3657ed6a10dfadd/webapp/src/pages/eventHome/EventHome.js#L127) of the form `${match.path}/applicationform` - (this makes it an _event specific route_)
  - Add a link to this route in the Event Admin menu in [App.js](https://github.com/deep-learning-indaba/Baobab/blob/7dcc44b008f4291854079db1e3657ed6a10dfadd/webapp/src/App.js#L120) (it should therefore only be visible to users with the “admin” event role)
- Create components to represent an ApplicationForm, Section and Question
- Add a checkbox to allow the user to select whether the application form accepts nominations or not. To make it clear to the user what this does, the checkbox text should say “Allow candidates to nominate others using this application form” and add a further description saying “Users will be able to submit multiple nominations, including for themselves. If this option is unchecked, a candidate can only apply for themselves”. The value of this checkbox should correspond to the `nominations` field on the application form's JSON representation.
- Query the [event details service](https://github.com/deep-learning-indaba/Baobab/blob/7dcc44b008f4291854079db1e3657ed6a10dfadd/webapp/src/services/events/events.service.js#L14) to get more information about the event (more than what's in the event object passed as a prop to the component) 
  - Display the application form's open and close date, which can be obtained from the `event` object's `application_open` and `application_close` dates
- Maintain a list of Section in the ApplicationForm state
- Maintain a list of Question in the Section state
- Render a card for each Section (add a few of the fields) 
- Render a card for each Question (add a few of the fields) 
- You can use the example JSON below to populate the sections and questions
- Experiment with the UI, making it look decent. 

## Definition of Done
- Code has been developed and tested, following the code style and conventions used by other pages and components
  - Note: any "hard-coded" text in the UI must be added to the two `translation.json` files for [English](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/public/locales/en/translation.json) and [French](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/public/locales/fr/translation.json) and accessed using the component's i18n `t` property. This is easier than it sounds, see other pages for an example of how to do it. Use Google Translate to translate English to French
- A test has been created that checks that the page renders
- A PR to merge the changes into develop has been reviewed, comments addressed and merged.

# Example JSON
```
{
  "id": 2,
  "event_id": 2, 
  "is_open": true, 
  "nominations": false, 
  "sections": [
    {
      "name": {
         "en": "Section 1",
         "fr": "Section 1"
      },
      "description": {
         "en": "Description of the section",
         "fr": "Description de la section"
      },
      "order": 1,
      "depends_on_question_id": null,
      "show_for_values": null,
      "key": "candidate_information",
      "questions": [
        {
          "validation_regex": {
             "en": null, 
             "fr": null
          },
          "options": {
              "en": [
                ],
              "fr": [
                  {"value": "undergrad", "label": "un étudiant de premier cycle"}, 
                  {"value": "masters", "label": "un étudiant en maîtrise"}, 
                ],
          },
          "description": {
              "en": null, 
              "fr": null
          },
          "headline": {
              "en": "Question 1", 
              "fr": "Question 1"
          }
          "placeholder": {
              "en": "Select an Option...",
              "fr": "Sélectionnez une option"
          } 
          "is_required": true, 
          "type": "multi-choice", 
          "id": 20, 
          "validation_text": {
              "en": null,
              "fr": null
          } 
          "order": 1,
          "depends_on_question_id": 2,
          "key": null,
          "show_for_values": {
              "en": ["the-matrix", "terminator"],
              "fr": ["the-matrix", "terminator"]
          }
        }
      ], 
 
    }
  ]
}
```

Malaba6:
Resolves #749


webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+          type: null,
+          options: [],
+          value: '',
+          label: '',
+          required: false
+        }
+      ]
+    }]);
+  }
+  const addQuestion = (id) => {
```
avishkar58, line 70: Nit: can you call this `sectionId` so that it's clear from the function signature that this is referring to the section and not the question

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+    let newSections = sections.map(e => {
+      if(e.id === id) {
+        return {...e, questions: [...e.questions, {
+          id: `${Math.random()}`,
```
avishkar58, line 74: (Note: The backend expects the IDs to be integers not strings, and for new questions to have no id property. )

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+          label: '',
+          required: false
+        }]}
+      }
+      return e;
+    });
+    setSections(newSections);
+  } 
+  const dateFormat = (date) => {
+    return new Date(date).toLocaleDateString('en-US', {
```
avishkar58, line 90: Can we use en-GB for the date formats please :)

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+      hour: '2-digit'
+    })
+  }
+
+  const TopBar = () => {
+    return (
+      <div className="top-bar">
+        <div className="icon-title">
+          <img src={icon} alt="form" className="icon" />
+          <span className="title">{t("Form title")}</span>
```
avishkar58, line 104: Rather say "Application Form" instead of "Form title"

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+      </div>
+    );
+  }
+  const FormSection = () => {
+    const {loading, event: evnt, error } = event;
+ 
+    if (loading) {
+      return (
+        <div>
+          <i className="fa fa-spinner fa-lg" aria-hidden="true"></i>
```
avishkar58, line 116: FYI, we have a [loading component](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/Loading.js) which you can use here instead

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+        </div>
+      )
+    }
+    if (error) {
+      return (
+        <div>{error}</div>
```
avishkar58, line 122: Please style this as a bootstrap danger alert for consistency with other pages

webapp/src/pages/createForm/components/Question.js
```diff
@@ -0,0 +1,265 @@
...
+      </div>
+      {t(props.label)}
+    </div>
+    return {
+      label,
+      value
+    }
+  }
+
+  const options = [
```
avishkar58, line 64: Note these labels also need to be translated, you can probably just wrap `value` in `t(value)` in your option function though :)

webapp/src/pages/createForm/components/Question.js
```diff
@@ -0,0 +1,265 @@
...
+                   backgroundColor: state.isFocused && "lightgray",
+                   color: state.isFocused && "#fff"
+                })
+              }}
+              menuPlacement="auto"
+            />
+          </div>
+          {withPlaceHolder.includes(input.type && input.type.value) && (
+            <input
+            name="section-desc"
```
avishkar58, line 161: Why is this called "section-desc" ?

webapp/src/pages/createForm/components/Question.js
```diff
@@ -0,0 +1,265 @@
+import React, { useState } from 'react';
+import { default as ReactSelect } from "react-select";
+
+const Section = ({
```
avishkar58, line 4: Shouldn't this be `Question`?

webapp/src/pages/createForm/components/Section.js
```diff
@@ -0,0 +1,100 @@
...
+    setInput({...input, questions: inpt})
+  }
+  return (
+    <>
+      <div
+        className="section-wrapper"
+        // onBlur={updateSections}
+      >
+        <div className="section-number">
+          <label>{t('Section') + ` ${num} ` + t('of') + ` ${sections.length}`  }</label>
```
avishkar58, line 32: FYI: No need to change this, but for future reference the [`Trans` component](https://react.i18next.com/#what-does-my-code-look-like) allows you to avoid breaking up the strings like this.

webapp/src/pages/createForm/components/Section.js
```diff
@@ -0,0 +1,100 @@
...
+              aria-expanded="false"
+            >
+              <i className='fa fa-ellipsis-v fa-lg fa-dropdown'></i>
+            </button>
+            <div className="dropdown-menu" aria-labelledby="title-desc-toggle">
+              <button
+                className="delete-section"
+                onClick={() => addSection()}
+              >
+                {t("Add Section")}
```
avishkar58, line 57: I think it's a bit unintuitive that the "Add Section" action is inside a Section's menu... It probably makes more sense to have add section be a button on the Form.

webapp/src/pages/createForm/components/Section.js
```diff
@@ -0,0 +1,100 @@
...
+                {t("Delete Section")}
+              </button>
+              <button className="delete-section">
+                {t("Duplicate Section")}
+              </button>
+              <button
+                className="delete-section"
+                onClick={() => addQuestion(inputs.id)}
+              >
+                {t("Add Question")}
```
avishkar58, line 69: Would be nice to make the Add Question button more prominent (eg as a button underneath the questions in a section) to reduce the number of clicks for the user who will probably be adding lots of questions.

webapp/src/pages/createForm/components/Question.js
```diff
@@ -0,0 +1,265 @@
...
+  const withOptions = ['multi-choice', 'multi-checkbox'];
+  return (
+    <>
+      <div
+        className="section-wrapper"
+        onBlur={updateQuestions}
+      >
+        <div className="headline-description">
+          <div className="question-header">
+            <input
```
avishkar58, line 126: Please add a placeholder, it's not obvious at the moment that the user should type something here :)

avishkar58 (changes requested):
Thank you @Malaba6 !! This is FANTASTIC, I really really like the look of the UI and your code is very clean and an absolute pleasure to review :) I've left some comments for fairly minor changes and suggestions for improvement.

webapp/src/pages/createForm/components/ApplicationForm.js
```diff
@@ -0,0 +1,181 @@
...
+    })
+  }, []);
+  const handleCheckChanged = (e) => {
+    setNominate(e.target.checked);
+  }
+  const handleSection = (input) => {
+    setSections(input)
+  }
+  const addSection = () => {
+    const id = sections.length + 1;
```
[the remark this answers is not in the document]
Malaba6: > This is fine for creating a form from a blank slate, but will cause issues when editing an existing form where the IDs for existing sections come from the backend - in that case this code might cause an ID collision. The collision could be solved easily by using the maximum ID + 1, however that still leaves another problem, which is that the backend for the PUT request relies on the IDs to know whether to update or insert a section (update if ID is populated, insert if not). I know you rely on this ID later for adding questions, so either you'll have to add a surrogate ID property to the sections to use just in the front-end, or I don't know if there's a way to push the question state entirely into the Section component and recover it easily in the ApplicationForm component when you need it? (there isn't as far as I understand, but I don't know if the new Hooks stuff adds anything like this?)

Noted, Thanks for the heads up!
Yes these IDs will be used only for frontend purpose. In case of editing a existing form I will be handling this issue by renaming some properties or using aliases.

webapp/src/pages/createForm/components/ApplicationForm.js
Malaba6: > (Note: The backend expects the IDs to be integers not strings, and for new questions to have no id property. )

Yes I noticed this, These IDs will not be sent to the backend. I need them on the front end when adding new options for a questions.

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -0,0 +1,286 @@
...
+                        type='text'
+                        placeholder={t('Value')}
+                        onChange={handleChange('value')}
+                        className='option-inputs'
+                      />
+                    </td>
+                    <td className='options-row'>
+                      <input
+                        type='text'
+                        placeholder='Label'
```
avishkar58, line 205: Wrap this string 'Label' in `t()`

avishkar58 (changes requested):
Excellent thank you @Malaba6 ! 
Just two things: 
1. Please remove the migration file
2. I noticed that if I add a question, type something in the headline and then press the "add question" button again, it doesn't register and I have to press it twice to get it to add a new question. If I don't type a headline before pressing the add question it works though.

Will merge once those two points have been addressed - thanks!

webapp/src/context/ContextProvider.js
```diff
@@ -0,0 +1,17 @@
+import React, { useState } from 'react';
```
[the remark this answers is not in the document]
Malaba6: ContextProvider there provides access to local context to all the child components that are under the Context.
In other words, all the child components to the ContextProvider will have access to the local store of this context

avishkar58 (approved):
Excellent work, thanks Eric!

PR #860, opened 2021-01-28
Add space between cookie text and privacy policy

gausam:
This is how it currently looks without the space:

image


avishkar58 (approved):
Awesome thanks!

PR #867, opened 2021-02-13
Fix failing integration test (signup spec)

### LINKED ISSUE #852: Fix failing integration test
Fix the following two integration tests in [signup.spec.js](https://github.com/deep-learning-indaba/Baobab/blob/develop/integration-tests/cypress/integration/signup.spec.js) - it seems to be an issue with the test because the actual signup page is working fine. 
- Signup form cannot be submitted in if you havent accepted the policy or if you are underage.
- Singup form can be submitted if you fill in everything correctly.

gausam:
It created an actual user account in the test, which on the next run would of course be unavailable for re-registration with the same credentials.

Closes #852

### DIFF
--- a/integration-tests/cypress/integration/signup.spec.js
+++ b/integration-tests/cypress/integration/signup.spec.js
@@ -46,11 +46,12 @@ describe("Sign up", function() {
   });
 
   // Disabling these tests because they're failing. See issue #852
-  /*
+
   it("Signup form cannot be submitted in if you havent accepted the policy or if you are underage.", function() {
     let user = testUser();
-    cy.visit("/createAccount");
     cy.wait(2000);
+    cy.visit("/");
+    cy.get("#nav-signup").click();
     cy.get("#title")
       .find("input")
       .click({ force: true })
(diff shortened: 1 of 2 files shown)


integration-tests/cypress/integration/signup.spec.js
```diff
@@ -68,10 +69,12 @@ describe("Sign up", function() {
     cy.get("#btn-signup-confirm").should("be.disabled");
   });
 
-  it("Singup form can be submitted if you fill in everything correctly.", function() {
+  it("Signup form can be submitted if you fill in everything correctly.", function() {
     let user = testUser();
-    cy.visit("/createAccount");
```
gausam, line 73: Router behavior will redirect away from `/createAccount` if the user is not authenticated.

gausam:
> Timed out retrying: Expected to find element: '#title', but never found it.
Hmm. Seems it doesn't run fast enough on CircleCI 🤔

integration-tests/cypress/integration/signup.spec.js
avishkar58: Really? That seems like a bug then... /createAccount should be accessible if the user is not authenticated...

gausam: Yup, that's the behavior I get on my end at least.
Should I fix the bug as well?

avishkar58: Ah I get that behaviour too... yes please fix that bug as well, it makes sense for `/createAccount` to be accessible by a non-authenticated user

avishkar58:
Hi @gausam - is this ready to be merged? It seems to be passing on CircleCI now? :D

gausam:
@avishkar58 It's ready! 🚀

avishkar58 (approved):

Issue #864, opened 2021-02-07
Create ReviewResponseDetailList API


avishkar58:
Create a new ReviewResponseDetailList API in [reviews/api.py](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/reviews/api.py) which does the following

- Accept `event_id` (int) and `language` (str) parameters
- Ensure that the logged in user is an event admin for the event
- Query the database for all review responses associated with the given event using a new method on the [review repository](https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/reviews/repository.py), as well as some details about the response, the applicant and the reviewer. These details include:
  - The user who submitted the response (called "response_user" in the JSON below)
  - The user representing the reviewer (called "reviewer_user" in the JSON below)
  - Questions with a key of "review-identifier" and the corresponding Answers from the response 
  - The review questions and review scores associated with the review response
- Return a JSON response to the client containing all of the review responses which looks like the example below

## Example JSON Response
Notes:
- `response_identifiers` are the headlines (translated) and answers to all questions from the relevant application form that have a `key` value of "review-identifier"
- The scores should be filtered to questions of type "multi-choice", "long-text" and "short-text"

```
[
  {  
    "review_response_id": 1,
    "reviewer_user_title": "Mr",  
    "reviewer_user_firstname": "Joe",
    "reviewer_user_lastname": "Soap",
    "response_id": 3,
    "response_user_title": "Ms",
    "response_user_firstname": "Jane",
    "response_user_lastname": "Bloggs",
    "identifiers": [    
        {
            "headline": "Organisation",     // From question_translation
            "value": "Pets R Us"                // From answer
        },
        {
            "headline": "Country",
            "value": "Nigeria"
        },
    ],
    "scores": [
        {
            "headline": "Ethical Considerations",   // From review_question_translation
            "description": "How ethical is the candidate's proposal from 1 to 5?",   // From review_question_translation
            "type": "short-text",   // From review_question
            "score": 4,   // From review_score
            "weight": 1    // From review_question
        },
        {
            "headline": null,
            "description": "Comments for the candidate",
            "type": "long-text",
            "score": "This is a very good proposal",
            "weight": 0
        },
        {
            "headline": null,
            "description": "What is your overall rating?",
            "type": "multi-choice",
            "score": 5,
            "weight": 2
        }
    ],
    "total": 14
  },

  {  
    "review_response_id": 2
    "reviewer_user_title": "Ms",  
[...]

PR #868, opened 2021-02-14
Feature/review response detail list

amritpurshotam:
Resolves: #864 

Note. Make sure to hide whitespace changes when reviewing. The merge seemed to think the whole api.py file changed.


api/app/reviews/api.py
```diff
@@ -1,580 +1,650 @@
...
+            question_translation = answer.question.get_translation('en')
+            LOGGER.warn('Could not find {} translation for question id {}'.format(language, answer.question.id))
+        
+        return {
+            'headline': question_translation.headline,
+            'value': answer.value_display
+        }
+    
+    @staticmethod
+    def _serialise_scores(review_score, language):
```
avishkar58, line 596: should be `_serialise_score` (not plural) :)

api/app/reviews/repository.py
```diff
@@ -257,6 +257,15 @@ def delete_response_reviewer(response_id, reviewer_user_id):
         db.session.query(ResponseReviewer).filter_by(response_id=response_id, reviewer_user_id=reviewer_user_id).delete()
         db.session.commit()
 
+    @staticmethod
+    def get_all_review_responses_by_event(event_id):
+        return (
+            db.session.query(ReviewResponse)
+            .join(ReviewForm, ReviewForm.id == ReviewResponse.review_form_id)
+            .join(ApplicationForm, ApplicationForm.id == ReviewForm.application_form_id)
+            .filter_by(event_id=event_id)
```
avishkar58, line 266: Is the `.all()` not needed? I guess not since the tests are working

avishkar58:
Excellent work as usual, thanks @amritpurshotam ! Great to have you back! :D

avishkar58 (approved):

api/app/reviews/api.py
amritpurshotam: Ah yes good catch!. Fixed now.

api/app/reviews/repository.py
amritpurshotam: Good catch. That is strange how the tests are still passing. I've added the `.all()` now just to be safe.

PR #869, opened 2021-02-16
Add rename parameter to file download URLs

### LINKED ISSUE #862: Update FileUpload on front-end
Depends on #861 

## Motivation
At the moment, our files are uploaded to cloud storage using a GUID and without a file extension. This means that the filename the users originally had for their files are lost. This causes problems when downloading the file again especially for Windows users because Windows relies on file extensions to determine the file type, unlike Mac/Linux. The backend FileUpload API has been updated to accept a `rename` parameter and will set the filename to this before returning it to the client. If we keep track of the filename we want, the front-end can pass this to the File api when downloading a file. 

## Requirements
Update the [FileUpload](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/FileUpload.js) and [FormFileUpload](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/form/FormFileUpload.js) components such that when FileUpload calls props.onChange, it passes not just the GUID of the file that was uploaded, but a JSON object, converted to a string, that looks like this:
```
{
  "filename": "<UPLOADED_FILE_GUID>",
  "rename": "<ORIGINAL_FILE_NAME>"
}
```
The component should also be able to parse and load this JSON into its state if it's initialised with such a value.
Have a look at the FormMultiFile and MultiFileUpload components as an example of a component that uses complex state
Test this by ensuring that whenever a file is uploaded using the application form, that the Answer value is a string containing this JSON and not just the file GUID

Next, update the [AnswerValue component](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/answerValue/AnswerValue.js) so that the "file" type is compatible with the new JSON encoded file representation and also backwards compatible with the old GUID-only representation. 
- When the value is a string-encoded JSON object, the component displays the "rename" value instead of "Uploaded file" and the URL link is `{baseUrl + "/api/v1/file?filename=" + filename + "&rename=" + rename}` where `filename` and `rename` were extracted from the JSON. Remove the note about changing the filename in this case.
- When the value is just a single string (ie the string value doesn't start with "{" and end with "}"), then keep the existing behaviour.

gausam:
Closes #862

### DIFF
--- a/webapp/src/components/FileUpload.js
+++ b/webapp/src/components/FileUpload.js
@@ -28,7 +28,13 @@ class FileUploadComponent extends Component {
           })
           .then(response => {
             if (response.fileId && this.props.onChange) {
-              this.props.onChange(this.props.id, response.fileId);
+              this.props.onChange(
+                this.props.id,
+                JSON.stringify({
+                  filename: response.fileId,
+                  rename: file.name
+                })
+              );
             }
             this.setState({
               uploaded: response.fileId !== "",
--- a/webapp/src/components/ViewFile.js
+++ b/webapp/src/components/ViewFile.js
@@ -1,8 +1,7 @@
 import React, { Component } from "react";
 import { withRouter } from "react-router";
 import { withTranslation } from 'react-i18next';
-
-const baseUrl = process.env.REACT_APP_API_URL;
+import { getDownloadURL } from '../utils/files';
 
 class ViewFile extends Component {
   
@@ -16,7 +15,7 @@ class ViewFile extends Component {
     componentDidMount() {
         if (this.props.match && this.props.match.params) {
             const filename = this.props.match.params.filename;
-            const fileLocation = baseUrl + "/api/v1/file?filename=" + filename;
+            const fileLocation = getDownloadURL(filename);
             this.setState({
                 fileLocation: fileLocation
             });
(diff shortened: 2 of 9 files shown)


gausam:
@avishkar58 Will these have to be updated as well?

- https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/answerValue/MultiFileValue.js#L10
- https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/pages/ResponsePage/ResponsePage.js#L161
- https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/pages/ResponsePage/ResponsePage.js#L168

avishkar58:
> @avishkar58 Will these have to be updated as well?
> 
> * https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/components/answerValue/MultiFileValue.js#L10
> * https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/pages/ResponsePage/ResponsePage.js#L161
> * https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/pages/ResponsePage/ResponsePage.js#L168

Ah yes - good catch, those need to be updated too please! 
This PR looks great, so once you've updated those we can merge - thank you!

gausam:
Will do!

avishkar58:
Hi @gausam - please merge develop into your branches to fix the tests on CircleCI - thanks!

avishkar58 (approved):

gausam:
Done!

PR #870, opened 2021-02-22
Upgraded to Python 3

avishkar58:
Finally upgrading to Python 3! 
Code converted with 2to3, all requirements upgraded to latest versions possible and tested locally (only Werkzeug had an incompatibility and needed to remain on a previous version)


amritpurshotam:
This is great, nice one! There's a lot to get through here so will have a look properly tomorrow

api/requirements.txt
```diff
@@ -1,42 +1,42 @@
...
-Flask-Admin==1.5.3
-Flask-Bcrypt==0.7.1
-Flask-Failsafe==0.2
-Flask-Migrate==2.3.0
-Flask-RESTful==0.3.7
-Flask-Redis==0.3.0
-flask-restplus==0.13.0
-Flask-SQLAlchemy==2.3.2
-Flask-Script==2.0.6
+Flask
```
amritpurshotam, line 1: Should we completely unpin the package versions? With the upgrade I think the case can be made to take the latest versions now can also see how that would be quite tedious 🙈

amritpurshotam (approved):
Nicely done! Glad this is finally out the way 😄 And dare I say was this upgrade easy? It seemed a lot harder when it was tried last year

api/requirements.txt
avishkar58: They are completely unpinned now except for Werkzeug which has a version incompatibility with the latest Flask-restful, unless I didn't understand correctly what you mean?

avishkar58:
Thanks for the review @amritpurshotam ! I know it was a large one! It was actually easier than I thought it would be, 2to3 did most of the heavy lifting and there was only a handful of changes I had to make manually mostly to do with bytes vs unicode incompatibilities :)

PR #879, opened 2021-02-27
Add logic to section and question fields

### LINKED ISSUE #850: Add logic to section and question fields
## Background

## Dependencies
Depends on the fields created in #849 

## Requirements
Add the following logic to the section and question fields on the application form builder.

#### `show_for_values`
If the question referred to in `depends_on_question_id` has type "multi-choice":
Allow the user to select one or more of the options from the dependency question. (For example, if the user selects the question "What is your favourite movie" in the `depends_on_question_id` field, this field should display the list of the options that the "What is your favourite movie" question has. The user should be able to check some of these options, for example "The Matrix", "Terminator". The final value for the field is a JSON array of these values, for example `["The Matrix", "Terminator"]`). See below the description of options below and make sure that the values are used in the field value and the labels are used in the display.

If the question referred to in `depends_on_question_id` has type "single-choice":
The user can select "True" or "False" and this becomes the value of the field. 

#### `show_for_values`
This logic is the same as described in the Section above

#### `validation_regex`
The majority of users will not be able to write regex themselves, so the UI should make it easy to enter the most common type of regex validation rule, *word limits*. By default, the validation rule should be presented to the user in the form of a minimum (MIN) and maximum (MAX) word limit input box ("friendly mode"). Either minimum or maximum can be blank. They should be translated into the following regex:
`^\s*(\S+(\s+|$)){MIN,MAX}$`  (if the minimum is blank, set MIN to 0, if the maximum is blank, leave out MAX). If both the minimum and maximum are blank, the validation_regex should be set to null. 

Add an advanced checkbox which removes the minimum and maximum inputs and allows the user to edit the regex directly. 

One complication here is that when editing an existing application form, if the regex is of the form `^\s*(\S+(\s+|$)){MIN,MAX}$` (or null) then the MIN and MAX should be extracted and displayed in the friendly format. If the validation_regex does not match this form, it should be displayed in "advanced mode" 

#### `validation_text`
This depends on the value of the `validation_regex` field. If the validation for this question is in "advanced mode", simply display an input box that allows the user to edit this field directly. If the validation is in "friendly mode", the validation_text property should also be set to “You must enter {MIN} to {MAX} words” if the minimum and maximum are greater than 0. If the minimum is 0 (assume blank is 0), the validation_text property should be set to “You may enter no more than {MAX}” words. If the maximum is blank but the minimum is greater than 0, the validation_text should be “You must enter at least {MIN} words”. If both the minimum and maximum are blank, the validation_text should be blank. 
**TODO**: What about populating the other languages besides English? Can we use the Google Translate API to do this?

#### `options`
If the question's `type` is "multi-choice" or "multi-checkbox":
Allow the user to enter a list of "value and "label" pairs (using input boxes for value and label, a plus button to add a new row and trash icon to remove a row). The list of value and label pairs should be translated into a JSON representation that looks like the example below to store in the `options` field:
```
[
  {"value": "industry", "label": "Industry Professional"},
]
```
In this case, at least one option in the list is **required**.
(Remember when editing an existing form, the JSON representation should also be mapped to this UI list)
(Remember this is translatable, so this should be replicated for each language)

If the question's type is "reference", prompt the user for a minimum and maximum number of referrals and translate this to/from the following JSON example to store in the options field:
`{"max_num_referral": 3, "min_num_referral": 1}`
In this case, the options is **required**

If the question type is "file" or "multi-file", prompt the user for a list of file extensions and map this to/from a JSON representation that looks like this example:
`{"accept": [".csv", ".xls", ".xlsx"]}`
In this case, the options are not required.

### Deleting a question
If the user deletes a question and the question is referred to as a dependency of another question or section (eg if the user deletes question ID 10 and the `depends_on_question_id` property of one or more other sections or questions contains the value 10), add a note to the confirmation modal dialog that the question is a dependency of other sections/questions. If the user confirms the deletion, remove the deleted ID from any `depends_on_question_id` properties.

Malaba6:
Resolves #850


webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -219,8 +407,27 @@ const Question = forwardRef(({
     }),
   ];
 
+  const validationOptions = [
+    option({
+      value: 'friendly-mode',
+      label: 'Friendly Mode',
```
avishkar58, line 413: Please label this as "Word Limit" instead of "Friendly mode" so that it's clear that they're setting a max and min word limit.
Also, this text needs to be translated please.

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -219,8 +407,27 @@ const Question = forwardRef(({
...
+      t
+    }),
+    option({
+      value: 'advanced-mode',
+      label: 'Advanced Mode',
```
avishkar58, line 418: Also needs to be translated.

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -359,9 +567,61 @@ const Question = forwardRef(({
...
+              <span className='error-label'>{t(errorMessage)}</span>
           </div>
           )}
-          
+          {withReferals.includes(inputs.type && inputs.type.value) && (
+            <div className='referals-wrapper'>
+              <input
+                type='number'
+                min={0}
+                placeholder={t('Enter Minum Number of Referrals')}
```
avishkar58, line 578: Typo: should be "Minimum"

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -359,9 +567,61 @@ const Question = forwardRef(({
...
+                className='referals-input'
+                value={inputs.options.min_num_referral}
+                onChange={handlereferralsChange('min')}
+               />
+              <input
+                type='number'
+                min={0}
+                placeholder={t('Enter Maximum Number of Referrals')}
```
avishkar58, line 586: One issue with using placeholders here instead of labels is that when the user comes to edit an existing form at some point, the placeholder won't be visible and it's not clear what these values refer to. This applies in a few cases, so although it may not look as clean, it's actually better UX to use always-visible labels in cases like this...

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -359,9 +567,61 @@ const Question = forwardRef(({
...
+               />
+              <div className='error-label'>{t(errorMessage)}</div>
+            </div>
+          )}
+          {withExtention.includes(inputs.type && inputs.type.value) && (
+            <div className='extensions-wrapper'>
+              <input
+                type='text'
+                placeholder={
+                  t('Please enter a list of file extensions, each starting with a period and separated with comas. E.g .csv,.xls')}
```
avishkar58, line 599: Same comment as above regarding the placeholder, here the placeholder contains some valuable information. Suppose the user wants to edit an existing form that has ".xls" already entered, they won't know that they can add another separated by commas because this placeholder won't be visible...

avishkar58, line 599: Typo: should be "commas"

webapp/src/pages/createApplicationForm/components/util.js
```diff
@@ -222,3 +233,240 @@ export const handleMove = ({
...
+        <div className='dependency-radio-wrapper'>
+          <input
+            id='single_choice'
+            type='checkbox'
+            className='single-choice-check'
+            checked={inputs.show_for_values}
+            onChange={handleDependencyChange('single')}
+            />
+          <label htmlFor='single_choice'>
+            {inputs.show_for_values ? 'True' : 'False'}
```
avishkar58, line 283: I found the checkbox which toggles between True and False when the dependency question is type "single-choice" to be a bit confusing... Can you rather change this to two radio buttons with "True" and "False" please?

webapp/src/pages/createApplicationForm/components/util.js
```diff
@@ -222,3 +233,240 @@ export const handleMove = ({
...
+      ))}
+    </div>
+  )
+}
+
+export const validationText = (min, max) => {
+  if (max && !min) {
+    return `You may enter no more that ${max} words`
+  } else if (!max && min){
+    return `You must enter atleast ${min} words`
```
avishkar58, line 312: Typo: space between "at least"

avishkar58 (changes requested):
Thank you @Malaba6 for once again delivering such excellent work! 
I've left some fairly small changes and suggestions for improvements in the comments. Two other things I noticed while testing:
1. I see that the list of questions that appears in the "Depends on Question" is the set of questions that are in the current section. This is fine for questions, because generally a question will depend on a previous question in the same section (remember, in the application form, this functionality will show or hide a question depending on the value the user selected in some previous question). Ideally the menu should only show questions that appeared before the current question, but if that's tricky to do, don't worry about it. However, for the **Section** the logic unfortunately doesn't work since we can't hide a section depending on the value of a question inside that very same section. So for the sections, the questions that appear in the "depends on question" menu needs to be questions from all of the previous sections and **not** the current section. Hope this makes sense - let me know if not!
2. I noticed that when I add a validation, and when I first enter a value in the MIN box, there is a weird animation that happens, which is a bit disorienting.

Thank you!

Malaba6:
@avishkar58 done!

avishkar58 (approved):
Thank you @Malaba6 ! The changes look great!
Will merge this for now, but I did notice some minor things when testing that you can please address in your next PR:
1. The "Duplicate Section" option in the Section menu doesn't seem to work (not sure if it was working before)
2. If I add a new section, then choose a dependency question, I get the unexpected animation

PR #880, opened 2021-03-05
Add section and question fields to application #849

### LINKED ISSUE #849: Add section and question fields to application form builder
## Background

## Dependencies
Depends on the page and components created in #749 

## Requirements
Update the components created in #749 to include all relevant fields for sections and questions, including the *translatable* fields. Details on how to handle translatable fields are included later.

### Sections
Add the following fields to the sections (here we list the field names as they appear in the JSON representation, but please use descriptive labels for the UI)
- name (short text, translatable, required)
- description (long text, translatable, not required)
- depends_on_question_id (selection, not required)
  - This should be a dropdown box that lists all the "multi-choice" and "single-choice" questions in the application form. It can be left blank and a user should also be able to clear an existing selection.
- show_for_values (short text)
  - Only show this field if the `depends_on_question_id` is not blank. 
- key (short text)

### Questions
Add the following fields to the questions (here we list the field names as they appear in the JSON representation, but please use descriptive labels for the UI)
- type (dropdown, the options are listed below, required)
- headline (short-text, translatable, not required)
- description (long-text, translatable, not required)
- placeholder (short-text, translatable, not required)
  - Only visible if the `type` is "long-text", "short-text" or "multi-choice"
- options (long text for now, translatable, not required)
  -  Only visible if the `type` is "multi-choice", "multi-checkbox", "file", "multi-file" or "reference"
- validation_regex (short-text, translatable, not required)
  - Only visible if the `type` is "long-text", "short-text", or "markdown"
- validation_text (short-text, translatable, not required)
  - Only visible if validation_regex is visible AND not blank
- is_required (checkbox, required)
  - Add some visual indication when this checkbox is checked to make it clear to the user that the question has been marked as required (for example an asterisk or a border colour)
- key (short-text, not required)
- depends_on_question_id (selection, not required)
  - This should be a dropdown box that lists all the "multi-choice" and "single-choice" questions in the application form. It can be left blank and a user should also be able to clear an existing selection.
- show_for_values (short text)
  - Only show this field if the `depends_on_question_id` is not blank. 
   
### Question Types
The following table lists the question type that should be available in the `type` dropdown. The label is what's displayed to the user in the dropdown and the value is the value stored in the object's field. 

Label | Value
-- | --
Short Answer | short-text
Paragraph Answer | long-text
Paragraph Answer with Formatting | markdown
Single Checkbox | single-choice
Dropdown | multi-choice
Multiple Choices (Checkboxes) | multi-checkbox
File Upload | file
Date | date
Reference Request | reference
Multiple File Uploads | multi-file

### Translatable fields
For all the fields marked as translatable, the user should be able to enter these in all the languages supported by the event. So for example, if the event supports English and French, the user should be able to enter the `name` property of the section in both English and French (There should be an input box corresponding to each language). The JSON representation of this field is then a nested object which has the languages as keys like this example:
```
"name": {
         "en": "Section 1",
         "fr": "Section 1"
      }
```
The languages supported by the event can be determined by querying the [event service](https://github.com/deep-learning-indaba/Baobab/blob/7dcc44b008f4291854079db1e3657ed6a10dfadd/webapp/src/services/events/events.service.js#L14) for the event detail and checking which keys exist in the event's name property (the name property will be a nested object with keys for each language)

Malaba6:
Resolves #849


avishkar58 (approved):
Excellent, thank you @Malaba6 ! I like the way you did the labels, they still look very good and definitely improve the usability.

PR #884, opened 2021-03-08
Application form builder integration

### LINKED ISSUE #851: Application Form Builder Integration
## Background

## Dependencies
Depends on #847 and #850

## Requirements
This issue involves tying up any loose ends remaining after the previous issues have been completed and ensuring that the front-end works correctly with the backend. 

- Create methods in [applicationForm.service.js](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/services/applicationForm/applicationForm.service.js) to interact with the GET, POST and PUT methods of the `applicationformdetail` backend API
  - The GET method will retrieve an existing application form represented similarly to the example JSON below. It will return null if there is no application form for the event (in which case the front-end should be in "new" mode)
  - The POST and PUT methods will accept a JSON representation of an application form similar to the JSON below, note that for the POST method, none of the ids will be populated. For the PUT method, the ids of the application form itself and any existing sections/questions that need to be updated should be populated, and the ids of new sections/questions should be excluded.
- When the main component mounts, query the service's GET method to determine whether there is already an application form for the event. If there is an application form, the page should enter "edit" mode, the sections and questions and their respective fields should be populated and allow the user to edit these. If there is no application form, the page should enter "new" mode and allow the user to create a new application form from scratch. 
- Add a Save button to the Application Form builder page created in previous issue.
  - If this is a new application form, call the appropriate service method to send a POST request to the backend
  - If the user was editing an existing application form, call the appropriate service method to send a PUT request to the backend.
- After saving the application form, redirect the user to the event home page.

# Example JSON
```
{
  "id": 2,
  "event_id": 2, 
  "is_open": true, 
  "nominations": false, 
  "sections": [
    {
      "name": {
         "en": "Section 1",
         "fr": "Section 1"
      },
      "description": {
         "en": "Description of the section",
         "fr": "Description de la section"
      },
      "order": 1,
      "depends_on_question_id": null,
      "show_for_values": null,
      "key": "candidate_information",
      "questions": [
        {
          "validation_regex": {
             "en": null, 
             "fr": null
          },
          "options": {
              "en": [
                ],
              "fr": [
                  {"value": "undergrad", "label": "un étudiant de premier cycle"}, 
                  {"value": "masters", "label": "un étudiant en maîtrise"}, 
                ],
          },
          "description": {
              "en": null, 
              "fr": null
          },
          "headline": {
              "en": "Question 1", 
              "fr": "Question 1"
          }
          "placeholder": {
              "en": "Select an Option...",
              "fr": "Sélectionnez une option"
          } 
          "is_required": true, 
          "type": "multi-choice", 
          "id": 20, 
          "validation_text": {
              "en": null,
              "fr": null
          } 
          "order": 1,
          "depends_on_question_id": 2,
          "key": null,
          "show_for_values": {
              "en": ["the-matrix", "terminator"],
              "fr": ["the-matrix", "terminator"]
          }
        }
      ], 
 
    }
  ]
}
```

Malaba6:
Resolves #851


avishkar58 (changes requested):
Thank you very much Eric, this is looking sooo good!!! I did some initial testing and noticed the following issues that need to be addressed please.
- The description (of both section and question) can be long, and may consist of multiple paragraphs, so should be a textarea rather than an input box. Please see the description of Section 1 on http://localhost:8080/test2021/applicationform as an example.
- Now that it’s integrated, please use the actual languages of the organisation/event rather than the hard-coded English and French
- When editing an existing form where some questions have validation applied to them (ie validation_regex is not null), then the validation box should be shown for those questions by default, instead of being hidden. The same applies to questions and sections that have a key
- When editing a form with questions that already have some validation_text, this is not appearing in the validation text box.
- The form can be quite long and the save button is right at the top - can you add the save button to the side menu instead? 
- Can you add a confirmation if the user tries to navigate away from the page and hasn’t saved their changes?
- The headline and type of questions is required and the name of sections is required. Please prevent the user from saving if these are not filled in. (The backend also requires the description of sections, but if these are not filled in by the user, please send them as blank strings in the JSON instead of null) 
- The page immediately navigates to the home page after saving even if there was an error from the backend - if there is an error it should remain on the page and show the error to the user.

avishkar58:
Will do some more through testing and review the code once these have been addressed. Thanks!

avishkar58:
Hi Eric,

Thank you for the updates. I have done some more thorough testing and have hit a few more issues. Updating existing forms seems to mostly work, but creating new forms is currently not working and a few question types also aren't working correctly which is preventing me from running through all my test scenarios.

1. At the moment creating a new form for a new event isn't working. This is because the `createApplicationForm` function in applicationForm.service.js is being called with the `event_id` and `is_open` parameters set to `undefined` and so they aren't getting passed to the backend.
2. After clicking save on a new form, when the backend correctly returns a 400 error due to the above missing fields in the request, the page still hides the error and returns to the home page. As mentioned before, it should show this error to the user.
3. If you open the console while the page is running, you’ll notice a lot of warnings/errors from i18next::translator: missingKey  which means that there is translated text missing from the translation.json files. Please check these.
4. The duplicate button isn’t working for questions and the “Duplicate Section” menu option isn’t working for sections.
5. For the references question type, it’s setting the `options` field to `“{min_num_referral: "4", max_num_referral: “6”}”`, but the backend expects that options is always a translated field, so it’s expecting to see 
```
{
  “en”: “{min_num_referral: "4", max_num_referral: “6”}”,
  “fr”: “{min_num_referral: "4", max_num_referral: “6”}”
}
```
6. When there is an error like the above, after clicking Save, the form at least doesn’t go to the home page, but it also doesn’t display an error anywhere (not that I can see at least)
7. When entering the min or max value for a validation, the page crashes with the following error:
```
TypeError: Cannot read property 'split' of null
(anonymous function)
src/pages/createApplicationForm/components/Question.js:59
  56 | if (q.id === inputs.id) {
  57 |   let regex = q.validation_regex[lang];
  58 |   if (prop === 'min' || prop === 'max') {
> 59 |     if (!q.validation_regex[lang].split('{')[1]) {
     | ^
  60 |       regex = `^\s*(\S+(\s+|$)){0,0}$`
  61 |     }
  62 |   }
```

You should be able to reproduce the following issues locally on http://localhost:8080/test2021/applicationform :
8. The validation text still doesn’t appear in some questions when editing an existing form (see first question in section 2)
9. The depends on question functionality doesn’t seem to be working correctly - eg if I select the “May we add your CV and email address […]” question, I expect to see the options “Yes” and “No” to choose from, but these aren’t appearing
10. The depends on question selection is only showing the last bit of the question headline (see question 4 of section 2, the 3rd question in the “depends on question” dropdown is showing “a paper, blog post, algorithm […]” - I realise you truncated the question headline due to space, but it should rather show the first bit of the question headline rather than the last bit)

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
```diff
@@ -70,6 +85,49 @@ const ApplicationForm = (props) => {
...
+            setInitialState(mapedQuestions);
+            setSections(mapedQuestions);
+          } else {
+            setCreateMode(true);
+          }
+        } else {
+          setCreateMode(true);
+        }
+      }).catch(err => {
+        console.log('Error occured ', err)
```
avishkar58, line 129: Please display this error to the user.

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
```diff
@@ -182,100 +347,119 @@ const ApplicationForm = (props) => {
   }
   return (
     <>
+      <Prompt
+        when={!isSaved && !saved}
+        message="Some Changes have not been saved. Are you sure you want to leave without saving them?"
```
avishkar58, line 352: This message needs to be translated please.

webapp/src/pages/createApplicationForm/components/Question.js
```diff
@@ -80,7 +95,7 @@ const Question = forwardRef(({
               let max = e.target.value;
 
               if (max <= min) {
-                max = '';
+                // max = '';
```
avishkar58, line 98: Please remove commented code

webapp/src/services/applicationForm/applicationForm.service.js
```diff
@@ -202,4 +204,73 @@ function getQuestionList(eventId) {
             status: error.response && error.response.status
         };
     });
-}
\ No newline at end of file
+}
+
+function getForEventDetails(eventId) {
```
avishkar58, line 209: Nit: Please call this `getDetailsForEvent` because you're getting "application form details" for an event, not an application form for an "event details"

webapp/src/services/applicationForm/applicationForm.service.js
```diff
@@ -202,4 +204,73 @@ function getQuestionList(eventId) {
...
+      "is_open": is_open,
+      "nominations": nominations,
+      "sections": sections
+    }
+  
+    return axios.post(baseUrl + `/api/v1/application-form-detail`, form, {headers: authHeader()})
+      .then(resp=> {
+        return resp;
+      })
+      .catch(error => {
```
avishkar58, line 268: I think this is why the errors aren't displayed, because you catching them here, but then trying to catch them again in applicationForm.js line 285?

avishkar58 (changes requested):

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
```diff
@@ -113,20 +130,28 @@ const ApplicationForm = (props) => {
   
             setNominate(formSpec.nominations);
             setFormDetails({
-              isOpen: formSpec.is_open,
+              isOpen: props.event.is_registration_open,
```
avishkar58, line 133: This should be `props.event.is_application_open`

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
```diff
@@ -113,20 +130,28 @@ const ApplicationForm = (props) => {
...
               id: formSpec.id,
-              eventId: formSpec.event_id
+              eventId: eventId
             })
             setInitialState(mapedQuestions);
             setSections(mapedQuestions);
           } else {
+            setFormDetails({
+              isOpen: props.event.is_registration_open,
```
avishkar58, line 141: Also `is_application_open`

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
```diff
@@ -113,20 +130,28 @@ const ApplicationForm = (props) => {
...
+              eventId: eventId
+            })
             setCreateMode(true);
           }
         } else {
+          setFormDetails({
+            isOpen: props.event.is_registration_open,
```
avishkar58, line 148: And here

avishkar58 (changes requested):

webapp/src/pages/createApplicationForm/components/ApplicationForm.js
Malaba6: @avishkar58 Done!

avishkar58:
Hi @Malaba6 I'm not sure if you saw my Slack message, so copying it here for visibility:

Hi Eric, just checking if you are missing some commits from the PR? I tried re-testing a few things and immediately noticed that various items still aren't working... in particular:

- The options still aren’t appearing when I select a dependency question which is a multi-choice question
- When there is a 500 error returned from the backend it's still not being shown
- There are still warnings from missing translations
- The reference type still isn't setting the options field correctly. It looks like now it's doing this: {min_num_referral: null, max_num_referral: null, en: {min_num_referral: "2", max_num_referral: "4"}}

I also noticed that
- There is no way to remove a dependency question once you've selected one
- If I add a word limit validation for a question in English, the validation text doesn't appear when I switch to French

I stopped testing at this point because of these issues

avishkar58 (changes requested):
Hi Eric, thanks for the updates! It's looking a lot better now and I really like the placeholders showing the English text when you switch to French - a nice simple solution that improves the useability a lot! :) 

The editing of an existing form seems to work well, however there are still issues with creating a new form: 
- The options are appearing now when I select a dependency for a question, but _not for a section_ (this works when editing the form)
- The new "Clear" option in the dependency question box needs to be translated please
- If I add a second section and start typing the question headline, the first section keeps animating for every character I press in the second section's question headline box
- If I try to highlight the text "Untitled Section" with my mouse, it doesn't let me and instead tries to drag the entire section, Im pretty sure it wasn't doing this before
- When editing the French form, the validation text needs to be in French, not English
- For a referral question, when I switch to the French form, it still shows me the min and max referral values from the English form (fine), but it has the error that I need to select these values, so it seems like the underlying state doesn't match what's being displayed
- I created a new form, and saved it, the operation on the backend was successful and the backend returned a 201 status code, but the page is showing an error "Error: [object Promise]" 
- It seems like the Save button remains disabled until I add a new section (if I just add questions to the default first section that appears for a new form, I couldn't get the Save button to enable even though I filled in everything that was required on both the English and French, only when I added a new section with a new question did it enable) 

Please let me know if you need some assistance with creating a new event in the database so you can test creating a new form locally.

avishkar58:
Thanks for the updates @Malaba6 ! It seems to be working now :) Only issue at the moment is that when creating a new form with dependencies, the dependencies seem to get lost when I go back to edit them, but I'm pretty sure this is a backend issue, so will debug that and merge this PR in the meantime - great work! :)

avishkar58 (approved):

Issue #878, opened 2021-02-24
Create a ReviewResponseSummaryList API


avishkar58:
Create a new ReviewResponse**Summary**List API in reviews/api.py
This is very similar to the ReviewResponse**Detail**List with the key difference that the summary version _groups by response_ and _aggregates the numerical scores_.

In particular, the API should: 
- Accept event_id (int) and language (str) parameters
- Ensure that the logged in user is an event admin for the event
- Query the database for all review responses associated with the given event using a new method on the review repository, as well as some details about the response, the applicant and the reviewer. These details include:
  - The user who submitted the response (called "response_user" in the JSON below)
  - Questions with a key of "review-identifier" and the corresponding Answers from the response
  - The review questions and review scores associated with the review response
- Group the review responses by response_id, drop any scores that are not numerical and aggregate the values
- Compute a **total** which is the weighted average of the numerical scores (weighted by the `weight` value found on review_question)

## Example JSON response
Notes:
- response_identifiers are the headlines (translated) and answers to all questions from the relevant application form that have a key value of "review-identifier"
- The scores should be filtered to questions of type "multi-choice", "long-text" and "short-text"

```
[
  {  
    "response_id": 3,
    "response_user_title": "Ms",
    "response_user_firstname": "Jane",
    "response_user_lastname": "Bloggs",
    "identifiers": [    
        {
            "headline": "Organisation",     // From question_translation
            "value": "Pets R Us"                // From answer
        },
        {
            "headline": "Country",
            "value": "Nigeria"
        },
    ],
    "scores": [
        {
            "headline": "Ethical Considerations",   // From review_question_translation
            "description": "How ethical is the candidate's proposal from 1 to 5?",   // From review_question_translation
            "type": "short-text",   // From review_question
            "score": 4,   // An **average** of the values of review_score
            "weight": 1    // From review_question
        },
        {
            "headline": null,
            "description": "What is your overall rating?",
            "type": "multi-choice",
            "score": 5,
            "weight": 2
        }
    ],
    "total": 12
  },

  {  
    "response_id": 4,
    "response_user_title": "Dr",
    "response_user_firstname": "Alex",
    "response_user_lastname": "Person",
    "identifiers": [    
        {
            "headline": "Organisation",     
            "value": "Nokia"                
        },
        {
            "headline": "Country",
            "value": "South Africa"
        },
    ],
    "scores": [
        {
            "headline": "Ethical Considerations",   // From review_question_translation
[...]

PR #886, opened 2021-03-16
Review Response Summary List

amritpurshotam:
Just worried about the performance of the query since it's sequentially pulling the averages per response which will get worse the more responses there are.

Resolves: #878


avishkar58:
Hey @amritpurshotam - FYI pinning SqlAlchemy to 1.3.23 fixes the weird build error.

amritpurshotam:
@avishkar58 Ah good catch! I was wondering what changed that was causing the issue when I saw it yesterday. I looked up the release notes now on SQL Alchemy and I think it could be [this](https://docs.sqlalchemy.org/en/14/changelog/changelog_14.html#change-d583e32983678010f0173c30c745f874) change related to `select` that then caused the cascading errors. It looks like explicitly setting some parameters should fix it so I'll give it a go when I get stuck into this issue.

Also I have seen your PR and I'll review it first thing tomorrow morning

avishkar58 (approved):
Looks great, thank you Amrit! Lets keep an eye on performance and if necessary we can switch to eagerly loading all the scores and aggregating in the API

PR #887, opened 2021-03-16
Review sections

### LINKED ISSUE #881: Add ReviewSection
At the moment the review form does not support explicit sections, like the application and registration forms. Instead, it uses dummy "header" questions that make it look like there are section dividers when there aren't, which is a bad hack. In order to align better with the application form so that we can reuse as much of the application form designer as possible for the review form designer, add a ReviewSection to the ReviewForm. So the updated model should look like below.

Also update the API to return sections in the form.

# Model

### ReviewForm
- id
- application_form_id
- is_open
- deadline

### ReviewSection
- id
- review_form_id
- order

### ReviewSectionTranslation
- id
- review_section_id
- language
- headline
- description

### ReviewQuestion
- id
- review_section_id
- question_id
- type
- is_required
- order
- weight

### ReviewQuestionTranslation
- id
- review_question_id
- language
- description
- headline
- placeholder
- options
- validation_regex
- validation_text

avishkar58:
Resolves #881


amritpurshotam (approved):
Looks good!

PR #891, opened 2021-04-07
Review stages

### LINKED ISSUE #883: Add Review Stages
Update the ReviewForm model to add stages, in particular, add a `stage` integer and an `active` boolean to the ReviewForm model. Default the stage to 1 and the active to True for all existing ReviewForms in the migration.

Update the any api in reviews/api.py that returns a review form to filter for the *active* review form only. 

Add a new ReviewStageAPI to reviews/api.py which has a POST method which does the following:
- Accept a stage integer and event_id integer as parameters
- Verify that the logged in user is an event admin for the given event_id
- Check that there is a ReviewForm for the event with the given stage (and return an error if there isn't)
- Set the active field to False on all ReviewForms for the event which don't match the incoming stage, and True for the single ReviewForm which does match the incoming stage.

avishkar58:
Resolves #883


api/app/reviews/api.py
```diff
@@ -763,4 +763,54 @@ def get(self, event_id):
...
+        selected_form = selected_form[0]
+
+        for form in review_forms:
+            form.deactivate()
+        
+        selected_form.activate()
+
+        db.session.commit()
+        
+        return {}, 201
```
amritpurshotam, line 816: Just an overall question on the stages. 

Is there a case where no form will be active (like say after all the stages have been completed)? Or will the last stage activated just stay in that state?

api/app/reviews/repository.py
```diff
@@ -86,6 +86,7 @@ def count_unassigned_reviews(event_id, required_reviews_per_response):
     def get_review_form(event_id):
         review_form = (
             db.session.query(ReviewForm)
+                    .filter_by(active=True)
```
amritpurshotam, line 89: This comment all applies to all the repo methods updated with `active=True`. Is there going to be a case where an event organiser is going to want to pull review data from a previously deactivated stage?

amritpurshotam (approved):
Looks good! Just some questions about some extra possible cases around the stages.

api/app/reviews/repository.py
avishkar58: Great point @amritpurshotam - I didn't think about it, but you're totally right, an organiser will definitely want to pull data from previous stages, so in fact we should pass the stage to these repo methods rather than active=True. I'll make these changes before merging this PR.

avishkar58: Will do this in a subsequent PR because I realised I have another branch building on top of this and don't want to deal with the merge conflicts 😂

PR #926, opened 2021-06-08
Feat/iss846/pdf export

smhall97:
@avishkar58 Thank you for the review. We are looking forward to your feedback!


api/app/responses/api.py
```diff
@@ -458,3 +463,80 @@ def get(self, event_id):
         num_reviewers = review_config.num_reviews_required + review_config.num_optional_reviews if review_config is not None else 1
 
         return ResponseDetailAPI._serialize_response(response, language, review_form_id, num_reviewers)
+
+class ResponseExportAPI(restful.Resource):
+
+    def get(self):
+
+        def _get_answer(id, answers):
```
avishkar58, line 471: Better to call this parameter `question_id` to make it clear which ID it's referring to (Answer's also have IDs)

api/app/responses/api.py
```diff
@@ -458,3 +463,80 @@ def get(self, event_id):
...
+
+            return None
+
+        def _get_files(application_form, answers): 
+            # Get the response files that should be exported in the ZIP file
+
+            file_names = []
+
+            for section in application_form.sections:
+                if not section.questions:
```
avishkar58, line 485: This check shouldn't be necessary, I don't think section.questions can be None, and if it's an empty list the for loop will just be skipped.

api/app/responses/api.py
```diff
@@ -458,3 +463,80 @@ def get(self, event_id):
...
+                    continue
+
+                for question in section.questions:
+                    answer = _get_answer(question.id, answers)                
+                    if answer is not None:
+                        # We are only interested in the files, 
+                        # the text answers will be in the main PDF file
+                        if question.type == 'multi-file':
+                            file_names.extend([f for f in json.loads(answer.value)])
```
avishkar58, line 494: The list comprehension here seems unnecessary?

api/app/responses/tests.py
```diff
@@ -985,3 +991,172 @@ def test_response_detail_admin_only(self):
...
+        # Section 1, two questions
+        section1 = self.add_section(application_form.id)
+        self.add_section_translation(section1.id, 'en', name='Section1')
+        question1 = self.add_question(application_form.id, section1.id)
+        self.add_question_translation(question1.id, 'en', headline='Question 1, S1')
+        question2 = self.add_question(application_form.id, section1.id)
+        self.add_question_translation(question2.id, 'en', headline='Question 2, S1')
+
+        # Supplementary file included in application upload (e.g. CV) - Section 1
+        question_supp1 = self.add_question(application_form.id, section1.id, question_type = 'file')
```
avishkar58, line 1012: It would be good to also add a multi-file question, since your API does have some specific logic for handling multi-file

api/app/responses/tests.py
```diff
@@ -985,3 +991,172 @@ def test_response_detail_admin_only(self):
...
+            'response_id': self.response1_id,
+            'language': 'en'
+        }
+        
+        response = self.app.get(
+            '/api/v1/response-export',
+            headers=self.get_auth_header_for('event1admin@mail.com'), 
+            json=params)
+            
+        assert response.mimetype == 'application/zip'
```
avishkar58, line 1090: Please use self.assertEqual()  wherever you've used the Python assert statement. The advantage is that assertEqual gives you better error messages when things don't match.

api/app/responses/tests.py
```diff
@@ -985,3 +991,172 @@ def test_response_detail_admin_only(self):
             json=params)
 
         self.assertEqual(response.status_code, 403)
+
+class ResponseExportAPITest(ApiTestCase):
```
avishkar58, line 995: This is great! :)

api/app/utils/strings.py
```diff
@@ -62,3 +62,47 @@ def build_response_email_body(answers, language, application_form):
                 stringified_summary += '{question}\n{answer}\n\n'.format(question=question_translation.headline, answer=answer_value)
 
     return stringified_summary
+
+
+def build_response_html_app_info(response, language):
+    """
+    Stringifying the dictionary application information, for output in a html file, with the response_id and applicant name contact info as
```
avishkar58, line 69: There isn't a dictionary here though?

api/app/utils/strings.py
```diff
@@ -62,3 +62,47 @@ def build_response_email_body(answers, language, application_form):
...
+     paragraphs (<p>)
+    """
+    
+    stringified_app_info = f"<p><b> Response ID:<b> {response.id}<p> <p><b>Full name:<b> {response.user.firstname} {response.user.lastname}<p>"
```
avishkar58, line 73: The closing tags don't seem correct here? Should be `</b>` and `</p>`  and `</title>` below.

avishkar58 (changes requested):
This is really excellent work @smhall97 and @greg-newman ! Thank you both very much! :) 
I just left a few minor style comments :)

avishkar58:
Please check the tests, seems to be trying to connect to  the real Cloud Storage

avishkar58 (approved):
Great work, thank you!

Issue #915, opened 2021-05-16
Updates to application form designer


avishkar58:
The following features need to be updated/added to the application form designer, they also apply to the review form designer.

1. Add a "sub-heading" question type. This type should allow a header and description as usual, has no placeholder. Questions of this type should always have "is_required" set to *False*

2. Questions with type "multi-file" should also have a placeholder.

3. When the user enters file extensions for "file" and "multi-file" question types, those extensions should be used for all translations. (So for example, if the user enters ".pdf" when editing the English form, ".pdf" should be automatically entered when they switch to the French form)

4. Questions with type "multi-file" should also allow the user to optionally choose a number of files. This will be set in the question options, so the options JSON if the user selects a number of files will look like this:
  `{"accept": ".pdf,.xlsx", "num_uploads": 5}`
and if the user doesn't select a number of files, the JSON should not have the "num_uploads" key, so
  `{"accept": ".pdf,.xlsx"}`
The value should also be the same for all translations

5. The validation text that the form automatically enters needs to be translated. At the moment if I select validation on the English form with a word limit, the validation text is set to something like "You must enter between 10 and 50 words" - but when I switch to the French form, the numbers are remembered, but the text is still in English.

avishkar58:
@Malaba6 Hi Eric - here are the updates needed for the application form and review form designers, please let me know if there are any questions.

avishkar58:
Fixed in #929

Issue #893, opened 2021-04-25
Create Review Form Detail API


avishkar58:
Create a ReviewFormDetailAPI in `api/app/reviews/api.py`.

- A GET method returns the serialised review form for the given event_id and stage, according to the JSON representation below.
- A POST method creates a new review form for the given event_id and stage.
- A PUT method updates an exiting review form for the given event_id and stage, returning an error if the form being updated doesn't exist. This should add any new sections and questions if their ID property is missing/null, update fields for existing sections and questions and delete any sections and questions that are in the database but not in the JSON input.

All methods should verify that the logged in user is an event admin for the given event_id.

## Example JSON
```
{
  "id": 5,  // Not on POST requests
  "event_id": 1,
  "application_form_id": 1,  
  "stage": 2,
  "is_open": true,   // Just default this to true
  "deadline": "2021-06-01 00:00:00"       // ISO 8601 format, default this to the review_close value from the event
  "active": true,     // Just default this to true
  "sections": [
      {
          "id": 8,   // Not on POST requests, only on PUT requests if editing an existing section
          "order": 1,
          "headline": {
             "en": "Section 1",
             "fr": "Section 1"
          },
          "description": {
             "en": "This is the first section",
             "fr": "Ceci est la première section"
          },
          "questions": [
              {
                 "id": 2,  // Not on POST requests, only on PUT requests if editing an existing question
                 "type": "short-text",
                 "is_required": false,
                 "order": 1,
                 "weight": 0,
                 "question_id": null,
                 "headline": {
                     "en": "Question 1",
                     "fr": "Question 1"
                 },
                 "description": {
                     "en": "This is the first question",
                     "fr": "C'est la première question"
                 },
                 "placeholder": {
                     "en": "Enter something interesting",
                     "fr": "Entrez quelque chose d'intéressant"
                 },
                 "options": {
                     "en": null,
                     "fr": null
                 },
                 "validation_regex": {
                     "en": "(a|b|c|d)",
                     "fr": "(a|b|c|d)"
                 },
                 "validation_text": {
                     "en": "You must enter a, b, c or d",
                     "fr": "Vous devez entrer a, b, c ou d"
                 }
              },   // End question 1
              {
                 "id": 3, 
                 "type": "radio",
                 "is_required": true,
                 "order": 2,
                 "weight": 1.5,
                 "question_id": null,
                 "headline": {
                     "en": "Question 2",
                     "fr": "Question 2"
                 },
                 "description": {
[...]

PR #929, opened 2021-06-16
Review form creator #893

### LINKED ISSUE #892: Review Form Designer Front End
# Background

(_**Bold** text represents a concept or entity in Baobab._)

Baobab collects **applications** from **candidates** to **events**. These **responses** consist of **answers** to **questions** posed in an **application form**. Each **response** from a **candidate** is **reviewed** by one or more **reviewers**. The **reviewer** sees **answers** to some of the **questions** from the **candidate's** **response** and then enters **scores** in response to **review questions** posed to them. These **scores** may be numerical selections or long-form comments. Since the **questions** on the **application form** differ per **event** and the **questions** the **reviewers** are required to answer also differ per **event**, the **review form** itself is _configurable_ and consists of **sections** and **questions**, very similar to the way the **application form** is configured. Reviews can also occur in *stages*, for example, applications may go through a light first stage review with only a handful of questions and then short-listed applications proceed to a more in depth second stage review. 

# Requirements

## New Page
Create a new review form builder page. 
* Add a route to it in `EventHome.js` of the form `${match.path}/reviewform` - (this makes it an event specific route)
* Add a link to this route in the Event Admin menu in App.js (it should therefore only be visible to users with the “admin” event role) 
  * Specifically, make the current "Form Settings" menu item a sub-menu, with two items, "Application Form" which points to the existing application form designer and "Review Form" which points to this new review form designer page. 

## Components
The review form builder consists of sections and questions just like the application form builder. Users should be able to add sections, add questions to sections and reorder sections in the form and questions in sections. The review form is also _translatable_ and the user should be able to define translations for the form in all the languages supported by the event. These components should be shared as much as possible. However, there are some key differences.

## Review Form Level
Query the event details service to get more information about the event (more than what's in the event object passed as a prop to the component)
* Display the review form's open and close date, which can be obtained from the event object's `review_open` and `review_close` dates

Query the [`reviewstage`](https://github.com/deep-learning-indaba/Baobab/pull/891/files) API (`api/v1/reviewstage`) to get the current active stage and the number of stages - see example JSON below.
* Display a "timeline" at the top of the page that shows all the available stages, with the current stage highlighted. If the total number of stages is N, then this timeline should show 1, 2, 3, ..., N. The user should be able to view a different stage by clicking on it.
* There should also be a "plus" button that allows the user to add a new stage to the end. So if the current total_stages is N, this plus button will add stage N+1.

Query the `review-form-detail` API, initially with only the `event_id` as a parameter, which will return the review form details for the current stage. 
* If this API returns 404 it means that no review form has been configured, so the page should enter "create" mode. If the API returns something, the page should enter "edit" mode and display the contents of the review form returned by the backend. 
* Whenever the user selects a different stage in the timeline, re-query this API, including a `stage` parameter in addition to the `event_id` to get the review form details for that stage. You don't necessarily need any special logic for handling the plus button adding a new stage, because querying the API with this new stage as the parameter will still result in a 404 which should put the form in "create" mode. 
* See the example JSON below which shows the format of data this API is expected to return. 
* This API hasn't been created yet, see issue #893 for progress.

Query the `applicationformdetail` API (the same one used by the application form builder, at `/api/v1/application-form-detail`) to get the details of the application form. These details will be used by a special question type, "information".

There should be a "Save" button, which sends a POST request to the review-form-detail API if the form is in "create" mode and a PUT request if the form is in "edit" mode. These API methods expect inputs according to the JSON example below. Note that for the PUT method, the "id" field should be present for any sections/questions that are being updated and it should be absent for any new sections/questions so that the backend can distinguish between questions/sections that are new and being updated. 

## Section Level
Just like the application form, the review form consists of multiple sections. Again, just like the application form builder, the user should be able to add sections, remove sections, reorder sections, collapse sections and duplicate sections and therefore ideally the component(s) should be shared with the application form builder. 

Like the application form, Sections have a `headline` and `description` which are **translatable** fields.

The review form does not support the following features, which should be disabled:
* Review Form Sections don't have a `key` property, so the option to add a key should be disabled
* Review Form Sections don't have dependencies, so the "depends on question" feature should be hidden. We will however want to add this feature in the future, so ideally this should be controlled with a flag on the component that we can re-enable in the future. 

There is also a "special question type" that can appear on review forms. So, in addition to the "add question" button, a section should also have an "Add answer from application form" button which will add a special kind of question block that is described more below. 

## Question Level
Just like the application form, each section in the review form consists of multiple questions. Like the application form builder, a user should be able to add questions to sections, remove questions using a trash icon, duplicate questions and reorder questions inside a section. 

Questions have a `headline`, `description`, `placeholder`, `options`, `validation_regex` and `validation_text` properties in their JSON representation which are all **translatable** and handled exactly the same as the application form. 

Question can be marked as _required_, which corresponds to an `is_required` property on the question JSON.

Questions **do not** have a `key` property, so the option to add a key should be removed if reusing the component from the application form builder.

Questions **do not** have dependencies, so the "depends on question" feature should be disabled if reusing the component from the application form builder. Like the section though, we will want to enable this feature in the future so it should be easy to re-enable.

A property unique to the review form questions is a `weight`. If the selected question type supports weights (see table below), then a field should appear which allows the user to enter a numeric (floating point) value, labelled "Score Weight". The default value should be 0 (and this is the value that should also be set for question types that don't support weights) 

Review form questions support the following question types (where there is overlap with the application form builder, these are treated the same, types that are unique to the review form are in bold)

| Label | Value | Has Placeholder | Has Weight | Has Options |
| --- | --- | --- | --- | --- |
| Short Answer | short-text | True | False | False |
| **Numeric Input** | numeric-text | True | True | False |
| Paragraph Answer | long-text | True | False | False |
| Single Checkbox | single-choice | False | True | False |
| Dropdown | multi-choice | True | True | True |
| **Radio Buttons** | radio | False | True | True |
| Multiple Choices (Checkboxes) | multi-checkbox | False | False | True |

In the table above, "Has Placeholder" means that the use should be able to enter (translatable) Placeholder text when that question type is selected. "Has Weight" means that the user should be able to enter a numeric "Score Weight" when that type is selected, "Has Options" means that the user should be able to enter options when that question type is selected (the options work exactly the same as the application form builder)

Just like the application form, the user should also be able to add validation to questions, and the logic around validation is exactly the same as the application form builder.

In addition to the above types, there is also a question type with a value of "information". which corresponds to "Answer From Application Form". Questions with these types should be treated specially and displayed differently according to the description below:

## Answer From Application Form (Special Question Type)
When the question type is "information", the question should be handled differently. In particular:
- There are still headline and description translatable properties.
- There should be a dropdown box which lists all of the questions (headlines) from the **application form**. Ideally this dropdown list should have section titles in it (which are not selectable), with the questions appearing underneath their section title in the correct order. 
  - Remember to display the headline in the dropdown in the correct language (corresponding to the language the user is editing)
- When the user selects a question from the dropdown list, the _headline_ and _description_ of the review form question should be automatically populated by copying the values from the application form question. This should happen for all available languages. (The user can still edit these if they wish to) 

- The question selected in the dropout list should correspond to the `question_id` property in the JSON representation.
- This question type has no placeholder, weight or options and the user shouldn't see or be able to change the question type. 
- The required checkbox should be hidden and the value of the `is_required` property defaulted to False. 
- These question types can still be deleted, reordered and duplicated just like other question types.

Note: When the user clicks on the "Add answer from application form" button in the section, it should add a question with type "information" and display it according to the description above.

# Example JSON
## ReviewStageAPI
The ReviewStageAPI GET method returns JSON that looks like this:

```
{
  "current_stage": 2,
  "total_stages": 3
}
```

## Review Form Detail API
The ReviewFormDetailAPI GET method will return JSON that looks like this. The POST and PUT methods expect JSON formatting in the same way. 

```
{
  "id": 5,  // Not on POST requests
  "event_id": 1,
  "application_form_id": 1,  
  "stage": 2,
  "is_open": true,   // Just default this to true
  "deadline": "2021-06-01 00:00:00"       // ISO 8601 format, default this to the review_close value from the event
  "active": false,     // If this is a POST request (ie a new form stage) and the stage number is 1, set this to true, otherwise false. If it's a PUT request, don't change the value
  "sections": [
      {
          "id": 8,   // Not on POST requests, only on PUT requests if editing an existing section
          "order": 1,
          "headline": {
             "en": "Section 1",
             "fr": "Section 1"
          },
          "description": {
             "en": "This is the first section",
             "fr": "Ceci est la première section"
          },
          "questions": [
              {
                 "id": 2,  // Not on POST requests, only on PUT requests if editing an existing question
                 "type": "short-text",
                 "is_required": false,
                 "order": 1,
                 "weight": 0,
                 "question_id": null,
                 "headline": {
                     "en": "Question 1",
                     "fr": "Question 1"
                 },
                 "description": {
                     "en": "This is the first question",
                     "fr": "C'est la première question"
                 },
                 "placeholder": {
                     "en": "Enter something interesting",
                     "fr": "Entrez quelque chose d'intéressant"
                 },
                 "options": {
                     "en": null,
                     "fr": null
                 },
                 "validation_regex": {
                     "en": "(a|b|c|d)",
                     "fr": "(a|b|c|d)"
                 },
                 "validation_text": {
                     "en": "You must enter a, b, c or d",
                     "fr": "Vous devez entrer a, b, c ou d"
                 }
              },   // End question 1
              {
                 "id": 3, 
                 "type": "radio",
                 "is_required": true,
                 "order": 2,
                 "weight": 1.5,
                 "question_id": null,
                 "headline": {
                     "en": "Question 2",
                     "fr": "Question 2"
                 },
                 "description": {
                     "en": "This is the second question",
                     "fr": "C'est la deuxième question"
                 },
                 "placeholder": {
                     "en": null,
                     "fr": null
                 },
                 "options": {
                     "en": [
                          {"value": "0", "label": "Bad"}, 
                          {"value": "1", "label": "Average"}, 
                          {"value": "2", "label": "Good"}, 
                      ],
                     "fr": [
                          {"value": "0", "label": "Mal"}, 
                          {"value": "1", "label": "Moyenne"}, 
                          {"value": "2", "label": "Bien"}, 
                      ],
                 },
                 "validation_regex": {
                     "en": null,
                     "fr": null
                 },
                 "validation_text": {
                     "en": null,
                     "fr": null
                 }
              }  // End Question 2
          ]
      },   // End section 1
      {
          "id": 9,   
          "order": 2,
          "headline": {
             "en": "Section 2",
             "fr": "Section 2"
          },
          "description": {
             "en": "This is the first section",
             "fr": "Ceci est la deuxième section"
          },
          "questions": [
              {
                 "id": 3,
                 "type": "information",
                 "is_required": false,
                 "order": 1,
                 "weight": 0,
                 "question_id": 3,    // This should be the id of a question on the application form
                 "headline": {
                     "en": "Question 1 from application form",
                     "fr": "Question 1 from application form"
                 },
                 "description": {
                     "en": "This is the first question from the application form",
                     "fr": "C'est la première question du formulaire de candidature"
                 },
                 "placeholder": {
                     "en": null,
                     "fr": null
                 },
                 "options": {
                     "en": null,
                     "fr": null
                 },
                 "validation_regex": {
                     "en": null,
                     "fr": null
                 },
                 "validation_text": {
                     "en": null,
                     "fr": null
                 }
              }
          ]
      }   // End section 2
  ]
}
```

Malaba6:
Resolves #892, #915


webapp/src/components/answerValue/__tests__/MultiFileValue.test.js
```diff
@@ -16,21 +16,17 @@ describe('MultiFileValue Tests', () => {
   });
 
   it('Renders list of files.', () => {
-    const value = `[{"id":1,"file":"1.tst","name":"lorem.tst"},{"id":2,"file":"2.tst","name":"ipsum.tst"}]`
+    const value = `[{"id":1,"filename":"1.tst","rename":"lorem.tst"},{"id":2,"filename":"2.tst","rename":"ipsum.tst"}]`
```
Malaba6, line 19: made changes here to resolve this failing test

avishkar58: Awesome thank you!

api/migrations/versions/186e11949103_.py
```diff
@@ -0,0 +1,26 @@
+"""empty message
```
avishkar58, line 1: Please remove this file

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,310 @@
...
+      </div>
+    )
+  }
+  return (
+    <>
+      <Prompt
+        when={!isSaved && !saved}
+        message="Some Changes have not been saved. Are you sure you want to leave without saving them?"
+        />
+      {homeRedirect && <Redirect to="/" />}
```
avishkar58, line 144: Can we redirect to the event home page instead of the main home page? When using the application form builder I found it a bit confusing to go back to the main home page, when I started on the event home page.

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,310 @@
...
+      {errorResponse ? (
+        <div className='tooltiptext-error response-error'>
+          <Trans i18nKey='errorResponse' className='response-error'>{{errorResponse}}</Trans>
+        </div>
+      ) : (
+        <div className='application-form-wrap'>
+          <TopBar
+            title={setNominate ? "Application Form" : "Review Form"}
```
avishkar58, line 152: Can you pass in the title rather? In future we'll want to use this for even more form types and conditioning on setNominate seems brittle.

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,310 @@
...
+              value="Save"
+              style={!isReview ? { marginTop: '16em'} : {}}
+              className='save-form-btn'
+              data-title="Save"
+              onClick={handleSave}
+              disabled={isSaveDisabled || disableSaveBtn}
+              />
+          </div>
+          <div className="application-form-wrapper">
+            {setNominate && (
```
avishkar58, line 176: All this logic that depends on setNominate is specific to the application form, so I'd expect to find it in ApplicationForm.js and not here.

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,310 @@
...
+                  {t('Application closes ') + ' :'}
+                  <span className="date">
+                    {`${dateFormat(evnt.application_close)}`}
+                  </span>
+                </span>
+              </div>
+            ) : (
+              <div className="dates-container">
+                <span className="dates">
+                  {t('Review opens ') + ' :'}
```
avishkar58, line 210: same here, it would make sense to keep the review specific stuff in ReviewForm.js. Otherwise this will get especially tricky when we introduce a third form type!

webapp/src/pages/createApplicationForm/components/Section.js
```diff
@@ -370,15 +340,16 @@ export const Section = forwardRef(({
...
-            <button
-              className="dropdown-item delete-section"
-              onClick={handleKey}
-            >
-              {isKeyOn
-                && <i className="fas fa-check fa-duplicate fa-section"></i>
-              }
-              {t("Add Key")}
-            </button>
+            {!isReview && (
```
avishkar58, line 343: Would prefer to have explicit feature flags like "hasKey" has "hasDependency" etc rather than using isReview, again because there might be a different set of features used in a third form type

avishkar58 (changes requested):
The code looks excellent, thank you @Malaba6 ! 
I will do some testing tomorrow and get back to you if I find any issues.

webapp/src/pages/createApplicationForm/components/Section.js
Malaba6: Done! You can review again @avishkar58

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,261 @@
...
+      </div>
+    )
+  }
+  const eventId = evt.id;
+
+  return (
+    <>
+      <Prompt
+        when={!isSaved && !saved}
+        message="Some Changes have not been saved. Are you sure you want to leave without saving them?"
```
avishkar58, line 145: This text needs to be translated...

avishkar58 (changes requested):

webapp/src/components/form/FormCreator.js
```diff
@@ -0,0 +1,261 @@
...
+        />
+      {homeRedirect && <Redirect to={`/${eventId}`} />}
```
avishkar58, line 147: The URL needs to be `/eventKey`   not the ID :)

webapp/src/components/form/FormCreator.js
Malaba6: Oh yes!

avishkar58:
Hi Eric,

I did some more testing on this feature and while it generally looks amazing, I did pick up some issues, the first of which meant I couldn't complete testing:
1. I get a bad request when trying to save, because it seems the request is missing the application_form_id parameter. 

Other minor issues:
1. When the above bad request happened, the and the error didn't seem to get extracted correctly and the UI displayed “Error: [object Object]”
2. When I have just started editing a review form and try to delete the first question it says “it is a dependency of another question are you sure you want to delete it”, when it isn’t a dependency
3. When I try to edit the weight, when I press backspace to delete the weight of 0, it doesn’t delete, I have to highlight the 0 and type 1 to change the value

To avoid lengthy back and forth, can you please try to test yourself that creating a new form and editing an existing form works correctly with the backend. If you need to create a new event to test with, you can run the following SQL statements on your local database.

```
INSERT INTO event(start_date, end_date, email_from, "key", organisation_id, url, application_close, application_open, offer_close, offer_open, registration_close, registration_open, review_close, review_open, selection_close, selection_open, event_type, travel_grant, miniconf_url)
VALUES('2021-08-01', '2021-08-01', 'baobab@deeplearningindaba.com', 'test-event', 3, 'deeplearningindaba.com', '2021-07-01', '2021-06-01', '2021-12-31', '2021-12-31', '2021-12-31', '2021-12-31', '2021-12-31', '2021-06-01', '2021-12-31', '2021-12-31', 'EVENT', False, Null);

INSERT INTO event_translation("event_id", "name", description, "language") values
((select max(id) from event), 'Test Event', 'Test Event', 'en');

INSERT INTO event_translation("event_id", "name", description, "language") values
((select max(id) from event), 'Test Event', 'Test Event', 'fr');
```

avishkar58 (approved):
Excellent work, thanks @Malaba6 !

Issue #922, opened 2021-05-25
Add more logging/audit to form responses


avishkar58:
At the moment it's not possible to see when someone edits the responses to an application or review form. Add some form of audit trail that will allow us to see what happened to a form - this will help diagnose issues that users report. 
- Ideally we would want to store a full copy of every edit that's made to a response eg by introducing versioning to the Response/ReviewResponse/RegistrationResponse tables.
- At the very least we should store the timestamps of every edit somewhere.

bdhobare:
Hi @avishkar58 , is this task claimed? If not, I would love to claim it :)

avishkar58:
Hi @bdhobare ! This hasn't been claimed, please feel free to assign it to yourself :) Let me know if you need any more information

avishkar58:
@bdhobare if you do take this issue, please write a description of how you intend to solve it so that we can ensure we're all on the same page before you start writing code :)

bdhobare:
Hello @avishkar58 
To minimize modifications to the API code, I believe using a `TRIGGER` function would be the fastest way to have an audit trail. The function would be executed for updates to the columns of interest.
### Example using `registration_answer` table
1.  Create an audit trail table:
```
DROP TABLE IF EXISTS registration_answer_audit;
CREATE TABLE  registration_answer_audit(
    id SERIAL,
    registration_answer_id INT NOT NULL,
    value VARCHAR,
    changed_on TIMESTAMP NOT NULL,
    PRIMARY KEY(id)
);
```
2. Create a migration to create a PL/PGSQL function:
```
CREATE OR REPLACE FUNCTION log_registration_answer_changes()
    RETURNS TRIGGER
    LANGUAGE plpgsql
    AS
$$
BEGIN
    IF NEW.value <> OLD.value THEN
        INSERT INTO registration_answer_audit(registration_answer_id, value, changed_on)
        VALUES(OLD.id, OLD.value, NOW());
    END IF;
    RETURN NEW;
END;
$$
```
2. Create a trigger to attach this function to the  `registration_answer` table:
```
CREATE TRIGGER registration_answer_changes
    BEFORE UPDATE
    ON registration_answer
    FOR EACH ROW
    EXECUTE PROCEDURE log_registration_answer_changes();
```
3. Now for every UPDATE to the `value` column we log the changed value to `registration_answer_audit`  table:
```
UPDATE registration_answer SET value = 'Modified first answer' WHERE id = 2;
[...]

PR #935, opened 2021-08-16
(feat): Add audit trail support #922

bdhobare:
Issue #922 
## Summary
- Added new column `is_active` to the tables `answer`, `review_score` and `guest_registration_answer`. This column is `true` when the entry is the latest and `false` for historical/updated entries.
- Updated repository methods to filter using this new column.
- Added new column `created_on` which tracks when entries were created.
- Updated the PUT api to insert new entries and mark them as active and mark previous entries as not active.
- Updated `Response.answers` to also filter only active answers.

## Tests done

1. `docker-compose run web bash -c 'cd api; nosetests -v app.responses.tests'`
2. `docker-compose run web nosetests`
3. CircleCI 
3. Manual verification using postman.


avishkar58 (changes requested):
This looks great, thanks @bdhobare ! 
Can you also add a timestamp so that we know when the answers were updated?

api/migrations/versions/794506c08e10_.py
```diff
@@ -0,0 +1,38 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('review_score', sa.Column('is_active', sa.Boolean, nullable=True))
+    op.add_column('guest_registration_answer', sa.Column('is_active', sa.Boolean(), nullable=True))
+    op.add_column('answer', sa.Column('is_active', sa.Boolean(), nullable=True))
```
amritpurshotam, line 21: Can we also please update the accompanying `GuestRegistrationAnswer` model in `./api/app/invitedGuest/models.py` and `ReviewScore` model in `./api/app/reviews/models.py` and their accompanying references similar to what you did for `Answer` as well?

amritpurshotam (changes requested):
We do need to make some updates but overall good job @bdhobare, thanks for your contribution!

api/migrations/versions/794506c08e10_.py
bdhobare: Hi @amritpurshotam , I thought it would be better for me to submit these changes in 2 separate PRs since putting them here might make this PR too big and hard to review. What do you think?

amritpurshotam: Hey @bdhobare, that's a good point but since these are quite similar and tightly coupled components I don't feel it would be more difficult on me to review. Perhaps if it helps you to develop against and keep it small on your side, maybe branch off this one, make the changes there, then merge back here once it's ready for review?

bdhobare: @amritpurshotam please see the changes.

amritpurshotam:
Hey @bdhobare. Thanks for the updates! Just been a little swamped to review them properly but I should get a chance this weekend. Apologies for the delay.

api/migrations/versions/794506c08e10_.py
```diff
@@ -0,0 +1,53 @@
...
+
+    op.add_column('review_score', sa.Column('updated_on', sa.TIMESTAMP, nullable=True,
+                                            server_default=sa.func.now(), onupdate=sa.func.now()))
+    op.add_column('guest_registration_answer', sa.Column('updated_on', sa.TIMESTAMP, nullable=True,
+                                                         server_default=sa.func.now(), onupdate=sa.func.now()))
+    op.add_column('answer', sa.Column('updated_on', sa.TIMESTAMP, nullable=True,
+                                      server_default=sa.func.now(), onupdate=sa.func.now()))
```
amritpurshotam, line 28: None of these columns have been added to the models. Was this intentional? I can see they would default to the current date and the columns and values would exist in the database but I'm uncomfortable with them being _hidden_ in these scripts especially since we're using a model/code first approach. What do you think about moving these to the model class and also explicitly defining them there?

amritpurshotam (commented):
Hey @bdhobare. Looks mostly good! Just have a question about the `updated_on` columns, let me know what you think.

api/migrations/versions/794506c08e10_.py
bdhobare: This definitely makes sense @amritpurshotam . I overlooked defining them in the model class because I saw they are not used in any part of the code because the date is only set and not modified anywhere. Let me declare them in the models too.

amritpurshotam: Sorry missed this update 😓 I see they're on the model class now, thanks!
There is one more thing though, I should have mentioned this on the last comment so apologies for the inconvenience.
Can we please rename the column to `created_on` instead and remove the `onupdate=sa.func.now()` bit (in fact I'm in two minds if we should even have the `server_default` and just explicitly set the field manually since that's what's done elsewhere in the codebase)? The reason is that we don't want this value to ever change (especially accidentally) and having the name reflect that would help communicate this rule. 

P.S. you mentioned that the field isn't used anywhere which is a fair assumption. But it will become important later for when we perhaps want to display all the answers (for say an admin of some kind), it would be nice to order them by the creation date and having the field on the model will be required for this. A second reason is that any database changes we make must flow from the codebase (and never the other way round). We have documented the process [here](https://github.com/deep-learning-indaba/Baobab/tree/develop/api#database-migrations) with the commands to run to generate the initial migration. We then edit it if necessary, for eg. with new not null columns, or we edit the code again to make sure the migration we want gets generated correctly). We shouldn't normally be writing migrations from scratch.

bdhobare: Updated with the new recommendations. `server_default` left for now.

amritpurshotam (approved):

api/migrations/versions/794506c08e10_.py
amritpurshotam: @bdhobare Thanks for the update! Approved from my side 😄

bdhobare:
Hi @avishkar58 , waiting for your approval to merge.

avishkar58 (approved):
Apologies for the delay. Looks good to me, thank you for the contribution! :)

PR #942, opened 2021-10-10
Fix/build

amritpurshotam:
- Fixed building of docker images
- Fixed local connection string to the `db` and the fake gcs `storage` by using the service name instead of the IP address.
- Optimised image by reducing layers and increasing the amount of reuse
- Fixed running of front end CI tests along with using latest CircleCI ubuntu images
- Pinned itsdangerous, markupsafe, and wtforms due to breaking changes introduced in latest updates

Still investigating why the integration tests are failing. Currently it looks like the frontend is just not starting at all.

I just have one question regarding `api/app/utils/storage.py` in the `get_storage_bucket()` method, I saw there were some hotfixes made to only use the real storage bucket due to issues in prod but this of course makes the tests fail (specifically the ones in `ResponseExportAPITest` which can be seen [here](https://app.circleci.com/pipelines/github/deep-learning-indaba/Baobab/2480/workflows/c1a2c36d-f225-4a71-bb67-fa7540ba0613/jobs/7843)). In this PR I've undid those changes to get the tests to pass but do we know what the underlying issue was?


amritpurshotam:
@avishkar58 So everything looks to still be working (at least on Win11 with the latest docker version). Do you remember though what was happening with the storage buckets (details in the PR description)?

And @greg-newman I thought to add you to this in case you have any insight into the storage bucket issue as well since I saw you worked on the Response API Export.

avishkar58 (approved):

PR #1082, opened 2023-02-17
New event for admins UI

avishkar58:
@danielamassiceti 
Created a PR for easier reviewing


webapp/src/pages/home/Home.js
```diff
@@ -105,6 +108,13 @@ class Home extends Component {
         if (this.state.organisation && this.state.organisation.name === "AI4D Africa" && this.props.i18n.language === "fr") {
             logo = "ai4d_logo_fr.png";
         }
+        let new_event_definition = {
```
avishkar58, line 111: const ;)

webapp/src/pages/home/Home.js
```diff
@@ -97,6 +96,10 @@ class Home extends Component {
         }
         return <div></div>
     }
+    
+    renderButton = (definition) => {
```
avishkar58, line 100: Not sure you really need this function. The `renderEventTable` is done that way because we're applying a fairly complex template multiple times, but in this case I think it's clearer to just have the HTML directly where you want the button to go :)

avishkar58 (commented):
Hey @danielamassiceti ! This is looking really good so far, thank you for all your hard work on this!! I added a lot of comments as you requested 😅 please let me know if you need any more info or context

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -1,19 +1,47 @@
...
 import { withTranslation } from 'react-i18next';
+import FormTextBox from "../../../components/form/FormTextBox";
+import FormTextArea from "../../../components/form/FormTextArea";
+import FormDate from "../../../components/form/FormDate";
 
 class EventConfigComponent extends Component {
   constructor(props) {
     super(props);
 
+    let today = new Date().toLocaleDateString();
```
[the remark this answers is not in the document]
danielamassiceti: Good to know, thanks

webapp/src/pages/home/Home.js
danielamassiceti: Done. And also removed new_event_definition - just creating the button directly

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -1,19 +1,47 @@
...
+      selection_open: today,
+      selection_close: today,
+      offer_open: today,
+      offer_close: today,
+      registration_open: today,
+      registration_close: today  
+    }
+
     this.state = {
       preEvent: this.emptyEvent,
```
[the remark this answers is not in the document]
danielamassiceti: I am not quite sure what the preEvent is doing currently. Might it be useful if a user is editing an existing, rather than creating a new, event? Should this form be used for both? If so, I was thinking of adding some conditional displays (e.g. use Update Event or Create Event buttons depending on which state they're in)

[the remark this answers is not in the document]
avishkar58: Yes definitely this form should be used for both editing and creating new events, but even in the editing case I can't really see a use for preEvent, unless maybe you want to do the "field edited" thing for validation :) If using editing mode you would know from the route that you're editing and could just put a simple flag in your state to control the submit button text

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -167,13 +170,17 @@ class EventConfigComponent extends Component {
...
-    let isValid = this.validateEventDetails() && this.validateDateTimeEventDetails();
+  validate = () => {
+    const isValid = this.validateEventDetails() && this.validateDateTimeEventDetails();
     this.setState({
-      hasBeenValidated: isValid
+      isValid: isValid
     });
   }
 
+  hasBeenValidated = () => {
```
avishkar58: Do you use this anywhere?

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -57,9 +87,113 @@ class EventConfigComponent extends Component {
...
     }
+    this.validateEventDetails();
     this.setState({
       hasBeenUpdated: isEdited
     });
+
+  };
+
+  getErrorMessages = errors => {
+    let errorMessages = [];
```
avishkar58, line 98: This can also be const :)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -57,9 +87,113 @@ class EventConfigComponent extends Component {
...
+
+    for (let i = 0; i < errors.length; i++) {
+      errorMessages.push( //TODO: warning appears for this - each child in list should have key prop
```
avishkar58, line 101: The key needs to be unique to distinguish items in a list, so you could do something like `key={"error" + i}`

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -78,7 +212,7 @@ class EventConfigComponent extends Component {
   updateDateTimeEventDetails = (fieldName, value) => {
     let u = {
```
avishkar58, line 213: const

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -137,275 +274,306 @@ class EventConfigComponent extends Component {
...
             </div>
 
             <div className={"form-group row"}>
-              <label
+              <label 
                 className={"col-sm-2 col-form-label"}
-                htmlFor="description">
-                {t("Description")}
+                htmlFor="key">
+                {t("Event Key")+"*"}
```
avishkar58, line 308: Is the asterisk to indicate that it's required? If so, have a look at the CSS class we have for this:
<span className="required-indicator">*</span>
(assumption with this is that it comes before the field name, rather than after like you have it)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
danielamassiceti: Removed

webapp/src/pages/eventConfig/components/EventConfigComponent.js
danielamassiceti: Done

webapp/src/pages/eventConfig/components/EventConfigComponent.js
danielamassiceti: Good to know. Thanks.

webapp/src/pages/eventConfig/components/EventConfigComponent.js
danielamassiceti: I wasn't sure where to find this CSS - it doesn't appear in EventConfig.css. If I just add 'required-indicator' into the label's className, it doesn't show any asterisk so I think it may not be finding it.

webapp/src/pages/eventConfig/components/EventConfigComponent.js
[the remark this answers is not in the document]
danielamassiceti: Got it. Though currently hasBeenEdited is cycling through this.state.preEvent to see if any fields have been edited before updating the state. What is the best way to replace this?

danielamassiceti:
Some questions/comments:
1) For the dropdown menus, it seems that the event handler requires a different call format to the event handlers for the text fields/dates. In this case, I just call it as onChange={updateDropDownEventDetails} rather than onChange={e => updateEventDetails('field_name', e)}. This seems a bit inconsistent though. Am I missing something?
2) I switched the emptyEvent to have nulls rather than empty strings for the various fields, but now I am getting a warning saying that the value of some components should not be null, but rather undefined. What's the best way to init these fields at the very beginning?
3) I was also getting a warning about 'Invalid DOM property `class`. Did you mean `className`?' which seems to originate from 'FormTextBox (at EventConfigComponent.js:320)'. But I have got className there, so not sure what the error is.

avishkar58:
> Some questions/comments:
> 
> 1. For the dropdown menus, it seems that the event handler requires a different call format to the event handlers for the text fields/dates. In this case, I just call it as onChange={updateDropDownEventDetails} rather than onChange={e => updateEventDetails('field_name', e)}. This seems a bit inconsistent though. Am I missing something?
> 2. I switched the emptyEvent to have nulls rather than empty strings for the various fields, but now I am getting a warning saying that the value of some components should not be null, but rather undefined. What's the best way to init these fields at the very beginning?
> 3. I was also getting a warning about 'Invalid DOM property `class`. Did you mean `className`?' which seems to originate from 'FormTextBox (at EventConfigComponent.js:320)'. But I have got className there, so not sure what the error is.

1. Yeah unfortunately that is the case, and the pattern you've suggested is also the one we've used elsewhere. Ideally the FormDropDown should wrap things so that it's consistent with the rest of the controls, but it's one of the earlier controls we built and I guess we didn't think that through 
2. Ah, this is one bit of React that still confuses me 😂 the whole controlled and uncontrolled components thing and how they don't like nulls... hmm.. empty strings should be fine though? Was there a reason to change them to null?
3. Ah it must be coming from somewhere inside FormText then. I can see MarkdownRenderer.js has an incorrect `div class=...` for example. Maybe try change it there and see if the warning goes away? There's also `<i class=` in FormToolTip.js but I can't remember if `i` needs to use class or className, I think that one might be correct.

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -414,22 +699,33 @@ class EventConfigComponent extends Component {
...
+            
             <div className={"col-sm-4 "}>
-              <button
-                onClick={() => this.onClickSubmit()}
-                className="btn btn-success btn-lg btn-block"
-                disabled={!hasBeenUpdated}>
-                {t("Update Event")}
-              </button>
+              {isNewEvent ? (
+                <Link to="..">
```
avishkar58, line 709: Ah sorry for not being more precise when I told you about Link! It will work fine for your cancel button, but for your submit you'll only want to go back conditionally if the post succeeds and if there are no validation or post errors, so Link won't work for this because i think it will always go back no matter what happens in your onClickSubmit function (is that what you're observing?) so for this case, you can put some code in your onClickSubmit after succeeding that navigates back instead, for example: `this.props.history.push('/')` to go back to the organisation home page.

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -31,23 +69,37 @@ class EventConfigComponent extends Component {
...
       });
-    });
+      }
+    }
+    else {
+      console.log('invalid');
+      //TODO: this doesn't work - errors not showing
+      return <div>
+        {this.state.errors && this.state.showErrors && this.getErrorMessages(this.state.errors)}
+        </div>
```
avishkar58, line 101: You should move this block into your render function, maybe just above your submit button, then it will automatically render when this.state.errors becomes non-empty and this.state.showErrors becomes true.

danielamassiceti: Done

danielamassiceti:
Hi Avishkar
I think I'm almost there! A couple of questions:
1. I have added 'live' validation (i.e. the error boxes appear at the bottom of the page as the user types) if the user tries to submit once but receives errors. I thought this is better than leaving all the errors visible there until they've clicked submit again. One problem, however, is that it seems there is a 'delay': e.g. if I edit the org email to be invalid, the relevant error appears below, but if I fix it again, the error doesn't go away until i've edited _another_ field. Any ideas why?
2. Referencing our earlier thread, I haven't been able to get the asterisks appearing next to the required fields. On that note, are all the fields required (e.g. perhaps we want to make the website optional)?
3. I am having some challenges with the times of the dates. I automatically set the times of the closing dates to be 23:59:59Z and the times of the opening dates to be 00:00:00Z. This works okay if you're in UTC, but if you're significantly ahead/behind UTC (i.e. me) it can mean situations arise where you pick say 28 Feb in the date picker, but it displays 27 Feb (i.e. if your current timezone is on the cusp of a UTC day change). It's a bit of an edge case, but wondering the best way to fix

avishkar58:
Hey @danielamassiceti,

This is looking really GREAT - Thank you!

Regarding your questions:
1. Pretty sure this is a race condition because you call `this.validateEventDetails()` after setState, but setState is an async operation, so validateEventDetails is getting run before the state has actually updated (so it's validating the old updatedEvent, not the new one). You'll need to put validateEventDetails inside the setState callback, where you have your call to `this.areAllFieldsComplete()`  :) 
2. [Looks like they're all required](https://github.com/deep-learning-indaba/Baobab/blob/a3c7ecd7357afd3ca715150913a439beaab766bb/api/app/events/models.py#L20) except for miniconf_url (and strangely key is also set to nullable, but that is definitely wrong!) 
3. Ahh dates are so annoying haha. ummm.. looks like the DatePicker has a `setDefaultLocale` method, maybe we should just set that to UTC? Actually don't even know if there is such a thing as a "UTC locale"? so maybe en-GB ?  Let me think about this some more...

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -137,275 +370,378 @@ class EventConfigComponent extends Component {
...
+                  } />
               </div>
 
             </div>
+
             <div className={"form-group row"}>
               <label
                 className={"col-sm-2 col-form-label"}
                 htmlFor="review_open">
+                {<span className="required-indicator">*</span>}
```
avishkar58, line 577: Is this working now? I think i remember you leaving a comment saying it wasn't working, but I can't find it... (Github's PR commenting interface is confusing!)

avishkar58: Also you don't need the squigly brackets around the span :)

danielamassiceti: Yes, I worked out how to do it :)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -22,78 +55,277 @@ class EventConfigComponent extends Component {
...
+    
+    // if user has clicked submit and gotten errors, run validation as they're editing so as to update error displays
+    if (this.state.showErrors) {
+      this.validateEventDetails();
+    }
   };
 
-  updateDateTimeEventDetails = (fieldName, value) => {
-    let u = {
+  updateDropDownEventDetails = (fieldName, dropdown) => {
```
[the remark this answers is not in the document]
danielamassiceti: Done!

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -22,78 +55,277 @@ class EventConfigComponent extends Component {
...
-    });
+  onClickCreate = () => {
+    const isValid = this.validateEventDetails();
+    if (isValid) {
+      eventService.create(this.state.updatedEvent).then(result => {
+        this.setState({
+          updatedEvent: result.event,
+          error: result.error
+        });
+        this.props.history.push('/');
```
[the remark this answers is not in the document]
danielamassiceti: Hmm, it seems that error is returned as 'undefined' even if there is an error. Only statusCode seems to contain anything meaningful - could this be an API/services bug? I have attempted to parse the statusCode, but my current solution feels very hacky, so would welcome comments. In addition, even though I add 'Event key already exists' in errors in setState, the error doesn't actually appear on the page, potentially leaving the user rather confused as to why their event creation failed.

danielamassiceti:
A few comments:

1.
"Pretty sure this is a race condition because you call this.validateEventDetails() after setState, but setState is an async operation, so validateEventDetails is getting run before the state has actually updated (so it's validating the old updatedEvent, not the new one). You'll need to put validateEventDetails inside the setState callback, where you have your call to this.areAllFieldsComplete() :)"

I have attempted a solution, but am still getting a race condition I think - I enter all the fields, but it only becomes valid on the last field entry so requires one final form edit (of any kind) before the state's allFieldsComplete is updated to true. Not sure what I've doing wrong here

2. As per my comment above, I am currently parsing the statusCode returned from the POST/PUT in a very hacky way. Would welcome suggestions

webapp/src/pages/eventConfig/components/EventConfigComponent.js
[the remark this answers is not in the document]
avishkar58: Oh interesting! I had a quick look and it seems at first glance to be following the usual pattern which we use on the other pages (and which usually works). Can you use the Network tab to see exactly what JSON gets returned by the backend when there is an error and compare that to what the event service is trying to extract? (https://github.com/deep-learning-indaba/Baobab/blob/a3c7ecd7357afd3ca715150913a439beaab766bb/webapp/src/services/events/events.service.js#L69)  I can have a closer look tonight

avishkar58:
Ah I see, so now you you've fixed the race condition, but also you're deterministically accessing the old state ;) This is because in this code:
```
this.setState({
      updatedEvent: event,
      allFieldsComplete: this.areAllFieldsComplete(),
      isValid: this.validateEventDetails()
    });
```
both `areAllFieldsComplete` and `validateEventDetails` are accessing this.state.updatedEvent, and so these are getting evaluated and accessing the state *before* the call to setState is setting updatedEvent to the new event. You need to do something like this instead:

```
this.setState({
      updatedEvent: event,
    }, () => {
        // The code inside this callback is guaranteed to execute *after* this.state.updatedEvent gets updated
        this.setState({
            allFieldsComplete: this.areAllFieldsComplete(),
            isValid: this.validateEventDetails()
        });
});
```
Also just a nit, but it's a little weird that validateEventDetails is setting state and also returning a result that you use to update the state. I'd instead take out the `this.setState` at the end of validateEventDetails and instead return the `errors` array, then set both errors and isValid in one go inside updateEventState like this:

```
this.setState({
      updatedEvent: event,
    }, () => {
        // The code inside this callback is guaranteed to execute *after* this.state.updatedEvent gets updated
        const errors = this.validateEventDetails();
        this.setState({
            allFieldsComplete: this.areAllFieldsComplete(),
            errors: errors,
            isValid: errors.length === 0
        });
});
```
Then you can also simplify validateEventDetails a little by removing the all the setting of isValid :) 

Hope this helps!

webapp/src/pages/eventConfig/components/EventConfigComponent.js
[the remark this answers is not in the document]
avishkar58: So I had a closer look and found the issue - it's a bug in the backend. In the [EventAPI](https://github.com/deep-learning-indaba/Baobab/blob/a3c7ecd7357afd3ca715150913a439beaab766bb/api/app/events/api.py#L162) you'll see that all the methods get, post and put are decorated with `@marshal_with(event_fields)`, unfortunately and annoyingly, what this does is try to marshal *all* responses from the method with event_fields, whether they are errors or not (we actually have this bug in quite a few of our APIs :( ). What you can do is remove the `marshal_with` decorators and instead replace the last lines where it has something like `return event, 201` with: `return marshal(event, event_fields), 200`   (see an example in another API [here](https://github.com/deep-learning-indaba/Baobab/blob/a3c7ecd7357afd3ca715150913a439beaab766bb/api/app/attendance/api.py#L130)) - you'll also need to add `marshal` to the imports from `flask_restful` in line 6.

avishkar58:
Oh one more thing I noticed is that CALL event types don't have Offer or Registration stages, so we probably want to ignore the validations of those fields when the event_type is CALL. The backend still expects the dates to be set (something to fix in future), but we should allow people to set those to some random dates in the future or something. (or event better hide the fields and automatically set the dates to something... wdyt?)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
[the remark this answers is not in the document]
danielamassiceti: Thanks for this clarification. I am not exactly sure what marshal_with decorators are used for, but will investigate further and report back here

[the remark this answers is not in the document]
avishkar58: Doh! I forgot to say what marshal with does haha sorry! 🤦‍♂️  Our rest APIs need to return JSON but you'll see that we actually return an event object in the event API. So all the marshalling does is convert that object to JSON using the mapping between object properties and JSON keys specified in event_fields. The marshal_with decorator just says "convert anything returned by this function to JSON using this mapping". In fact it's not really providing much benefit for us because it's just as much work to construct a dictionary and return that instead, and dictionaries of course map very easily to JSON 🤷‍♂️

avishkar58:
I guess you'll know the awards requirement better than me :) Would you do "offers" for an award?  
Your suggestion for Calls sound good
And programme is a new category we created for the AIMS SA AI for Science programme - so I think that would have all the same stages as an event.. 
I was also thinking that some of the validation on the dates might be a bit too strict, for example: registration can open before offers close - this is what we do for the Indaba, where people can start registering as soon as they get their offers, but we do offers in waves so there is some overlap. Similarly, we do selection in stages based on whether people accept their offers or not, so selection and offers can also overlap. It's fine for applications and review and selection though - because those have to close before the next stage starts. So TL;DR I think we should allow selection, offers and registration to overlap :) 

By the way, just flagging in case there's some overlap in code you're editing, that Shreya is also adding new categories for journals (see [here](https://docs.google.com/document/d/1o7qianU1br85jGcclWtqOHLlTG8n4Zdv6z2lMawmSHo/edit#) for more info).

avishkar58:
One place where it will overlap which will also be quite useful for you, is that she's making the various dates optional/nullable in the backend for Events (because we're adding a new continuous_journal event type which can have all the stages running in parallel)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -271,12 +289,102 @@ class EventConfigComponent extends Component {
...
+    console.log(this.state.updatedEvent);
   }
 
   getFieldNameWithLanguage = (input, lang) => {
     return input + " in " + lang;
   }
 
+  setRequiredDateFields = (event_type) => {
+    //could shift this const globally?
+    const requiredDateFields_by_event = {
```
danielamassiceti, line 301: not sure if this const should be made global?

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -271,12 +289,102 @@ class EventConfigComponent extends Component {
...
+      requiredDateFields: requiredDateFields,
+      //sets all unrequired dates to a future date, and all required dates to ''
+      ...ALL_DATE_FIELDS.forEach(date => {
+        const updated_date = !requiredDateFields.includes(date) ? this.formatDate(future_date) : ""; //ensures all dates are reset each time user changes dropdown menu
+          this.setState(
+            {...this.state.updatedEvent,
+              [date]: updated_date
+            }
+          )
+      })
```
danielamassiceti, line 320: This is not working. I tried many variations, but the date fields are just not updated - any ideas?

danielamassiceti: Alright, I finally got this working (see latest commit) though not sure if this is the best place for the call - let me know if you'd recommend another way :)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
avishkar58: Yeah would make sense to move it up to the top of the file :)

webapp/src/pages/eventConfig/components/EventConfigComponent.js
```diff
@@ -302,21 +306,22 @@ class EventConfigComponent extends Component {
       "Call": ALL_DATE_FIELDS.slice(0, 3)
     }
     const requiredDateFields = requiredDateFields_by_event[event_type];
-    const future_date = new Date();
-    future_date.setFullYear(2099);
+
```
avishkar58, line 309: This is awesome! Some suggestions to improve the readability a bit: firstly it makes sense as you suggested in your comment to move this const to where ALL_DATE_FIELDS is defined. Secondly the `ALL_DATE_FIELDS.slice(0, 4)` while it works well, is not as readable as being more explicit about it. I'd consider adding more constants for your date pairs and then do something like this, which is less error prone (in fact it looks like you forgot to add the start and end dates in your code?) 

```
const ALL_DATE_FIELDS = [APPLICATION_DATES, REVIEW_DATES, SELECTION_DATES, OFFER_DATES, REGISTRATION_DATES, EVENT_DATES];

const REQUIRED_DATES_BY_EVENT_TYPE = {
      "Event": ALL_DATE_FIELDS,
      "Programme": ALL_DATE_FIELDS,
      "Award": [APPLICATION_DATES, REVIEW_DATES, SELECTION_DATES, OFFER_DATES, EVENT_DATES],
      "Call": [APPLICATION_DATES, REVIEW_DATES, SELECTION_DATES, EVENT_DATES]
    }
```

danielamassiceti:
@avishkar58 - I think I have finished this PR! 

The final change I made was with the timezone in the DateTimePickers. I couldn't find a way to manually set their display/value time zone (it seems this is a moment-timezone rather than a moment functionality). Rather than mess with the FormDate.js, which is used by a bunch of other pages, I instead just modified how they are rendered/captured in the EventConfigComponent file. 

Specifically, I have switched to storing the dates as just yyyy-MM-dd in this.state.updatedEvent, and only add in the time + timezone on submission to the server (see L116 and L138). Rather than formally set the state at this late point, I have instead just modify the event and send a copy to the server - not sure if this is bad practice. Can switch to a setState if needs be.

Similarly, when we get an event back from the server (i.e. in didComponentMount) I run it through a function which strips the time away again (ps: I don't think the server returns the "Z" anyway). 

With your final lookover, I think I might be able to close this PR! Thanks for all your feedback along the way :D

avishkar58:
> @avishkar58 - I think I have finished this PR!
> 
> The final change I made was with the timezone in the DateTimePickers. I couldn't find a way to manually set their display/value time zone (it seems this is a moment-timezone rather than a moment functionality). Rather than mess with the FormDate.js, which is used by a bunch of other pages, I instead just modified how they are rendered/captured in the EventConfigComponent file.
> 
> Specifically, I have switched to storing the dates as just yyyy-MM-dd in this.state.updatedEvent, and only add in the time + timezone on submission to the server (see L116 and L138). Rather than formally set the state at this late point, I have instead just modify the event and send a copy to the server - not sure if this is bad practice. Can switch to a setState if needs be.
> 
> Similarly, when we get an event back from the server (i.e. in didComponentMount) I run it through a function which strips the time away again (ps: I don't think the server returns the "Z" anyway).
> 
> With your final lookover, I think I might be able to close this PR! Thanks for all your feedback along the way :D

Oh amazing!! Thanks for all the hard work @danielamassiceti :) Will give this a quick test tomorrow. 
Are there any issues in particular you'd like to tackle next? (or if you want a break that's cool too!)

danielamassiceti:
I am happy to take on another task - I will take your lead on whatever is highest priority. I feel I have a better understanding of the front/back-end now, so can hopefully contribute on either end :)

avishkar58:
Looking GREAT @danielamassiceti :) Just did a little testing and noticed that:
- The email from field seems to default to the text "None" instead of being blank with a placeholder
- If I change the event type when editing an existing event then the dates all reset even the ones which overlap - ideally would be nice to remember those? wdyt?
Otherwise this is REALLY COOL and is going to save me so much time editing application closing dates in the database and make our users happier because they don't have to nag me to do it!! :) :) :) :)

avishkar58:
Was going to approve this PR but turns out i'm the requester and you are the reviewer, haha. so please go ahead and approve it!

danielamassiceti:
@avishkar58 
- Changed the code so that the dates persist if the event type changes
- The "None" email comes from the email=None for the deeplearningindaba org in the database. In the case where the email is actually set in the database, it will appear here - is this okay? :)

danielamassiceti (approved):

avishkar58:
Ohhh right of course, Silly me! that's perfect then thanks!

PR #1088, opened 2023-02-22
added new journal event types

### LINKED ISSUE #1085: Add Journal Events to model
Add new CONTINUOUS_JOURNAL and JOURNAL event type to the event model

shreyalpandit:
Adds new journal event types addressing Issue #1085 
Closes #1085


api/migrations/versions/349da0b8780a_.py
```diff
@@ -0,0 +1,27 @@
+"""empty message
```
avishkar58, line 1: Please add a description for the migration

avishkar58: Also please update the filename with some descriptive text after the underscore, so something like
349da0b8780a_journal_event_types.py

webapp/src/pages/home/Home.js
```diff
@@ -14,6 +14,10 @@ class Home extends Component {
         this.state = {
             upcomingEvents: null,
             awards: null,
+            journals: null,
+            continuous_journals: null,
+            journals: null,
```
avishkar58, line 19: duplicates?

avishkar58 (changes requested):

avishkar58 (approved):

Issue #1095, opened 2023-03-06
Journal Submission: Email Admin


shreyalpandit:
Update the response API to email all people with event_admin role on the CONTINUOUS_JOURNAL event whenever there is a submission (with is_submitted = True)

PR #1096, opened 2023-03-06
Email admin about journal response

shreyalpandit:
Addresses Issue #1095

### DIFF
--- a/api/app/events/repository.py
+++ b/api/app/events/repository.py
@@ -74,9 +74,10 @@ class EventRepository(BaseRepository):
         )
 
     @staticmethod
-    def get_all_admin(event_id):
+    def get_event_admins(event_id):
         return (
-            db.session.query(EventRole).get(event_id)
-            .filter_by(role='admin')\
+            db.session.query(AppUser)
+            .join(EventRole, EventRole.user_id == AppUser.id)
+            .filter_by(event_id=event_id, role='admin')
             .all()
         )
--- a/api/app/events/repository.py
+++ b/api/app/events/repository.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from app import db
-from app.events.models import Event, EventFee, EventType
+from app.events.models import Event, EventFee, EventType, EventRole
 from app.organisation.models import Organisation
 from app.responses.models import Response
 from app.applicationModel.models import ApplicationForm
@@ -70,5 +70,13 @@ class EventRepository(BaseRepository):
                 EventFee.event_id==event_id,
                 EventFee.id.in_(event_fee_ids),
                 EventFee.is_active==True)
+            .all()
+        )
+
+    @staticmethod
+    def get_all_admin(event_id):
+        return (
+            db.session.query(EventRole).get(event_id)
+            .filter_by(role='admin')\
             .all()
         )
--- a/api/app/responses/api.py
+++ b/api/app/responses/api.py
@@ -109,9 +109,9 @@ class ResponseAPI(ResponseMixin, restful.Resource):
                 
                 event = event_repository.get_event_by_response_id(response.id)
                 if event.event_type == EventType.CONTINUOUS_JOURNAL:
-                    all_admin_event_roles = event_repository.get_event_admins(event_id=event.id)
-                    for admin_event_role in all_admin_event_roles:
-                        self.send_confirmation(admin_event_role.user, response, admin_user=True)
+                    event_admins = event_repository.get_event_admins(event_id=event.id)
+                    for event_admin in event_admins:
+                        self.send_confirmation(event_admin, response)
         except:
             LOGGER.warn('Failed to send confirmation email for response with ID : {id}, but the response was submitted succesfully'.format(id=response.id))
         finally:
@@ -161,9 +161,9 @@ class ResponseAPI(ResponseMixin, restful.Resource):
                 
                 event = event_repository.get_event_by_response_id(response.id)
                 if event.event_type == EventType.CONTINUOUS_JOURNAL:
-                    all_admin_event_roles = event_repository.get_event_admins(event_id=event.id)
-                    for admin_event_role in all_admin_event_roles:
-                        self.send_confirmation(admin_event_role.user, response, admin_user=True)
+                    event_admins = event_repository.get_event_admins(event_id=event.id)
+                    for event_admin in event_admins:
+                        self.send_confirmation(event_admin, response)
         except:                
             LOGGER.warn('Failed to send confirmation email for response with ID : {id}, but the response was submitted succesfully'.format(id=response.id))
         finally:
@@ -203,7 +203,7 @@ class ResponseAPI(ResponseMixin, restful.Resource):
 
         return {}, 204
 
-    def send_confirmation(self, user, response, admin_user=None):
+    def send_confirmation(self, user, response):
         try:
             answers = response.answers
             if not answers:
@@ -227,7 +227,7 @@ class ResponseAPI(ResponseMixin, restful.Resource):
             else:
                 event_description = event.get_description('en')
 
-            if admin_user:
+            if user.is_event_admin(event.id):
                 emailer.email_user(
                 'assign-action-editor',
                 template_parameters=dict(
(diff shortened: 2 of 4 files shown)


avishkar58 (commented):
Just a quick comment I picked up, will leave the full review for @amritpurshotam Thank you!:)

api/app/responses/api.py
```diff
@@ -106,6 +106,12 @@ def post(self):
                 user = user_repository.get_by_id(user_id)
                 response = response_repository.get_by_id_and_user_id(response.id, user_id)
                 self.send_confirmation(user, response)
+                
+                event = event_repository.get_event_by_response_id(response.id)
+                if event.event_type == EventType.CONTINUOUS_JOURNAL:
+                    all_admin_users = event_repository.get_all_admin(event_id=response.id)
```
amritpurshotam, line 112: I think there's a typo here setting the `event_id` and I think also let's rather rename the variable to `all_admin_event_roles` to match the type we're working with. I know it's a little awkward since the class `EventRole` is actually an `EventUserRole` but I think it still helps with readability (especially since we do have an `AppUser` class which we generally name `users` so we need to make it clear they're different).
```python
all_admin_event_roles = event_repository.get_all_admin(event_id=event.id)
```

api/app/events/repository.py
```diff
@@ -72,3 +72,11 @@ def get_event_fees(event_id, event_fee_ids):
                 EventFee.is_active==True)
             .all()
         )
+
+    @staticmethod
+    def get_all_admin(event_id):
+        return (
+            db.session.query(EventRole).get(event_id)
```
amritpurshotam, line 79: When you use the `get` method it will query that tables primary key (in this case `EventRole.id`). It seems the intention here is to rather find all the admin `EventRoles` for a specific `event_id`. If that's the case, then the query should be something like
```python
def get_all_admin(event_id):
    return (
        db.session.query(EventRole)
        .filter_by(event_id=event_id, role='admin')\
        .all()
    )
```

api/app/responses/api.py
```diff
@@ -106,6 +106,12 @@ def post(self):
...
+                    for admin_user in all_admin_users:
+                        self.send_confirmation(admin_user, response, admin_user=True)
```
amritpurshotam, line 114: The `send_confirmation` method takes in a user object which we luckily have access to from an `EventRole`
```python
for admin_event_role in all_admin_event_roles:
    self.send_confirmation(admin_event_role.user, response, admin_user=True)
```

api/app/responses/api.py
```diff
@@ -152,6 +158,12 @@ def put(self):
                 user = user_repository.get_by_id(user_id)
                 response = response_repository.get_by_id_and_user_id(response.id, user_id)
                 self.send_confirmation(user, response)
+                
+                event = event_repository.get_event_by_response_id(response.id)
+                if event.event_type == EventType.CONTINUOUS_JOURNAL:
+                    all_admin_users = event_repository.get_all_admin(event_id=response.application_form.event_id)
+                    for admin_user in all_admin_users:
+                        self.send_confirmation(admin_user, response, admin_user=True)
```
amritpurshotam, line 166: Same as the other comment
```python
for admin_event_role in all_admin_event_roles:
    self.send_confirmation(admin_event_role.user, response, admin_user=True)
```

api/app/responses/api.py
```diff
@@ -152,6 +158,12 @@ def put(self):
                 user = user_repository.get_by_id(user_id)
                 response = response_repository.get_by_id_and_user_id(response.id, user_id)
                 self.send_confirmation(user, response)
+                
+                event = event_repository.get_event_by_response_id(response.id)
+                if event.event_type == EventType.CONTINUOUS_JOURNAL:
+                    all_admin_users = event_repository.get_all_admin(event_id=response.application_form.event_id)
```
amritpurshotam, line 164: Just the rename to `all_admin_event_roles` as before

api/migrations/versions/cebfdfef31cd_assign_action_editor_email_template.py
```diff
@@ -0,0 +1,73 @@
...
+        self.subject = subject
+        self.template = template
+        self.language = language
+
+
+def upgrade():
+    Base.metadata.bind = op.get_bind()
+    session = orm.Session(bind=Base.metadata.bind)
+
+    op.execute("""SELECT setval('email_template_id_seq', (SELECT max(id) FROM email_template));""")
```
amritpurshotam, line 47: Can you explain what this is doing?
If I had to guess you're trying to ensure the next primary key follows in sequence from the current highest. If I'm right I would say take this out (and from the downgrade) since the db should handle this on it's own (and also having gaps is fine).

amritpurshotam (changes requested):

api/app/events/repository.py
amritpurshotam: Actually I realised we can make this a bit better. Instead of querying for the `EventRole` and accessing the `user` attribute on it, let's just query for `AppUsers` directly. So on `/users/repository.py` you can do something like this. Then when sending the confirmation email, you'll just use the objects directly.

```python
def get_event_admins(event_id):
    return (
        db.session.query(AppUser)
        .join(EventRole, EventRole.user_id == AppUser.id)
        .filter_by(event_id=event_id)
        .all()
    )
```
Then

api/app/responses/api.py
```diff
@@ -106,6 +106,12 @@ def post(self):
                 user = user_repository.get_by_id(user_id)
                 response = response_repository.get_by_id_and_user_id(response.id, user_id)
                 self.send_confirmation(user, response)
+                
+                event = event_repository.get_event_by_response_id(response.id)
+                if event.event_type == EventType.CONTINUOUS_JOURNAL:
+                    all_admin_event_roles = event_repository.get_event_admins(event_id=event.id)
+                    for admin_event_role in all_admin_event_roles:
+                        self.send_confirmation(admin_event_role.user, response, admin_user=True)
```
amritpurshotam, line 114: Sorry I should have cleaned up my comments after changing my mind about how the event admins can be queried. This should now be similar to what you had before since we queried for and now get `AppUser` objects.
```python
event_admins = event_repository.get_event_admins(event_id=event.id)
for event_admin in event_admins:
    self.send_confirmation(event_admin, response, admin_user=True)
```

api/app/responses/api.py
```diff
@@ -152,6 +158,12 @@ def put(self):
...
```
amritpurshotam, line 166: Same here
```python
event_admins = event_repository.get_event_admins(event_id=event.id)
for event_admin in event_admins:
    self.send_confirmation(event_admin, response, admin_user=True)
```

api/app/responses/api.py
```diff
@@ -215,15 +227,26 @@ def send_confirmation(self, user, response):
             else:
                 event_description = event.get_description('en')
 
-            emailer.email_user(
-                'confirmation-response-call' if event.event_type == EventType.CALL else 'confirmation-response',
+            if admin_user:
```
amritpurshotam, line 230: Since we have access to the `AppUser` object (as the `user` variable here) and we know the `event` we're dealing with let's actually use both those together and then we can remove the `admin_user=None` parameter on this `send_confirmation` method.

```python
if user.is_event_admin(event.id):
```

amritpurshotam (changes requested):
Thanks for the changes and sorry for the mix-up with the earlier review comments. I think we should be good after these small updates.

amritpurshotam (approved):
Thanks! Looks good!

api/migrations/versions/cebfdfef31cd_assign_action_editor_email_template.py
```diff
@@ -0,0 +1,71 @@
...
+    
+A new application for {event_name} was received. Please assign an action editor to this application as soon as possible.
+
+"""
+
+    session.add(EmailTemplate('assign-action-editor', None, '{event_name} Response Receieved', template, 'en'))
+
+    template = """Dear {title} {firstname} {lastname},
+    
+A new application {event_name} was received but has not been assigned to an action editor yet. Please assign an action editor to this application as soon as possible.
```
avishkar58, line 57: A new application **for** {event_name}  :)

Issue #1097, opened 2023-03-06
Add Review State to Review Assignment


shreyalpandit:
Update the review assignment API to add a “review” state (via outcome) when a reviewer is assigned and the event is of type JOURNAL or CONTINUOUS_JOURNAL

PR #1100, opened 2023-03-16
Added Review State

shreyalpandit:
Addresses Iss #1097

### DIFF
--- a/api/app/outcome/models.py
+++ b/api/app/outcome/models.py
@@ -6,6 +6,9 @@ class Status(Enum):
     ACCEPTED = "accepted"
     REJECTED = "rejected"
     WAITLIST = "waitlist"
+    REVIEW = "in review"
+    ACCEPT_W_REVISION = "accept with minor revision"
+    REJECT_W_ENCOURAGEMENT = "reject with encouragement to resubmit"
 
 class Outcome(db.Model):
     id = db.Column(db.Integer(), primary_key = True, nullable = False)
--- a/api/app/reviews/api.py
+++ b/api/app/reviews/api.py
@@ -10,7 +10,7 @@ from datetime import datetime
 
 from app import db, LOGGER
 from app.applicationModel.models import ApplicationForm
-from app.events.models import Event, EventRole
+from app.events.models import Event, EventRole, EventType
 from app.events.repository import EventRepository as event_repository
 from app.responses.repository import ResponseRepository as response_repository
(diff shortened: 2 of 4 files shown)


amritpurshotam:
@shreyalpandit I'm not quite convinced about this PR. Having an internal reviewer assigned isn't quite an outcome in my opinion and seems to go against what's happening currently for applications.

The way I see it if we don't add the REVIEW status, we would still be able to tell which submissions are being reviewed by seeing which ones have an assigned reviewer but no review responses and/or an outcome.

Of course I could be missing something obvious. What do you think?

shreyalpandit:
That makes sense. I was thinking 1) for notifications and 2) for iterations on the submission. 1) It would be helpful to track and remind reviewers to submit responses since there are no deadlines for the continuous journal. However, it seems like this can also be checked by looking at reviewers and outcome for a response. 2) When an applicant gets feedback and then submits a new response, they would no longer be considered the previous outcome but rather, be back in review. I guess this might not apply if the outcome is connected to each individual response...in which case we can still check for assigned reviewers and no outcome for submissions in review.

@avishkar58 had also suggested this addition to outcomes, so I am not sure if he had something else in mind.

amritpurshotam:
@shreyalpandit Yeah you make good points and are definitely quite valid. This approach then would indeed solve this. Give me one more day to ponder on this one though.

We do have a concept of a `active` reviewer on the `ResponseReviewer` class which we don't seem to be using from a quick glance. Just want to think about if this could be what we need. So no need for action on your side. will get back to this tomorrow.

avishkar58:
Hey, just jumping in to give my 2 cents here. I think I may have suggested
adding a review outcome, since continuous journals don't have time
delineated phases, but instead each application can be in a different phase
at different times. The review outcome would then be to indicate to both
the admins and the applicant that their submission is now in the review
phase, even though reviewers might not yet have been assigned. It's
possible and maybe even preferable to derive that from it being submitted
and not having another outcome, but I thought it was better to be explicit.
Wdyt Amrit and Shreya?

amritpurshotam:
@shreyalpandit @avishkar58 Hey sorry I dropped off on this for a bit.

Yeah I agree being explicit makes more sense. With the implicit approach it could get a bit muddled especially with the bank and forth of the review process. I think I was getting too caught up in the "Outcome" wording which felt like it should be the end of the process (but it doesn't have to be).

amritpurshotam (approved):

Issue #1106, opened 2023-03-21
Add Outcome to Response


shreyalpandit:
Extend the response page to support adding an outcome to a response, choose from “accept”, “accept with minor revision”, “reject with encouragement to resubmit”,  or “reject” for event types JOURNAL or CONTINUOUS_JOURNAL;  “accept” or “reject” for CALL;  “waitlist”, “accept” or “reject” for EVENT

PR #1107, opened 2023-03-21
Assign Response Outcome

shreyalpandit:
Addresses Issue #1106 

Provides event admin with option to select an outcome once in review:

Before outcome is selected:
Screen Shot 2023-03-21 at 11 24 22 AM

After:
Screen Shot 2023-03-21 at 9 44 56 AM


api/app/outcome/api.py
```diff
@@ -51,9 +51,10 @@ class OutcomeAPI(restful.Resource):
     def get(self):
         req_parser = reqparse.RequestParser()
         req_parser.add_argument('event_id', type=int, required=True)
+        req_parser.add_argument('user_id', type=int, required=True)
         args = req_parser.parse_args()
 
-        user_id = g.current_user['id']
+        user_id = args['user_id']
```
amritpurshotam, line 57: Is it an event admin that you're intending on using this API endpoint?

The reason we would have auto-set the `user_id` is likely because this is the endpoint that it's intended for the logged in user (in this context an applicant) to find out their outcome (and _only_ their outcome i.e. it's insecure to allow any user to be able query any other users outcome).

If indeed you were intending for the event admin to be able to query any outcome (from a user that's part of their event) then we would need to create another API endpoint (`OutcomeAdminAPI` maybe?) with a `get` method and the `@event_admin_required` decorator like what's on the `post` method. Let me know if you need help with this

shreyalpandit: Yes, that makes sense! I was thinking this would be for the the event admin specifically and can make the change. I didn't see this endpoint being used elsewhere and was thinking of just updating it so that it's only for the admin. Would this make sense or should I still keep the regular endpoint and add the admin specific api.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -106,6 +110,162 @@ class ResponsePage extends Component {
...
+            }
+        });
+    }
+
+    outcomeStatus() {
+        const data = this.state.applicationData;
+        
+        if (data) {
+
+            this.getOutcome();
```
avishkar58, line 162: Note that setState calls in React are aync, so there's no guarantee that the code you have here after calling `getOutcome` will run before or after `getOutcome` actually updates the outcome in `this.state`. In order to ensure that code which depends on some state being set runs only after that state has been set, you need to use the callback in `setState`. (See [this explanation](https://stackoverflow.com/a/42038724) on StackOverflow for more info)

avishkar58: Also from the chain of calls it looks like getOutcome is effectively being called inside the render() function? that's not a good idea because render gets called every time state is updated, so you could get an infinite loop unless you do something like what you're doing with checkOutcome. But that should not be needed. Instead it's better to get the outcome in `componentDidMount` where all the other API calls are and then you won't need the checkOutcome either.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -106,6 +110,162 @@ class ResponsePage extends Component {
...
+                </div>
+            }
+            else if (this.state.event_type === 'CALL') {
+                return <div className='user-details'>
+                    <div className="user-details">
+                        <button
+                            type="button"
+                            class="btn btn-success"
+                            id="accept"
+                            onClick={(e) => this.goBack(e)}>
```
avishkar58, line 220: Why are these all going Back? (Rather leave out the buttons if they're not functional otherwise it will be very confusing for users). Anyway, may as well make them work because the logic is mostly the same as what you wrote for journals :) For calls, accept and reject can do the same as your journal one. For events, reject and waitlist will be the same, but accept will be a bit different, so you can leave that out and just remove the accept button for events.

avishkar58 (changes requested):
Thanks Shreya! :)

webapp/src/pages/ResponsePage/ResponsePage.js
shreyalpandit: My bad, I put those as placeholders to go back and make them functional! Will fix this, thanks!

amritpurshotam (approved):
Looks good!

api/app/outcome/api.py
```diff
@@ -46,14 +46,15 @@ def _extract_status(outcome):
 
 
 class OutcomeAPI(restful.Resource):
-    @auth_required
+    @event_admin_required
     @marshal_with(outcome_fields)
-    def get(self):
+    def get(self, event_id):
         req_parser = reqparse.RequestParser()
         req_parser.add_argument('event_id', type=int, required=True)
```
avishkar58, line 53: You don't need this line anymore since you have event_id already passed in as a paramter.

api/app/outcome/api.py
```diff
@@ -74,6 +75,7 @@ def get(self):
     @marshal_with(outcome_fields)
     def post(self, event_id):
         req_parser = reqparse.RequestParser()
+        req_parser.add_argument('event_id', type=int, required=True)
```
avishkar58, line 78: Also not needed since you already have event_id

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -44,13 +46,14 @@ class ResponsePage extends Component {
...
                 eventLanguages: responses[0].event ? Object.keys(responses[0].event.name) : null,
+                event_type: responses[0].event.event_type,
                 applicationForm: responses[1].formSpec,
                 applicationData: responses[2].detail,
                 tagList: responses[3].tags,
                 reviewers: responses[4].reviewers,
                 error: responses[0].error || responses[1].error || responses[2].error || responses[3].error || responses[4].error,
-                isLoading: false
             }, this.handleData);
+            this.getOutcome();
```
avishkar58, line 56: You'll still have the same issue here where you're trying to access this.state.applicationData inside getOutcome with no guarantee that it has been set yet by setState. So you'll need to call this where this.handleData is being called. You can update line 55 to be something like:
`}, ()=> {this.handleData(); this.getOutcome();}`  (with better formatting of course :) )

avishkar58 (changes requested):

avishkar58 (approved):

Issue #1086, opened 2023-02-22
Update application form to allow multiple submissions


shreyalpandit:
Change the current “nominations” functionality in the application form to be a more generic “allow multiple submissions”

PR #1108, opened 2023-03-23
Nomination Wording to Submission for Journals

shreyalpandit:
Addresses Issue #1086 

**Event Type = Event**
Screen Shot 2023-03-24 at 12 17 27 AM
**Event Type = Journal**
Screen Shot 2023-03-24 at 12 17 08 AM


webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -1009,14 +1010,22 @@ class ApplicationListComponent extends Component {
...
+
   getCandidate = (allQuestions, response) => {
     const nominating_capacity = answerByQuestionKey("nominating_capacity", allQuestions, response.answers);
     if (nominating_capacity === "other") {
       let firstname = answerByQuestionKey("nomination_firstname", allQuestions, response.answers);
       let lastname = answerByQuestionKey("nomination_lastname", allQuestions, response.answers);
       return firstname + " " + lastname;
     }
-    return this.props.t("Self Nomination");
+    return  (this.state.event_type ==='JOURNAL' || this.state.event_type ==='CONTINUOUS_JOURNAL') ? this.props.t("Submission " + response.id) : this.props.t("Self Nomination");
```
avishkar58, line 1028: This will give you errors because we won't find the text "Submission + arbitrary numbers" in the translation.json keys (have a look in the Console output and you should see them). You'll need to rather do `this.props.t("Submission") + " " + response.id`  (ie only translate the "Submission" text). Also please add this and any new text you've added on the UI to the two translation.json files for English and French.

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -1039,12 +1048,14 @@ class ApplicationListComponent extends Component {
 
   render() {
     let allQuestions = _.flatMap(this.props.formSpec.sections, s => s.questions);
+    let title = (this.state.event_type ==='JOURNAL' || this.state.event_type ==='CONTINUOUS_JOURNAL') ? this.props.t("Your Submissions") : this.props.t("Your Nominations");
```
avishkar58, line 1051: Nit: should be `const`, not `let` because it's not going to change.

avishkar58 (changes requested):
Very nice, thanks Shreya! Really cool that you were able to reuse a lot of the existing functionality to accomplish this :)

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -1009,14 +1010,22 @@ class ApplicationListComponent extends Component {
     }
   }
 
+  componentDidMount() {
+    eventService.getEvent(this.props.formSpec.event_id).then(response => {
```
avishkar58, line 1014: We do pass event as a prop to the ApplicationForm (from [EventHome.js](https://github.com/deep-learning-indaba/Baobab/blob/c4dd0240a4182ad8980a8abd2c9b38ccd7c7b2cc/webapp/src/pages/eventHome/EventHome.js#L153)). Can we not just get event_type from there instead of having to do an extra call to the backend?

avishkar58 (approved):

PR #1109, opened 2023-03-26
Updating grants to work with tags.

danielamassiceti:
Hi @avishkar58. After reading your doc and familiarising myself with the various files that will be involved in this PR, I've made an attempt at a set of detailed steps. Let me know if I am on the right track (I've got a few questions here and there), and I will then get started!

Steps:

1. Update Tag and TagTranslation models in tags/model.py - DONE
- Add tag type to Tag model - set by admins, enum with option 'registered' (possible to add more later)
- Add tag description - captured by admins on the front-end when they add a new tag
- Create migration to update database with new models
- Update/add all necessary tests

2. Update the Tag API in tags/api.py - DONE
- Update _serialize_tag_detail and _serialize_tag to work with the new fields
- Update TagAPI's get, post, and put to work with the new fields
- Update/add all necessary tests

3. Apply Tags to Offer and InvitedGuest in back-end
- Create class OfferTag in offer.model.py -- follow same template as ResponseTag
- Create class InvitedGuestTag in invitedGuest/model.py -- follow same template as ResponseTag
- Create migration to update database with new models
- Add all necessary tests

4. Add Configure Tabs page on front-end
- Functionality to add, delete, and update tags (type, name, description). Accessible via 'Configure Tags' in Admin drop-down menu

5. Update front-end to work with new tag models
- Add tag adding functionality to Invited Guest page (similar to ResponePage)
- Add tags to the Offer POST API (since no UI for Offers yet)

6. Clean-up Offer back-end (model and API) - in api/app/registration/model.py + /api.py
- Remove travel_award, accommodation_award, accepted_accommodation_award, accepted_travel_award from the offer model (registration/models.py) + create migration to update the database with the new model.
- Remove all award stuff from Offer API. Functions to update: offer_info, update_offer_info, OfferAPI (put, post, get),
+ Migrate all data from old events into new format
- Update Offer front-end (pages/offer/components/OfferComponent.js) remove all UI for travel_award, accommodation_award and accepted versions
- Add tag functionality to Offer API

7. Clean-up RegistrationSection (model and API) - in api/app/registration/model.py + /api.py
- Remove conditional display stuff for awards in models.py
- Replace with 'show_for_tag' logic in models.py -- needs to work for selected candidates and guests who have corresponding tag
- Update corresponding API - RegistrationAPI (get), RegistrationSectionAPI (registration_section_info, registration_section_fields, post)


webapp/src/App.js
```diff
@@ -168,6 +169,13 @@ class EventNav extends Component {
                   >
                     {t('Review Dashboard')}
                   </NavLink>
+                  <NavLink
+                    to={`/${this.props.eventKey}/tagConfig`}
```
avishkar58, line 173: Ah this will try to navigate to something like /indaba2023/tagConfig, but your route below is just defined for /tagConfig (ie at the organisation level). Since we do want this page to be at the event level, you should define the route in [EventHome.js](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/src/pages/eventHome/EventHome.js#L128) where the `${match.path}` in the path will match the event key :)

api/app/invitedGuest/models.py
```diff
@@ -11,12 +10,26 @@ class InvitedGuest(db.Model):
...
         self.event_id = event_id
         self.user_id = user_id
         self.role = role
 
+class InvitedGuestTag(db.Model):
+    id = db.Column(db.Integer(), primary_key=True)
+    user_id = db.Column(db.Integer(), db.ForeignKey('app_user.id'), nullable=False)
+    tag_id = db.Column(db.Integer(), db.ForeignKey('tag.id'), nullable=False)
+
+    invited_guest = db.relationship('InvitedGuest', foreign_keys=[user_id])
```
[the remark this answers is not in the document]
danielamassiceti: Resolved - thank you!

webapp/src/App.js
danielamassiceti: Resolved, thanks!

api/app/invitedGuest/tests.py
```diff
@@ -133,3 +133,126 @@ def test_create_invitedGuest_list(self):
...
+        self.assertEqual(response.status_code, 201)
+
+        params = {
+            'event_id': self.event1.id,
+            'invited_guest_id' : self.invited_guest1.id,
+            'language': 'en',
+            'include_unsubmitted': False
+        }
+
+        response = self.app.get(
```
danielamassiceti, line 182: This get call is causing the test to fail. The error message is : {'message': {'email': 'Missing required parameter in the JSON body or the post body or the query string'}}. Any ideas?

api/app/invitedGuest/tests.py
```diff
@@ -133,3 +133,126 @@ def test_create_invitedGuest_list(self):
...
+        self.assertEqual(response.status_code, 200)
+
+        params = {
+            'event_id': self.event1.id,
+            'invited_guest_id' : self.invited_guest1.id,
+            'language': 'en',
+            'include_unsubmitted': False
+        }
+
+        response = self.app.get(
```
danielamassiceti, line 217: Same error happening here

api/app/invitedGuest/tests.py
avishkar58: Yeah Get requests don't seem to work with the json property for params (not entirely sure why), so we usually do it in the query string like this: https://github.com/deep-learning-indaba/Baobab/blob/develop/api/app/invitedGuest/tests.py#L127

api/app/registration/tests.py
```diff
@@ -201,10 +201,157 @@ def test_update_offer(self):
...
+        db.session.commit()
+
+        self.headers = self.get_auth_header_for("something@email.com")
+        self.adminHeaders = self.get_auth_header_for("offer_admin@ea.com")
+
+        self.add_email_template('offer')
+
+        db.session.flush()
+
+    def test_tag_admin(self):
```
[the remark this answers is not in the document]
avishkar58: Ah because self.offer.id will only be set when it gets added to the database, which happens after db.session.add(self.offer) and db.session.commit(). But in line 238 you're trying to access self.offer.id before the offer object has been  committed to the database, so you will need to rearrange those lines to commit the offer first and then call tag_offer. I think there's also another way to do it by setting the tag objects on the offer object's tags property (so don't use the ids) and then you can commit the entire object graph in one go, but I've usually just done it step by step, create object, commit it, add other stuff, commit etc.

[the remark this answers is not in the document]
danielamassiceti: Ah I see. Thought when I do shift it afterwards, I get this message instead now:
sqlalchemy.orm.exc.DetachedInstanceError: Instance <InvitedGuest at 0x7f073d967b10> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/13/bhk3)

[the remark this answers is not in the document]
avishkar58: Oh no the dreaded DetachedInstanceError! :see_no_evil: 
This is a weird one, Amrit and I have been trying to figure it out for ages now, think we've concluded that it's some bug in SqlAlchemy or the way we're using it. Anyway, it only seems to happen sometimes, but if it does happen it is usually when you try to access an object property after doing an API call (which you're doing in some of your tests where you have a POST, then after the POST you're accessing self.offer.id and self.event1.id before doing a GET). It seems that something about the API call is causing these objects to become detached from their sessions. Anyway, the workaround which you'll see in many of our tests is to instead cache the properties that you want to use in the seed_static_data function, so for example [in this test](https://github.com/deep-learning-indaba/Baobab/blob/91b49ab42323ce480204a3bf27a27768b654d0aa/api/app/outcome/tests.py#L49) you'll see that we cache the IDs and use those instead of calling `self.event1_user1_outcome1.id`

api/app/registration/tests.py
```diff
@@ -204,130 +204,119 @@ def test_update_offer(self):
...
         params = {
-            'event_id': self.event1.id,
-            'offer_id' : self.offer.id,
+            'event_id': self.event.id,
+            'user_id' : self.test_user_id,
             'language': 'en',
-            'include_unsubmitted': False
         }
 
         response = self.app.get(
```
danielamassiceti, line 263: this GET is failing with the message 'No response found for the given event and user'. I think this is because the GET in the OfferAPI actually sets the user_id to g.current_user['id'] rather than using the args['user_id'] that I am passing in. Is this expected?

avishkar58: yeah this is expected because the offer API is for users to get their own offers, so it only uses the authenticated user rather than a provided user_id so that people can't look at other peoples offers. So in this case you can instead provide the authentication header for test_user: `headers=self.get_auth_header_for('test_user@mail.com')`

danielamassiceti: Resolved thanks!

webapp/src/pages/tagConfig/components/TagConfigComponent.js
```diff
@@ -86,9 +87,10 @@ class TagConfigComponent extends Component {
     }
   };
 
-  onClickDelete = () => {
-    const tag = null; //TODO
-    tagsService.deleteTag(tag).then(result => { //TODO: will fail, need to add deleteTag to tagsService
+  onClickDelete = (tag) => {
+    //TODO - debug, not working
```
danielamassiceti, line 91: This delete is not actually deleting the tag in the DB - it looks okay to me. Am I missing something?

webapp/src/pages/tagConfig/components/TagConfigComponent.js
```diff
@@ -261,12 +288,15 @@ class TagConfigComponent extends Component {
 
     tagEntryForm.push(
       <div className={"form-group row"}>
-        <button className="btn btn-primary" onClick={this.onClickAdd()}>
-          {t("Save")}
-        </button>
+        <button
+          onClick={this.onClickAdd}
+          className="btn btn-success btn-lg btn-block">
+          {t("Save Tag")}
```
danielamassiceti, line 294: I would like this button to be on the right of the screen - what's the best way to right align?

danielamassiceti:
Two questions:
- In componentDidMount in TagConfigComponent.js, I call tagsService.getTagList and then render the returned tags in the tag table. The name and descriptions of the tags are not in the 'dict' format ({'en': 'en name', 'fr': 'fr name'}) however, so I am not sure how we should deal with this. Could I change the API to return the dicts, rather than serialising them flat? Not sure if this is going to affect other places, however. I could alternatively just package them back into dicts once I've gotten them from the server. If so, how do I get the 'current' display language ?
- It seems that the OfferAPI doesn't yet have language functionality. I have added a 'tags' fields to the offer info (see offer_info in OfferAPI), however, this ideally should have a language argument too. Could I add this as a field to the Offer API? Not sure if it's going to disrupt other things

avishkar58:
> * In componentDidMount in TagConfigComponent.js, I call tagsService.getTagList and then render the returned tags in the tag table. The name and descriptions of the tags are not in the 'dict' format ({'en': 'en name', 'fr': 'fr name'}) however, so I am not sure how we should deal with this. Could I change the API to return the dicts, rather than serialising them flat? Not sure if this is going to affect other places, however. I could alternatively just package them back into dicts once I've gotten them from the server. If so, how do I get the 'current' display language ?

Do you need the tags from the tag list in dict format for an edit functionality? like to edit the name and description in all languages and then save? In that case, it probably makes sense to create a new API like `TagListConfigAPI` that returns the tags in the format you need them instead of editing the existing API. The existing API is used in places where we need a list of tags for the user to choose from and there it's better to return only the active language rather than wastefully returning data for all languages and just selecting one.  

> * It seems that the OfferAPI doesn't yet have language functionality. I have added a 'tags' fields to the offer info (see offer_info in OfferAPI), however, this ideally should have a language argument too. Could I add this as a field to the Offer API? Not sure if it's going to disrupt other things

Yes you can add this as a field in the Offer API, in fact our front-end always sends the language parameter with every request (see [here](https://github.com/deep-learning-indaba/Baobab/blob/55e90586491825689a69a1cbf654dbfdf297d8f7/webapp/src/services/base.service.js#L52) for the sneakiness) so you can always add a parameter to extract it in any backend API if you need it :)

danielamassiceti:
> Do you need the tags from the tag list in dict format for an edit functionality? like to edit the name and description in all languages and then save? In that case, it probably makes sense to create a new API like `TagListConfigAPI` that returns the tags in the format you need them instead of editing the existing API. The existing API is used in places where we need a list of tags for the user to choose from and there it's better to return only the active language rather than wastefully returning data for all languages and just selecting one.

Yes, I think so. Similar to the event config, the admin will need to enter the English _and_ French names and descriptions of each new tag right (if there are multiple languages)? In that case, if they choose to edit a tag, it should repopulate all the dual-language fields for them to edit. In the table, perhaps just the 'active' language can be displayed (i.e. as per the drop-down language menu shown in the top right). How do I access the current active language?

danielamassiceti:
When I click 'Save tag' after entering a new tags details, I get this error: "Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops." Not sure what this means, or why I am getting it? :/

webapp/src/pages/tagConfig/components/TagConfigComponent.js
```diff
@@ -185,10 +205,7 @@ class TagConfigComponent extends Component {
             <td>{tag.tag_type}</td>
             <td>
               <button
-                onClick={() => {
-                  tag.active = false; 
-                  this.onClickUpdate(tag)
-                }}>
+                onClick={this.onClickDelete(tag)}>
```
avishkar58, line 208: This is what's causing your maximum update depth exceeded error. You're calling the event handler here instead of passing a function to onClick, I guess you accidentally forgot to make it a lambda:
`onClick={()=>this.onClickDelete(tag)}`

avishkar58:
> When I click 'Save tag' after entering a new tags details, I get this error: "Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops." Not sure what this means, or why I am getting it? :/

This happens when you have an infinite loop, something updates the state, then render is called and while rendering something updates the state again. Found the culprit and left a comment. Also just realised that I had a bunch of draft comments from last week that I forgot to send! :(  Anyway, hope they're still useful. 

By the way, have added the UI functionality for adding and removing tags to invited guests now.

webapp/src/pages/tagConfig/components/TagConfigComponent.js
```diff
@@ -0,0 +1,375 @@
...
+    if (errors.length === 0) {
+      tagsService.updateTag(this.state.updatedTag, this.props.event.id).then(result => {
+        if (result.error) {
+          this.setState({
+            errors: [this.props.t(result.error)],
+            showErrors: true
+          });
+        }
+        else {
+          //go back to tag table
```
avishkar58, line 82: You'll need to update the tags state to see the change in the table :) 
see how its done in InvitedGuestComponent line 280 for an example

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
+
+        responsesService.getResponseList(this.props.event.id, includeUnsubmitted, []).then(resp => {
+            this.setState({
+                responses: resp.responses,
+                error: resp.error
+            });
+        });
+    }
+
+    generateColumns() {
```
avishkar58, line 70: Looks like you're not using this method because it's all inside the render anyway, so can be deleted

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
+import { NavLink } from "react-router-dom";
+import { tagsService } from '../../../services/tags/tags.service';
+import FormTextBox from "../../../components/form/FormTextBox";
+import { reviewService } from '../../../services/reviews/review.service';
+import { ConfirmModal } from "react-bootstrap4-modal";
+import TagSelectorDialog from '../../../components/TagSelectorDialog';
+import FormSelect from "../../../components/form/FormSelect";
+import { createColClassName } from "../../../utils/styling/styling";
+
+class ResponseListForm extends Component {
```
avishkar58, line 16: It's just... BEAUTIFUL! 😭😭😭😭 THANK YOU!!!!

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
+          {
+            id: "submitted_date",
+            Header: <div className="response-submitted-date">{t("Submitted Date")}</div>,
+            accessor: u => u.submitted_date,
+            minWidth: 150
+          },
+          {
+            id: "reviewers",
+            Header: <div className="response-reviewers">{t("Reviewers")}</div>,
+            accessor: u => u.reviewers.map(r => r.reviewer_name).join("; "),
```
avishkar58, line 132: reviewers could be null, so you'll need to check that first before trying to map over it.

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
```
avishkar58, line 362: r could be null because the backend will always return a list of reviewers of the expected size and pad that list will nulls if there are not that many reviewers assigned. So you'll need to handle that here for example by replacing this line with
`accessor: u => u.reviewers.filter(r=>r).map(r=>r.reviewer_name).join("; "),`

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
+          }
+          if (tagSearch && passed && tagSearch !== "none") {
+            passed = r.tags.some(t => t.name === tagSearch);
+          }
+          return passed;
+        });
+        this.setState({ filteredResponses: filtered });
+      }
+
+    render() {
```
avishkar58, line 275: Some minor comments about the UI: 
- The response link column seems to wide, maybe just change the header to "ID" and make it more narrow? The fact that its blue will hopefully indicate to people that it's a link better than how it was previously displayed
- Would be great to change the cursor to pointer when you hover over the tag badges to indicate they're clickable 
- Can you get the Assign button to be not so wide? :) 
- We do need to be able to filter by multiple tags as an "AND" which the page could previously do, see how it's done in the [review assignment](https://github.com/deep-learning-indaba/Baobab/blob/d7ccb42ef6047466b0bcf64c19960e6a058670a8/webapp/src/pages/reviewAssignment/components/ReviewAssignmentComponent.js#L195) page for example. But if you want to keep the controls consistent so it looks like the other search boxes then perhaps we need a new form component like [the Multi example from React Select](https://react-select.com/home) -  and good news is that we are already using React Select so it should be easy to add this! :) 
Otherwise the new UI is beautiful!! :D :D

webapp/src/pages/ResponseList/components/ResponseListComponent.js
```diff
@@ -0,0 +1,494 @@
...
+          selectedResponse: response,
+          tagSelectorVisible: true,
+          filteredTags: this.state.tags.filter(t=>!tagIds.includes(t.id))
+        })
+    }
+
+    onSelectTag = (tag) => {
+        responsesService.tagResponse(this.state.selectedResponse.response_id, tag.id, this.props.event.id)
+        .then(resp => {
+            if (resp.statusCode === 201) {
```
avishkar58, line 191: Seems like this service calls it `status` instead of `statusCode`, so this `if` check is not passing and this is why you're not seeing the change in the table after you've added or removed a tag. An unfortunate inconsistency in our services conventions (or lack thereof)! :(

avishkar58: By the way, I figured this out by running the debugging in the Sources tab of the Chrome developer tools - not sure if you're using that tool already, but can highly recommend it if you're not!

webapp/src/pages/ResponseList/components/ResponseListComponent.js
danielamassiceti: I couldn't work out how to get the assign button to look good - any pointers? The rest, however, is done :)

webapp/src/pages/ResponseList/components/ResponseListComponent.js
danielamassiceti: Thanks - fixed! :D

api/app/registration/tests.py
```diff
@@ -150,11 +141,100 @@ def test_create_offer_with_template(self):
 
         self.assertEqual(response.status_code, 201)
         self.assertFalse(data['payment_required'])
-        self.assertFalse(data['travel_award'])
-        self.assertFalse(data['accommodation_award'])
+
+    def test_create_offer_with_grant_tags(self):
```
avishkar58, line 145: Ah I see this is the test failing with `ValueError: Could not find email template with key offer-awards`. All of our tests use a fake database that starts from nothing and we populate it with exactly the data we want for the test - this gives us maximum control about the test context, rather than relying on some "real" data which may change. This is what the `seed_static_data` methods do in the tests. So to get the email template populated for your test, you'll need to add a line to `seed_static_data`, like `self.add_email_template('offer-awards')`  (eg line 90)

api/app/registration/api.py
```diff
@@ -126,9 +114,14 @@ def put(self):
 
             offer.responded_at = datetime.now()
             offer.candidate_response = candidate_response
-            offer.accepted_accommodation_award = accepted_accommodation_award
-            offer.accepted_travel_award = accepted_travel_award
             offer.rejected_reason = rejected_reason
+            for a in accepted_awards:
+                award_id = a['id']
+                award_accepted = a['accepted']
+                offer_tag = db.session.query(OfferTag).filter(OfferTag.tag_id == award_id).first()
```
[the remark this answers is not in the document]
danielamassiceti: The offerTag id is the primary key, whereas we do actually want to filter by tag_id which is something different (i.e. the id of class Tag) :)

avishkar58 (approved):

Issue #1110, opened 2023-03-26
Add reviews to response page


shreyalpandit:
Extend the response page to also display all reviews for a response. This way, the action editor can look at the reviews with the application before submitting an outcome.

PR #1111, opened 2023-03-26
Render reviews

shreyalpandit:
Addresses Issue #1110

image


api/app/reviews/api.py
```diff
@@ -294,6 +315,33 @@ def get(self):
...
+
+        response = review_repository.get_response_by_reviewer(response_id, reviewer_user_id)
+        if response is None:
+            return RESPONSE_NOT_FOUND
+
+        review_response = review_repository.get_review_response(review_form.id, response_id, reviewer_user_id)
+        reviewer = user_repository.get_by_id(reviewer_user_id)
+        
+
+        return ReviewResponseUser(review_form, response, 0, args['language'], review_response=review_response, reviewer=reviewer)
```
avishkar58, line 343: This should be response.language instead of args['language']

webapp/src/pages/ResponsePage/ResponsePage.css
```diff
@@ -220,6 +220,12 @@ html .headings-lower label {
     font-size: 13px;
 }
 
+.reviewers-section h4 {
+    font-size: 17px;
+    /* color: gray; */
```
avishkar58, line 225: Please remove commented code

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -39,21 +40,30 @@ class ResponsePage extends Component {
...
                 applicationData: responses[2].detail,
                 tagList: responses[3].tags,
-                reviewers: responses[4].reviewers,
-                error: responses[0].error || responses[1].error || responses[2].error || responses[3].error || responses[4].error,
-                isLoading: false
-            }, this.handleData);
-        });
+                reviewers: responses[2].detail.reviewers,
+                error: responses[0].error || responses[1].error || responses[2].error || responses[3].error,
+            }, this.handleData, this.getReviewResponses(responses[2].detail, responses[2].detail.reviewers));
```
avishkar58, line 52: Hmm this is interesting, adding a second callback separated by a comma. According to the type specification of setState it only takes a single callback function though, so even if this does work, it would be safer to do it in a single callback function like 
```
()=>{this.handledata(); this.getReviewResponse(...) }
```
to protect ourselves from any changes that might break this

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -39,21 +40,30 @@ class ResponsePage extends Component {
...
+    });
     };
 
+    getReviewResponses(applicationData, reviewers) {
+        reviewers.map((val, index) => {
+            let num = index + 1;
+            reviewService.getResponseReviewAdmin(applicationData.id, this.props.event.id, val.reviewer_user_id)
```
avishkar58, line 59: Since you've made a specific backend API for this feature, it probably makes sense to return all the reviews in one call instead of having multiple calls for each review. This would be more efficient. That way you can also wrap it up in an object that maybe has a single review_form (with all the questions) and a list of all the review responses.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -106,6 +116,51 @@ class ResponsePage extends Component {
         };
     };
 
+    renderReviewResponse(review_response, section) {
+        const questions = section.review_questions.map(q => {
+            const a = review_response.scores.find(a => a.review_question_id === q.id);
+            if (q.type === "information") {
+                return <h4>{q.headline}</h4>
+            };
```
avishkar58, line 124: So "information" questions in the review form are used to refer back to answers in the original response, so that the reviewer can see what the candidate's answers were in the review form. Now on the response page, you of course already have the full response, so you don't need these information questions and can instead just filter them out. One simple way to do it is to just not display anything where you can't find `a`.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -106,6 +116,51 @@ class ResponsePage extends Component {
...
+            
+        });
+        return questions
+    };
+
+    // Render Reviews
+    renderCompleteReviews(){
+        if (this.state.reviewResponses) {
+                const reviews = this.state.reviewResponses.map((val, index) => {
+                    let num = index + 1;
```
avishkar58, line 146: num seems to be unused?

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -106,6 +116,51 @@ class ResponsePage extends Component {
...
+                    return <div className="section">
+                        {val.review_response.is_submitted === true &&
```
avishkar58, line 148: If you're not showing unsubmitted reviews on the front-end (which makes sense), then you also shouldn't return them from the backend so we don't send unnecessary data over the network. Will be easy to filter those out if you make your backend API just return all the (submitted) reviews in one go.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -39,21 +40,30 @@ class ResponsePage extends Component {
...
+            }, this.handleData, this.getReviewResponses(responses[2].detail, responses[2].detail.reviewers));
+    });
     };
 
+    getReviewResponses(applicationData, reviewers) {
+        reviewers.map((val, index) => {
+            let num = index + 1;
+            reviewService.getResponseReviewAdmin(applicationData.id, this.props.event.id, val.reviewer_user_id)
+            .then(resp => {
+                this.state.reviewResponses.push(resp.form);
```
avishkar58, line 61: You should never modify state directly like this because React always assumes state is immutable and modifying it will cause weird bugs and rendering glitches. To update a list in React state, use something like this: 
```
this.setState(prevState => {
   reviewResponses: [...prevState.reviewResponses, resp.form],
   isLoading: false
});
```

avishkar58 (changes requested):
Thanks Shreya! 
One nice feature would be to collapse the reviews in case the admin is only interested in seeing the original response. In fact, maybe we should show the reviews at the bottom of the response, think I have a mild preference for that, definitely not a strong opinion. WDYT?

api/app/reviews/api.py
```diff
@@ -79,7 +79,26 @@
...
-    'language': fields.String
+    'language': fields.String,
+    'is_submitted': fields.Boolean,
+    'submitted_timestamp': fields.DateTime(dt_format='iso8601')
+}
+
+extended_scores_fields = {
+    'review_question_id': fields.Integer, 
+    'headline': fields.String, 
+    'description': fields.String,
```
avishkar58, line 90: Are `headline` and `description` needed considering you can get them from the review form? (not too concerned about keeping it like this if it makes it easier for you though)

api/app/reviews/api.py
```diff
@@ -294,6 +332,28 @@ def get(self):
...
+    @event_admin_required
+    def get(self, event_id):
+        parser = reqparse.RequestParser()
+        parser.add_argument('response_id', type=int, required=True)
+        parser.add_argument('language', type=str, required=True)
+        args = parser.parse_args()
+
+        response_id = args['response_id']
+
+        review_form = review_repository.get_review_form(event_id)
```
avishkar58, line 345: So there can actually be multiple review forms for an event representing multi-stage review processes (see `stage` parameter on the review repository). So what you're doing is showing the reviews for the active review stage, which is fine, but maybe just put a note on the UI under your "Reviews" heading saying that this only shows the active stage (and in fact you can get the stage number from the review form and display that too). Don't worry about adding support for multiple stages now, instead please raise a new issue to add support for this and we can do that later if someone actually needs it.

api/app/reviews/api.py
```diff
@@ -294,6 +332,28 @@ def get(self):
...
+        if review_form is None:
+            return REVIEW_FORM_NOT_FOUND
+
+        response_reviews = (review_repository.get_all_review_responses_by_response(review_form.id, response_id))
+        serialized_reviews =  [
+            ReviewResponseDetailListAPI._serialise_review_response(response, args['language'])
+            for response in response_reviews
+        ]
+        return marshal(ReviewResponses(review_form, serialized_reviews, args['language']), response_review_fields), 201
```
avishkar58, line 354: The status code to return for a successful GET is 200 ("OK"), not 201 which stands for "CREATED" and would usually be used in a POST request.

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -50,15 +50,24 @@ class ResponsePage extends Component {
...
+                this.getReviewResponses(responses[2].detail);
             });
         });
     };
 
+    getReviewResponses(applicationData) {
+        reviewService.getResponseReviewAdmin(applicationData.id, this.props.event.id)
+            .then(resp => {
+                this.setState( {
+                    reviewResponses: resp.form.review_responses,
```
avishkar58, line 65: What happens if there is no review form defined for an event and the API raises a review_form_not_found error? I think it will work and silently ignore the error (which I guess is OK), but just wanted to double check if you've tested that the response page still works in that scenario?

avishkar58 (approved):
Thanks Shreya! This looks great. 
Some very minor changes still left, but approving in the meantime :)

webapp/src/pages/ResponsePage/ResponsePage.js
shreyalpandit: I didn't accounted for this, but good catch! Thanks!

PR #1115, opened 2023-03-31
Action Editor DB

shreyalpandit:
Creates new database to track which action editor is assigned to a response.


avishkar58:
> Hi Shreya, can you remind me what exactly an action-editor is meant to be please?

Oh nevermind, it makes sense now after seeing your next PR :) Some kind of admin that is assigned to individual responses.

shreyalpandit:
Yes! There are three main roles: Event Admin, Action Editor, and Reviewer. The Event Admins are tracking the submission of the submissions but aren't actually reviewing or approving. They will assign the action editor to a submission and this person is responsible for adding reviewers and setting the outcome of the submission. Action editors can be reviewers for submissions that they are not assigned to as an action editor role, but reviewers cannot be action editors.

avishkar58 (approved):
Awesome, thanks for the explanation!

PR #1116, opened 2023-04-03
added action editor view

shreyalpandit:
Added a view for users who are an action editor for a particular event. They are able to view the responses that they were assigned to and assign reviewers to a response.
(Still need to add tests for action editor permissions)

image

image

### DIFF
--- a/api/app/responses/api.py
+++ b/api/app/responses/api.py
@@ -22,7 +22,7 @@ from app.users.models import AppUser
 from app.users.repository import UserRepository as user_repository
 from app.utils import emailer, errors, strings, pdfconvertor, zipping, storage
 from app.utils.zipping import zip_in_memory
-from app.utils.auth import auth_required, event_admin_required
+from app.utils.auth import auth_required, event_admin_or_action_editor_required
 from flask import g, request, send_file
 from flask_restful import fields, inputs, marshal_with, reqparse
 from sqlalchemy.exc import SQLAlchemyError
@@ -276,7 +276,7 @@ def _serialize_tag(tag, language):
 
 class ResponseListAPI(restful.Resource):
 
-    @event_admin_required
+    @event_admin_or_action_editor_required
     def get(self, event_id):
         req_parser = reqparse.RequestParser()
         req_parser.add_argument('include_unsubmitted', type=inputs.boolean, required=True)
@@ -288,10 +288,15 @@ class ResponseListAPI(restful.Resource):
         include_unsubmitted = args['include_unsubmitted']
         question_ids = args['question_ids[]']
         language = args['language']
-        
+        current_user = user_repository.get_by_id(g.current_user['id'])
+
         print(('Include unsubmitted:', include_unsubmitted))
 
-        responses = response_repository.get_all_for_event(event_id, not include_unsubmitted)
+        if current_user.is_event_admin(event_id):
+            responses = response_repository.get_all_for_event(event_id, not include_unsubmitted)
+        
+        elif current_user.is_action_editor(event_id):
+            responses = response_repository.get_all_for_action_editor(event_id, g.current_user['id'], not include_unsubmitted)
 
         review_config = review_configuration_repository.get_configuration_for_event(event_id)
         required_reviewers = 1 if review_config is None else review_config.num_reviews_required + review_config.num_optional_reviews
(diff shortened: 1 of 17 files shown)


api/app/responses/api.py
```diff
@@ -288,10 +288,15 @@ def get(self, event_id):
...
-        
+        current_user = user_repository.get_by_id(g.current_user['id'])
+
         print(('Include unsubmitted:', include_unsubmitted))
 
-        responses = response_repository.get_all_for_event(event_id, not include_unsubmitted)
+        if current_user.is_event_admin(event_id):
+            responses = response_repository.get_all_for_event(event_id, not include_unsubmitted)
+        
+        elif current_user.is_action_editor(event_id):
```
avishkar58, line 298: I wonder if there's a better name than "action-editor" that's not a journal-specific term, since the functionality is not limited to a journal event type. Will try think of something too!

api/app/responses/api.py
```diff
@@ -306,10 +311,18 @@ def get(self, event_id):
...
         for response in responses:
-            reviewers = [_serialize_reviewer(r, reviewer_to_review_response.get(r.reviewer_user_id, None)) 
-                         for r in response_to_reviewers.get(response.id, [])]
+            reviewers = []
+            action_editor = None
+            for r in response_to_reviewers.get(response.id, []):
+                if not r.is_action_editor:
+                    reviewers.append(_serialize_reviewer(r, reviewer_to_review_response.get(r.reviewer_user_id, None)))
+                else:
+                    action_editor = _serialize_reviewer(r, reviewer_to_review_response.get(r.reviewer_user_id, None))
```
avishkar58, line 324: Nit: Since the line of code is fairly complex, would extract another variable here to make it cleaner (also a very minor comment that's it's generally easier to read "positive" conditions rather than negative ones) 

```
reviewer = _serialize_reviewer(r, reviewer_to_review_response.get(r.reviewer_user_id, None))
if r.is_action_editor:
  action_editor = reviewer
else:
  reviewers.append(reviewer)
```

api/app/responses/api.py
```diff
@@ -332,8 +345,10 @@ def get(self, event_id):
                 'tags': [_serialize_tag(rt.tag, language) for rt in response.response_tags]
             }
 
+            if (event.event_type == EventType.CONTINUOUS_JOURNAL or event.event_type == EventType.JOURNAL):
```
avishkar58, line 348: Think this condition isn't needed. We might want to use action-editors for other event types as well :)

api/app/reviews/api.py
```diff
@@ -431,11 +442,10 @@ class ReviewAssignmentAPI(GetReviewAssignmentMixin, PostReviewAssignmentMixin, r
...
-    @auth_required
+    @event_admin_or_action_editor_required
     @marshal_with(reviews_count_fields)
-    def get(self):
+    def get(self, event_id):
         args = self.get_req_parser.parse_args()
-        event_id = args['event_id']
         user_id = g.current_user['id']
 
         current_user = user_repository.get_by_id(user_id)
```
avishkar58, line 451: Now that you've added the decorator, you can remove the current_user.is_event_admin check

api/app/utils/auth.py
```diff
@@ -104,3 +104,20 @@ def wrapper(*args, **kwargs):
         return FORBIDDEN
 
     return wrapper
+
+def event_admin_or_action_editor_required(func):
```
avishkar58, line 108: Very cool! :)

webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -482,41 +481,55 @@ class ResponsePage extends Component {
...
     // Reviews
     // Render Reviews
     renderReviews() {
 
-        if (this.state.applicationData) {
-            if (this.state.applicationData.reviewers) {
-                const reviews = this.state.applicationData.reviewers.map((val, index) => {
-                    let num = index + 1;
-                    //   {"reviewer_user_id": 4, "user_title": "Mr", "firstname": "Joe", "lastname": "Soap", "status": "completed"},
+        if (this.state.applicationData.reviewers) {
```
avishkar58, line 501: This should be `if (this.state.applicationData && this.state.applicationData.reviewers)` in case applicationData is null (which is what the code was previously trying to do in a very ugly way). In fact, while you're editing this code, it would be better to structure it in the form of a pre-condition to reduce the nesting, so you can do:

``` 
if (! this.state.applicationData || ! this.state.applicationData.reviewers) {
  return <div/>
}

[...]  //Rest of your code now less nested
```

webapp/src/pages/reviewDashboard/components/ReviewDashboard.js
```diff
@@ -25,7 +25,9 @@ class ReviewListComponent extends Component {
     }
 
     componentDidMount() {
+        console.log('got to load', this.props)
```
avishkar58, line 28: Please remove debug logs

avishkar58 (commented):
This is really well done, thank you Shreya!!

avishkar58:
Will approve once you've added the tests for action editor permissions :)

PR #1148, opened 2023-07-08
Fix display of uploaded files when editing registration

amritpurshotam:
Fixed display of uploaded files at registration and guest registration. The uploaded file is clickable and should download.

Before
before

After
after

cc @smhall97


amritpurshotam:
Hey @avishkar58, just checking in if you've had a chance to look at this yet? A lot of users are sending in support emails to ask if their passport has been uploaded. With this fix, this can be avoided since it will now be clear that their upload did indeed work.

avishkar58:
Hey Amrit! Thank you for making this fix, and for reminding me about it - sorry for the delay!

avishkar58 (approved):

PR #1136, opened 2023-05-31
Updated response table and added one migration

lucas-meyer-7:
I added a field, called ''parent_id'', to the response table in the database. This field will be used to chain multiple revisions of a journal/article submission.

PR #1150, opened 2023-07-26
DB changes: Added two fields to reviews models for continuous journals

lucas-meyer-7:
@amritpurshotam  Hi Amrit, in this PR I added a new field to the review form table called `reviewer_type`. It is a string field which is used when the review form related to an `Event` of the type `CONTINUOUS_JOURNAL`. The value of `reviewer_type` should be `None` if the event type is continuous journal. If the event is a continuous journal, then the value of `reviewer_type` should be either "Meta-reviewer" or "Reviewer".

Let me know what you think and if I should change anything. I'll start with some other PRs in the meantime.

### DIFF
--- a/api/app/reviews/api.py
+++ b/api/app/reviews/api.py
@@ -958,6 +958,7 @@ class ReviewFormDetailAPI(restful.Resource):
         req_parser.add_argument('deadline', type=dt_format, required=True)
         req_parser.add_argument('active', type=bool, required=True)
         req_parser.add_argument('sections', type=dict, required=True, action='append')
+        # req_parser.add_argument('reviewer_type', type=dict, required=True, action='append')
         args = req_parser.parse_args()
 
         application_form_id = args['application_form_id']
@@ -966,6 +967,7 @@ class ReviewFormDetailAPI(restful.Resource):
         deadline = args['deadline']
         active = args['active']
         sections_data = args['sections']
+        # reviewer_type = args['reviewer_type']
 
         review_form = review_repository.get_review_form(event_id, stage)
         if review_form is not None:
@@ -978,8 +980,11 @@ class ReviewFormDetailAPI(restful.Resource):
             deadline, 
             stage, 
             # Only make this stage active if there is no other active stage.
-            active=not any(o.active for o in other_review_forms))
-        
+            active=not any(o.active for o in other_review_forms),
+            reviewer_type="Reviewer")
+            # reviewer_type="Meta-reviewer")
+            # reviewer_type=reviewer_type)
+
         review_repository.add_model(review_form)
 
         for section_data in sections_data:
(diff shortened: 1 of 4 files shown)


lucas-meyer-7:
@amritpurshotam I added another field to the database in the meantime. The new field is called `private` and it specifies whether a question from a journal reviewer is private or not. I think that should cover all of the required DB changes for supporting continuous journal events.

amritpurshotam:
Hey @lucas-meyer-7. I've gone over the Journal Submission Process doc again and it's not clear to me what the overall design of this feature is heading towards. My suggestion is to split up by the PRs according to completed features instead of horizontal pieces of several features (for e.g. database changes in isolation). For backend code like this, this would mean doing the database change, updating the API and tests, and any other logic that make the feature work. This will help with reviewing the code and the design of the feature as a whole as well as keeping the code isolated until it's working and integrated with the rest of the codebase. I know this will be more challenging but I'm happy to provide reviews in the same PR for each horizontal piece until the feature is completed at which point then we can merge it in.

From what I can gather your intention here is you're adding pieces to two independent pieces of functionality

1. Adding the concept of a meta-reviewer and reviewer
2. Addressing the following story: _As a reviewer, I can submit private review responses that are only viewable by the journal administrators (action editor). I can also submit public review responses that the applicant will also be able to look at to get feedback._

For 1, it's not clear to me why we need a `reviewer_type` column on the `ReviewForm`. Do we expect to have separate review forms for meta-reviewers and reviewers? From the stories, I gather that meta-reviewers have a few more privileges and in particular can add their own review (_As a journal meta-reviewer (editor), I can desk-reject a submission by adding response text and moving it to a “rejected” state._). If this is the case then it sounds like this problem would be solved by adding a `meta-reviewer` role to the user (see the `AppUser` class for an idea of this where we specify what roles and therefore the level of access a user has to the system). Then we can add special cases for users with this role to provide their own review and immediately move a submission to a reject status. What do you think of this approach and does it fit in to what you have in mind?

For 2, I see you've added the `private` column to the `ReviewQuestion`. Do we expect to have separate public and private review questions (and hence public/private responses to them) or just that some questions will have private answers? From the story we have, it seems to indicate that it's the latter in which case it may make more sense to put it on `ReviewScore` perhaps.

So I think for now, let's address these questions and also if you can split up the PR for each feature that will be helpful. Happy for you to work on multiple concurrent PRs concurrently if you find that useful (just be careful about managing your local db on each branch, follow the steps I mentioned [previously](https://github.com/deep-learning-indaba/Baobab/pull/1136#discussion_r1257270166) to reset your db when switching branches)

cc @avishkar58 in case you have more insight to these points

lucas-meyer-7:
@amritpurshotam Thank you a lot for the review. I appreciate you going over the doc and pointing out your concerns. I am going to send an e-mail to Ulrich to understand what he wants, and I will seperate the PRs once I get a reply from him. :+1:

PR #1151, opened 2023-07-29
Remove outdated test

amritpurshotam:
Removes the test that checks if a user pays for an invoice that's not their own. Invoices were made payable by anyone here #1134.


amritpurshotam:
Hey @avishkar58 Just got the build green again by fixing the last (but unapplied) migration and an outdated test.

PR #1152, opened 2023-07-30
Added is_private field to Review Response table

lucas-meyer-7:
Hi @amritpurshotam , I think that you are right, and I made the DB changes that you recommended for the feature of specifying whether a response from a reviewer is private/public.

### DIFF
--- a/api/app/reviews/api.py
+++ b/api/app/reviews/api.py
@@ -388,12 +388,14 @@ class ReviewResponseAPI(GetReviewResponseMixin, PostReviewResponseMixin, restful
         scores = args['scores']
         language = args['language']
         is_submitted = args['is_submitted']
+        # is_private = args['is_private']
 
         response_reviewer = review_repository.get_response_reviewer(response_id, reviewer_user_id)
         if response_reviewer is None:
             return FORBIDDEN
 
-        review_response = ReviewResponse(review_form_id, reviewer_user_id, response_id, language)
+        # review_response = ReviewResponse(review_form_id, reviewer_user_id, response_id, language, is_private)
+        review_response = ReviewResponse(review_form_id, reviewer_user_id, response_id, language, True)
         review_response.review_scores = self.get_review_scores(scores)
         if is_submitted:
             review_response.submit()
--- a/api/app/reviews/models.py
+++ b/api/app/reviews/models.py
@@ -195,6 +195,7 @@ class ReviewResponse(db.Model):
     language = db.Column(db.String(2), nullable=False)
     is_submitted = db.Column(db.Boolean(), nullable=False)
     submitted_timestamp = db.Column(db.DateTime(), nullable=True)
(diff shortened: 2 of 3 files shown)


amritpurshotam (changes requested):

lucas-meyer-7:
@amritpurshotam  Thank you for your review. I appreciate the effort. About the "P.S. can we remove the commented code?", I am not sure exactly how the requests are handled (I'm very new to using ``reqparse``) and I am still figuring it out. Will the ``is_private`` field be automatically generated for the reviewer to specify in the UI? I assume that we would still need to add some JS for all of that to work.

I have another question about applying for events [1] (see screenshots below for context). When I run the project locally, and apply to an event that I created locally, it throws an error [2] (Line 1094). This is because ``formResponse`` is null and an error is thrown "" [3]. I'm not sure why this happens, but my best guess is that the error is thrown because I am running the project locally.

1:
image

2:
image

3:
image

amritpurshotam:
@lucas-meyer-7 

>  I am not sure exactly how the requests are handled (I'm very new to using reqparse) and I am still figuring it out.

Here's the [docs](https://flask-restful.readthedocs.io/en/latest/reqparse.html). Have a look at the `PostReviewResponseMixin` specifically for the class you will need to modify for this request specifically. The code is quite self-explanatory. Pretty much just add a new argument for `is_private`.

>  Will the is_private field be automatically generated for the reviewer to specify in the UI? I assume that we would still need to add some JS for all of that to work.

This is correct, there will still need to be updates made to the front end.

And then the last part with your errors. I think that sounds like an error with the data. Did you insert the data manually? Have a look at these ERDs to see a higher level view of the tables and hopefully a better of the view of the app without having to click through every screen. There's a lot of configuration and access granted to the system based on the type of user you are and it's easy to get wrong (when doing so manually) and quite time consuming as well to do so through the front end. My advice for now though is not to get too caught up in the front-end and try to gain understanding from reading the API code, tests, and these tables.

OrganisationEventsUsers
Forms
EventLeadup

lucas-meyer-7:
@amritpurshotam  Thank you for your answers and advice. I'll quickly make the changes required for this feature, excluding the front-end code required to make the ``is_private`` field appear to the applicant.

amritpurshotam (commented):
@lucas-meyer-7 Thanks this looks good so far. Some minor points though, can you change the branch you're merging into back to `develop`. Then merge in the `fix/payment-test` into this branch i.e.
```
git fetch
git checkout db_review_response_private_field
git merge fix/payment-test
```
I know this will clutter things a bit temporarily but once the `fix/payment-test` branch is merged into `develop` those changes will automatically disappear from your PR. This is just to get the build passing on your branch again and we can rely on all the automated checks it provides.

Otherwise going to leave this PR as is. We can only approve once the feature is complete. Feel free to branch off further from here though if you need the separation.

lucas-meyer-7:
@amritpurshotam Ready for merge?

PR #1182, opened 2024-02-28
Creation of Journal event.

umair-nasir14:
Added journal event type in eventConfig page and allowed journal events to be created. Replaced continuous journal to journal for consistency across the code base.


avishkar58 (changes requested):
Hey @umair-nasir14 , thanks for the PR and apologies for taking long to review it! 
This is a really great start and it's fantastic to simplify things and remove the distinction between "JOURNAL" and "CONTINUOUS_JOURNAL". I've left a few pointers to things you wouldn't have known about, but which need to be updated before we can submit this please. Thanks!

umair-nasir14:
Hi @avishkar58! Thank you for your kind words and great review. I have catered to the changes you advised. I apologize for changing the older migrations. Now I know to check after changing through general search 😄

api/migrations/versions/2d7f5a0819e3_.py
```diff
@@ -0,0 +1,52 @@
+"""empty message
```
avishkar58, line 1: Please add a descriptive message

api/migrations/versions/2d7f5a0819e3_.py
```diff
@@ -0,0 +1,52 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index('answer_question_lookup', table_name='answer')
+    op.drop_index('answer_response_lookup', table_name='answer')
+    op.alter_column('event', 'event_type',
```
avishkar58, line 21: This existing_type doesn't include the CONTINUOUS_JOURNAL which is in the current database and there is a bunch of generated lines in here that shouldnt' be like all these drop_index entries - so i suspect what's happened is that this is a result of generating the migration when your database was in an invalid state. This was most likely caused by the editing of old migrations. It's possible to do a series of migration downgrades to get your database back to an old state pre-edits, delete this migration, upgrade again, and then generate this migration again. But what I suggest doing instead is delete this migration. delete your db docker container and build a new one, then apply all the migrations with an upgrade, then generate this migration again - that way you can be sure it was generated by comparing to the proper latest database state.

avishkar58 (commented):
Thanks for the changes! Unfortunately it looks like the new migration you've added was possibly generated from an invalid database state, because there are a bunch of entries in there that are not expected.

api/migrations/versions/2d7f5a0819e3_.py
umair-nasir14: I have followed your instructions and after migrating and upgrading, no errors persist. Please let me know if there is something to be addressed.

avishkar58 (approved):
Thanks for the changes! It looks great - will merge it in now :)

umair-nasir14:
Thank you very much @avishkar58! :)

PR #1206, opened 2024-04-01
Fix failing tests and other project health updates

amritpurshotam:
The reason for the failing tests with the integrity error during drop_all (for e.g. see [here](https://app.circleci.com/pipelines/github/deep-learning-indaba/Baobab/3082/workflows/f6cfe4b4-5a25-4b3f-bb39-56f87f4620bf/jobs/9958)) is because `Section` depended on `Question` via the `depends_on_question_id` column. During `tearDown()` when `drop_all()` is called, it would attempt to drop the `question` table but then would fail since `section` is still referencing rows in it which enforced it via the foreign key constraint.

To fix this we specify that when a question is deleted, if there's a `depends_on_question_id` set to it on `Section`, it would then be set to `NULL`. This does introduce a risk that if there's a legitimate reason for a question to be deleted, an error will not be raised if a section is still depending on it. 

The complexity though is that there's essentially a circular reference between `Section` and `Question`. A section can have many questions. And a question can have many sections (via the depends_on). So especially during the tests, we have to choose one table to be dropped first somehow without raising the integrity error. 

Other fixes
- Added the indexes to the python models to align them to the db where they were added in this migration #1161 
- Minor change but returned migrations back to original state after they were changed in #1182 most likely by a find replace all
- Completed the removal of `CONTINUOUS_JOURNAL` by also adding a migration to remove it from the db enums. Important since if there's any other changes made to this column in the future, this extra enum will not pop up again in the migration. 
- Explicitly name the `event_type` and `tag_type` enums to align with the db so migrations don't unintentionally get created to try to change it to `eventtype` and `tagtype` (unclear why it's trying to do this only now). Also had to align a previous migration `downgrade()` to use the names with the underscore so it actually matches the `upgrade()`
- Updated currency conversion to use the latest API endpoint. See [migration guide](https://github.com/fawazahmed0/exchange-api/blob/main/MIGRATION.md) here.
- Remove unused field introduced in #1204 causing null errors in the tests.
- Also from #1204, updated the `reviewers` to the renamed `availableReviewers` in the front end test (with the additional fields returned from the review assignment API)


api/app/applicationModel/models.py
```diff
@@ -91,7 +91,7 @@ class Section(db.Model):
     id = db.Column(db.Integer(), primary_key=True)
     application_form_id = db.Column(db.Integer(), db.ForeignKey('application_form.id'), nullable=False)
     order = db.Column(db.Integer(), nullable=False)
-    depends_on_question_id = db.Column(db.Integer(), db.ForeignKey('question.id', use_alter=True), nullable=True)
+    depends_on_question_id = db.Column(db.Integer(), db.ForeignKey('question.id', name="section_depends_on_question_id_fkey", use_alter=True, ondelete="SET NULL"), nullable=True)
```
amritpurshotam, line 94: This is what fixes 99% of the broken tests.

api/app/events/models.py
```diff
@@ -47,7 +47,7 @@ class Event(db.Model):
     offer_close = db.Column(db.DateTime(), nullable=True)
     registration_open = db.Column(db.DateTime(), nullable=True)
     registration_close = db.Column(db.DateTime(), nullable=True)
-    event_type = db.Column(db.Enum(EventType), nullable=False)
+    event_type = db.Column(db.Enum(EventType, name="event_type"), nullable=False)
```
amritpurshotam, line 50: alembic is trying to rename the deb enum to `eventtype` in the migrations for some reason. This prevents that.

api/app/invoice/generator.py
```diff
@@ -170,7 +167,7 @@ def __init__(self, name, value):
 
 def _get_exchange_rate():
     exchange_data = requests.get(_USDGBP_PATH)
-    return exchange_data.json()["gbp"]
+    return exchange_data.json()["usd"]["gbp"]
```
amritpurshotam, line 170: Endpoint was updated to this now.

api/app/registration/models.py
```diff
@@ -181,6 +181,9 @@ def confirm(self):
 
 class RegistrationAnswer(db.Model):
     __tablename__ = "registration_answer"
+    __table_args__ = tuple([
+        db.Index("registration_answer_lookup", "registration_id", "registration_question_id"),
+    ])
```
amritpurshotam, line 186: #1161

api/app/responses/models.py
```diff
@@ -65,6 +65,10 @@ def withdraw(self):
 
 class Answer(db.Model):
     __tablename__ = "answer"
+    __table_args__ = tuple([
+        db.Index("answer_question_lookup", "question_id", "response_id", "is_active"),
+        db.Index("answer_response_lookup", "response_id", "is_active")
+    ])
```
amritpurshotam, line 71: #1161

api/app/tags/models.py
```diff
@@ -12,7 +12,7 @@ class Tag(db.Model):
     __tablename__ = 'tag'
     id = db.Column(db.Integer(), primary_key=True)
     event_id = db.Column(db.Integer(), db.ForeignKey('event.id'), nullable=False)
-    tag_type = db.Column(db.Enum(TagType), nullable=True)
+    tag_type = db.Column(db.Enum(TagType, name="tag_type"), nullable=True)
```
amritpurshotam, line 15: Same thing with the alembic migrations trying to rename this

api/migrations/versions/0edb89e87e72_indabaxza_offer_email_templates.py
```diff
@@ -80,7 +80,7 @@ def __init__(
         miniconf_url=None
     ):
         self.start_date = start_date
-        self.end_date = None if event_type == EventType.JOURNAL else end_date
+        self.end_date = None if event_type == EventType.CONTINUOUS_JOURNAL else end_date
```
amritpurshotam, line 83: This was changed accidentally so putting back to what it should be

api/migrations/versions/2ae07ae036c1_.py
```diff
@@ -32,12 +32,12 @@ def upgrade():
...
     # ### commands auto generated by Alembic - please adjust! ###
     op.alter_column('tag', 'tag_type',
-               existing_type=sa.Enum('RESPONSE', 'REGISTRATION', 'GRANT', 'QUESTION', name='tagtype'),
+               existing_type=sa.Enum('RESPONSE', 'REGISTRATION', 'GRANT', 'QUESTION', name='tag_type'),
                type_=postgresql.ENUM('RESPONSE', 'REGISTRATION', 'GRANT', 'QUESTION', name='tag_type'),
                existing_nullable=True)
     
     op.alter_column('event', 'event_type',
-               existing_type=sa.Enum('EVENT', 'AWARD', 'CALL', 'PROGRAMME', 'JOURNAL', name='eventtype'),
+               existing_type=sa.Enum('EVENT', 'AWARD', 'CALL', 'PROGRAMME', 'JOURNAL', name='event_type'),
```
amritpurshotam, line 40: This just aligns the downgrade with the upgrade

api/migrations/versions/349da0b8780a_journal_event_type.py
```diff
@@ -1,4 +1,4 @@
-"""Add journal and JOURNAL event_type
+"""Add journal and continuous_journal event_type
```
amritpurshotam, line 1: also accidental update that I'm returning back to original state

api/migrations/versions/e511f343cb08_complete_removal_of_continuous_journal.py
```diff
@@ -0,0 +1,27 @@
...
+    op.execute("""DELETE FROM pg_enum
+        WHERE enumlabel = 'CONTINUOUS_JOURNAL' 
+        AND enumtypid = (
+        SELECT oid FROM pg_type WHERE typname = 'event_type'
+        )""")
+
+
+def downgrade():
+    op.execute("COMMIT")
+    op.execute("ALTER TYPE event_type ADD VALUE 'CONTINUOUS_JOURNAL'")
```
amritpurshotam, line 27: This enum wasn't removed from the db. Need to do this so it doesn't keep popping up when creating new migrations.

webapp/src/pages/ResponsePage/ResponsePage.test.js
```diff
@@ -107,7 +111,7 @@ test("Check if tag list renders.", async () => {
     wrapper.setState({
         isLoading: false,
         applicationData: applicationData,
-        reviewers: reviewersData,
+        availableReviewers: reviewersData,
```
amritpurshotam, line 114: this fixes the front end tests

amritpurshotam:
@avishkar58 Build is green again 🥳 There's a lot going on here but I've tried to make clear in line what each change is doing. Let me know if you have any questions. I only had one question around the `filtered_answers` bit if you could double check that

api/app/applicationModel/models.py
avishkar58: Wow, you are a wizard! 🧙

api/app/reviews/api.py
```diff
@@ -270,15 +270,14 @@ def _filter_response(response: Response) -> ResponseModel:
...
-                 review_response=None, reviewer=None, filtered_answers=None):
+                 review_response=None, reviewer=None):
         self.review_form = _serialize_review_form(review_form, language)
         self.response = response
         self.user = None if response is None else response.user
         self.reviews_remaining_count = reviews_remaining_count
         self.references = reference_responses
         self.review_response = review_response
         self.reviewer = reviewer
-        self.filtered_answers = filtered_answers or response.answers
```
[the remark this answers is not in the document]
avishkar58: Ah looks this is a change that accidently got released from something i was working on and haven't completed yet! Thanks for catching it! :)

avishkar58 (approved):
AMAZINGGGG - thank you so much Amrit!!! 🎉 🎉 🎉

amritpurshotam:
@avishkar58 Haha you welcome 😄 I also saw now that `migrate-test` didn't pass. It looks like one my migrations dropping CONTINUOUS_JOURNAL didn't work. I'll look into it today.

PR #1208, opened 2024-04-05
Drop enum value safely

amritpurshotam:
This attempts to fix the [permission error](https://app.circleci.com/pipelines/github/deep-learning-indaba/Baobab/3098/workflows/ef3d68b1-fcea-4ef8-89f6-fc6df9ee4c0b/jobs/10000?invite=true#step-102-21183_111) when running the migrations in `migrate-test`. The previous way was apparently quite unsafe so the database guarding against this type of query was good! Followed this [guide](https://blog.yo1.dog/updating-enum-values-in-postgresql-the-safe-and-easy-way/) now instead.

We also haven't released so should be fine to modify this migration.

### DIFF
--- a/api/migrations/versions/e511f343cb08_complete_removal_of_continuous_journal.py
+++ b/api/migrations/versions/e511f343cb08_complete_removal_of_continuous_journal.py
@@ -15,11 +15,12 @@ import sqlalchemy as sa
 
 
 def upgrade():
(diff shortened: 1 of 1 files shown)


avishkar58 (approved):
Definitely looks safer haha, thanks Amrit!!

PR #1285, opened 2024-09-21
DEST_REJECT added

umair-nasir14:
A DESK_REJECT status added, primarily for Journal events. To facilitate emails for this status, two new email templates added for journal desk reject and reject. A reason is also propagated to the email for desk_reject and reject, which is propogated through outcome services, therefore a new reason column is added in the outcome table. Finally, all migrations are performed.

Please let me know if there are any changes needed!


api/app/outcome/api.py
```diff
@@ -101,20 +110,33 @@ def post(self, event_id):
...
-            if (status == Status.REJECTED or status == Status.WAITLIST):  # Email will be sent with offer for accepted candidates  
-                email_user(
-                    'outcome-rejected' if status == Status.REJECTED else 'outcome-waitlist',
-                    template_parameters=dict(
-                        host=misc.get_baobab_host()
-                    ),
-                    event=event,
-                    user=user,
-                )
+            print("event_type::::::::", event.event_type)
```
upaq, line 119: "print" shouldn't be in the production codebase?

api/app/outcome/api.py
```diff
@@ -101,20 +110,33 @@ def post(self, event_id):
...
+            print("Type of event.event_type:::::::", type(event.event_type))
```
upaq, line 120: "print" shouldn't be in the production codebase?

 (please see comments above)

api/app/outcome/api.py
upaq, line 119: "print" shouldn't be in the production codebase?

api/app/outcome/api.py
umair-nasir14, line 1: Cleaned up. I think there is no other print statement that would go in prod.

PR #1297, opened 2024-11-25
Display all review sections in responsePage

Vilmo18:
Fix problem of method `renderReviewResponse()` and improve `renderCompleteReviews()`
Add a new method `getTotalScore()` for getting total review

Correct of minor bugs in:
- `reviewForm` page for computing totalscore during reviewing
- `responseList` page for displaying email instead of language

### DIFF
--- a/webapp/src/pages/ResponsePage/ResponsePage.js
+++ b/webapp/src/pages/ResponsePage/ResponsePage.js
@@ -1,937 +1,779 @@
-import React, { Component } from "react";
+import React, { Component } from 'react'
 import "react-table/react-table.css";
-import "./ResponsePage.css";
-import { withTranslation } from "react-i18next";
-import ReviewModal from "./components/ReviewModal";
-import { eventService } from "../../services/events/events.service";
-import { applicationFormService } from "../../services/applicationForm/applicationForm.service";
-import { reviewService } from "../../services/reviews/review.service";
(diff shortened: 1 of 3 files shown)


webapp/src/pages/ResponsePage/ResponsePage.js
```diff
@@ -114,42 +143,57 @@ class ResponsePage extends Component {
...
+                        </h4>
+                        {this.state.reviewForm.review_sections.map(section => (
+                            <div key={section.id} className="section">
+                            <div className="flex baseline"><h3>{section.headline}</h3></div>
+                                {this.renderReviewResponse(val, section)}
+                            </div>
+                        ))}
+                        <hr/>
+                        <div className="review-total-score">
+                        <h5>Total Score: {this.state.totalScore}</h5>
```
avishkar58, line 184: I think this should be the total score of the review that is being rendered, no?

avishkar58 (changes requested):
Thank you very much for the contribution Yvan! :) Have left some pointers where i think the naming of some of our services is confusing and doesn't quite correspond to what it actually does.

webapp/src/pages/ResponsePage/ResponsePage.js
Vilmo18: Indeed

Vilmo18:
I have fixed it.

PR #1304, opened 2024-12-09
Journal Event fixing bug front/back end

Vilmo18:
What's new?

-  **Possibility to submit a paper multiple times**: The condition allowing multiple applications has been enabled (by commenting out the original condition). If there is a more efficient or cleaner approach, I am open to suggestions.  
-  Addition of new statuses via migrations: Previously, new statuses were added without corresponding database changes. A migration script has now been created to ensure these changes are properly reflected in the database.  
- **Application and status visualization**: Users can now visualize their applications and track their current status.  
- **Outcome and response bug fix**: Previously, there was one outcome shared across all responses given an event and user. This has been updated so that each response has its own unique outcome. The database structure was modified, migrations were applied, and necessary changes were made to backend files.  
- **Minor bug fixes**: Several smaller bugs have been identified and resolved.


api/app/outcome/models.py
```diff
@@ -14,6 +14,7 @@ class Outcome(db.Model):
     id = db.Column(db.Integer(), primary_key = True, nullable = False)
     event_id = db.Column(db.Integer(), db.ForeignKey('event.id'), nullable = False)
     user_id = db.Column(db.Integer(), db.ForeignKey('app_user.id'), nullable = False)
+    response_id = db.Column(db.Integer(), db.ForeignKey('response.id'), nullable=False) ## add response_id
```
avishkar58, line 17: Please remove the comments "## add response_id" and "## add relationship"

api/app/outcome/repository.py
```diff
@@ -28,5 +28,19 @@ def get_latest_for_event(event_id):
     @staticmethod
     def add(outcome):
         db.session.add(outcome)
+
```
avishkar58, line 31: It's probably worth removing the get_*_for_event methods now that outcome is linked to a response? That way we can also ensure all the relevant code paths are updated.

api/app/responses/api.py
```diff
@@ -123,8 +123,9 @@ def get(self):
         responses = response_repository.get_all_for_user_application(current_user_id, form.id)
 
         # TODO: Link outcomes to responses rather than events to cater for multiple submissions.
-        outcome = outcome_repository.get_latest_by_user_for_event(current_user_id, event_id)
         for response in responses:
+            outcome = outcome_repository.get_latest_by_user_for_event_response(current_user_id,response.id, event_id)
+            # get_latest_by_user_for_event(current_user_id, event_id)
```
avishkar58, line 128: Please delete commented-out code

api/app/responses/api.py
```diff
@@ -146,8 +147,8 @@ def post(self):
         user = user_repository.get_by_id(user_id)
         responses = response_repository.get_all_for_user_application(user_id, application_form_id)
 
-        if not application_form.nominations and len(responses) > 0:
-            return errors.RESPONSE_ALREADY_SUBMITTED
+        # if not application_form.nominations and len(responses) > 0:
```
avishkar58, line 150: I think we still need this check - some events and application forms should still only support a single response and some should support multiple responses and we should validate this is the case. How about changing the name `nominations` to `allows_many_responses` to make it more generic? Then default `allows_many_responses` to True for event-type journal

api/migrations/versions/062abca36566_.py
```diff
@@ -0,0 +1,22 @@
+"""empty message
```
avishkar58, line 1: Please remove empty migration.

api/migrations/versions/41517b1f4730_.py
```diff
@@ -0,0 +1,22 @@
+"""empty message
```
avishkar58, line 1: Please remove empty migration

api/migrations/versions/4923563b3a63_add_reponseID_outcome.py
```diff
@@ -0,0 +1,42 @@
...
+revision = '4923563b3a63'
+down_revision = 'bf4cb562e36c'
+
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column(
```
avishkar58, line 19: This will break when we try to run it on the production database because there are already many records in the table, so as soon as it tries to add the column which doesn't allow nulls, it will fail. 
So what you'll need to do is:
- Add the column response_id with nullable=True
- Update the existing data to set the correct response_id
- Update the column to set nullable=False
Here's an example of a migration that does something similar: https://github.com/deep-learning-indaba/Baobab/blob/8fb7327c8976bcd54d278804a321c7609811cca5/api/migrations/versions/e8d1ee2746e0_event_travel_grant.py#L19

api/migrations/versions/67d04fb5a5ed_.py
```diff
@@ -0,0 +1,22 @@
+"""empty message
```
avishkar58, line 1: Please remove empty migration

api/migrations/versions/ba73c854003e_.py
```diff
@@ -0,0 +1,22 @@
+"""empty message
```
avishkar58, line 1: Please remove empty migration

webapp/src/components/EventStatus.js
```diff
@@ -65,14 +65,20 @@ class EventStatus extends Component {
   applicationStatus = (event) => {
       const applyLink = `${event.key}/apply`
       const submissionLink = `${event.key}/apply/new`
+      const viewLink=`${event.key}/apply/view`
+      console.log(event);
```
avishkar58, line 69: Please remove debug console logs

webapp/src/components/EventStatus.js
```diff
@@ -91,6 +97,19 @@ class EventStatus extends Component {
           
       }
       else if (event.status.application_status === "Withdrawn") {
+        if (event.event_type === "JOURNAL"){
```
avishkar58, line 100: It seems like we need some bigger changes here to properly cater for multiple submissions with the event statusus. Currently this doesn't make sense that the event application status is "withdrawn" but the status here says "You have submitted your article"

webapp/src/components/form/FormCreator.js
```diff
@@ -147,7 +147,8 @@ const FormCreator = ({
         />
       {homeRedirect && <Redirect to={`/${eventKey}`} />}
       {errorResponse ? (
-        <div className='tooltiptext-error response-error'>
+        // tooltiptext-error response-error
```
avishkar58, line 150: Thanks for the fix! Please remove commented code

avishkar58 (commented):
Hi Yvan! Thanks so much for this great and very comprehensive PR! Really appreciate all the effort you're putting into this :)
I haven't finished reviewing it yet, but thought I'd submit the few comments I had so far just so I don't block your progress too much. Will try to finish the rest ASAP!

Vilmo18:
Hi @avishkar58 ,
Thanks for your comments.
I have already updated the code.
I will submit another PR with more updates soon.

webapp/src/pages/ReponseDetails/ResponseDetails.css
```diff
@@ -0,0 +1,408 @@
+/* Universal */
```
avishkar58, line 1: Is all of this additional CSS really needed? Also there seems to be a some very generally selectors like html button which should not belong inside a component-specific CSS file.

webapp/src/pages/ReponseDetails/components/ReviewModal.js
```diff
@@ -0,0 +1,121 @@
+import React, { Component } from "react";
```
avishkar58, line 1: This component doesn't seem to be used? Also seems to be a copy of ResponsePage/components/ReviewModal.js ?

webapp/src/pages/applicationForm/components/ApplicationForm.js
```diff
@@ -1138,34 +1406,104 @@ class ApplicationForm extends Component {
...
+      );
+    } else {
+      return (
+        <ApplicationFormInstance
+          formSpec={formSpec}
+          response={selectedResponse}
+          event={this.props.event}
+        />
+      );
     }
```
avishkar58, line 1507: In future, please can you keep formatting changes in a separate PR that only does formatting. When formatting changes are mixed up with other logic changes it makes it very hard for a reviewer to see what has actually changed

webapp/src/pages/eventHome/EventHome.js
```diff
@@ -154,11 +155,26 @@ class EventHome extends Component {
...
           render={(props) => <Application {...props} event={event} />}
         />
+        
         <Route
           exact
           path={`${match.path}/apply/new`}
           render={(props) => <Application {...props} event={event} journalSubmissionFlag={true} /> }
         />
+
+        {/*new route */}
```
avishkar58, line 165: Remove uneccessary comment

webapp/src/pages/reviewForm/components/ReviewForm.js
```diff
@@ -204,9 +204,13 @@ const ReviewForm = (props) => {
             active: currentStage !== 1 ? true : false
           })
           setCreateMode(true);
+          console.log(event);
```
avishkar58, line 207: Remove debug logging

webapp/src/services/applicationForm/applicationForm.service.js
```diff
@@ -85,6 +85,7 @@ function submit(applicationFormId, isSubmitted, answers) {
 
     return axios.post(baseUrl + `/api/v1/response`, response, {headers: authHeader()})
       .then(resp=> {
+        console.log()
```
avishkar58, line 88: REmove

PR #1305, opened 2024-12-10
Application status Home page

Vilmo18:
What's done?**  
- **Minor modification** to retrieve the application opening date on the backend.  
- **Added a script** to display the application status.


webapp/src/components/EventStatus.js
```diff
@@ -285,6 +285,20 @@ class EventStatus extends Component {
     }
   };
 
+  applicationOpeningStatus = (event) => {
+    return {
+      title: this.props.t("Applications Opening Soon"),
```
avishkar58, line 290: By the way - whenever you add some new text to the UI, you also need to add the text to the [English](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/public/locales/en/translation.json) and [French](https://github.com/deep-learning-indaba/Baobab/blob/develop/webapp/public/locales/fr/translation.json) translation files - that's what the `this.props.t` does - it looks up the phrase in the translation file for the user's selected language.

webapp/src/components/EventStatus.js
```diff
@@ -285,6 +285,20 @@ class EventStatus extends Component {
...
+      titleClass: "text-warning",
+      longText:
+        this.props.t("Applications for") +
+        ` ${event.name} ` +
+        this.props.t("will open on") +
+        ` ${event.application_open_date}. ` +
+        this.props.t("Please check back later to apply."),
```
avishkar58, line 297: For this kind of long text, to ensure that it gets translated correctly, you should use a translation block with placeholders to subsitute in the variables. See here for an example: https://github.com/deep-learning-indaba/Baobab/blob/8fb7327c8976bcd54d278804a321c7609811cca5/webapp/src/pages/references/components/ReferenceComponent.js#L106

avishkar58: (I know the rest of the examples in this file are wrong 🙈  if you could fix those too that would be highly apreciated! Thank you :) )

avishkar58 (changes requested):
Thank you Yvan! Much appreciated

webapp/src/components/EventStatus.js
Vilmo18: Got it

Vilmo18:
For this case.
I didn't use `<Trans>` but if you want I can do it, I just followed the template function.
For future needs I'll keep this in mind

avishkar58 (approved):

PR #1306, opened 2024-12-11
fixing error review deadline for journal

Vilmo18:
Given that there is no review_close date(null) for event journal (deadline) that produce (error r-xx where xx-number) 
We have just fixed a date when we encounter this case

### DIFF
--- a/webapp/src/pages/reviewForm/components/ReviewForm.js
+++ b/webapp/src/pages/reviewForm/components/ReviewForm.js
@@ -200,7 +200,7 @@ const ReviewForm = (props) => {
             isOpen: props.event.is_review_open,
             eventId: event.event.id,
             stage: currentStage,
-            deadline: event.event.review_close,
+            deadline: event.event.review_close==null?"2099-12-05T23:59:59":event.event.review_close,
             active: currentStage !== 1 ? true : false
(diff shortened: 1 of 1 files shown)


webapp/src/pages/reviewForm/components/ReviewForm.js
```diff
@@ -200,7 +200,7 @@ const ReviewForm = (props) => {
             isOpen: props.event.is_review_open,
             eventId: event.event.id,
             stage: currentStage,
-            deadline: event.event.review_close,
+            deadline: event.event.review_close==null?"2099-12-05T23:59:59":event.event.review_close,
```
avishkar58, line 203: You can do this slightly more concisely (and actually more correctly because it also caters for other kinds of null-like values) by using the null coalescing operator instead: 
`event.event.review_close ?? "2099-12-05T23:59:59"`

avishkar58 (changes requested):
Thanks Yvan! Just a tiny nit, but otherwise looks great

Vilmo18:
This syntax requires few packages to be install.
So I've used the syntax with  `||`

avishkar58 (approved):

PR #1307, opened 2024-12-17
Add new changes 

Vilmo18:
Very big updates are coming!! (Sorry  for some files with differents formatting than the original)

-  Fixing the problem of multiple submissions by updating the condition
-  Adding a new components for visualizing submissions and related submissions (Using the chain mechanism between submissions and some checking for avoiding vulnerability `ApplicationForm`, `ApplicationList` and `ResponseDetails` 
- **Addition of new statuses via migrations**: Previously, new statuses were added without corresponding database changes. A migration script has now been created to ensure these changes are properly reflected in the database. (With improvement of the scripts based on previous reviews espicially `get` function)
- **Outcome and response bug fix**: Previously, there was one outcome shared across all responses given an event and user. This has been updated so that each response has its own unique outcome. The database structure was modified, migrations were applied, and necessary changes were made to back-end files.
- Adding new words for traduction
- **Minor bug fixes**: Several smaller bugs have been identified and resolved.


avishkar58:
Hi Yvan! Thanks for the changes. 
Just to clarify - is this the same as the previous PR with additional fixes? In that case, please don't raise a new PR, but rather update the existing PR so that I (generally, the reviewer) can see what has changed since the last review. Otherwise I have to start reviewing from scratch which for a big PR like this is a lot of extra work. 
Also please can you name your PRs more descriptively, to say what functionality it adds, rather than the generic "add new changes" - thank you very much, and wishing you a Merry Christmas! :)

Vilmo18:
Hi @avishkar58 , I’d be happy to break down this PR into smaller, more manageable PRs to facilitate your review process.

api/app/events/status.py
```diff
@@ -59,7 +59,7 @@ def get_event_status(event_id, user_id):
     else:
         application_status = 'Not Submitted'
 
-    outcome = outcome_repository.get_latest_by_user_for_event(user_id, event_id)
+    outcome = outcome_repository.get_latest_by_user_for_event_response(user_id,None, event_id)
```
avishkar58, line 62: Having the response_id sometimes hard-coded to None is a dangerous practice that can lead to bugs. If we're going to link outcomes to responses, then we need to do it properly, which means:
- response_id should not be nullable
- there should always be a valid response_id when querying for an outcome.

api/migrations/versions/4923563b3a63_add_reponseID_outcome.py
```diff
@@ -0,0 +1,42 @@
...
+revision = '4923563b3a63'
+down_revision = 'bf4cb562e36c'
+
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column(
```
avishkar58, line 19: In order to make this non-nullable, you'll have to start with this nullable column as you have it here, then run some SQL to update (and therefore set) the response_id for all existing outcome records, and then update the column to be non-nullable.

webapp/src/components/EventStatus.js
```diff
@@ -91,6 +93,19 @@ class EventStatus extends Component {
           
       }
       else if (event.status.application_status === "Withdrawn") {
+        if (event.event_type === "JOURNAL"){
+            return {
+                titleClass: "text-success",
```
avishkar58, line 98: I don't understand this, why would it say "you have submitted your article" when the application status is withdrawn?

webapp/src/pages/ReponseDetails/ResponseDetails.css
avishkar58, line 1: The formatting changes that have been applied to this file make it impossible for me to see and review what logic has changed

webapp/src/pages/applicationForm/components/ApplicationForm.js
avishkar58, line 1: Same here, i can't see what has changed outside of the formatting.

